PR #22, opened 2025-04-03
BAU Tweaks to templates/ routes

sfount:
Starting to think through consistent route/ handler names, definitely don't feel this is finished but hopefully gives us a jumping off point for when we want to standardise this. 

This will need rebasing with some of the testing/ test structure changes being otherwise worked on so will leave it as draft until I've done that.


app/platform/forms.py
```diff
@@ -5,5 +5,10 @@
...
 
 class GrantForm(FlaskForm):
-    name = StringField("Grant name", validators=[DataRequired()], widget=GovTextInput())
+    name = StringField(
+        "Grant name",
+        validators=[DataRequired("Enter a grant name")],
+        filters=[lambda x: x.strip() if x else x],
+        widget=GovTextInput(),
+    )
     submit = SubmitField("Submit", widget=GovSubmitInput())
```
samuelhwilliams, line 14:
```suggestion
    submit = SubmitField("Set up grant", widget=GovSubmitInput())
```

sfount: this form gets used in set up/ edit so this is done in the template - I don't love it

app/platform/templates/platform/grant_create.html
```diff
@@ -0,0 +1,32 @@
...
+{% block content %}
+    <div class="govuk-grid-row">
+         <div class="govuk-grid-column-two-thirds">
+             <h1 class="govuk-heading-l">Set up a new grant</h1>
+             <form method="post" novalidate>
+                 {{ form.csrf_token }}
+                 {{ form.name }}
+                 {{ form.submit(params={
+                    "text": "Set up grant"
+                }) }}
```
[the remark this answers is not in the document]
sfount: see other message

app/common/data/interfaces/grants.py
```diff
@@ -9,16 +9,16 @@
...
+    return db.session.get_one(Grant, grant_id)
 
 
 def get_all_grants() -> Sequence[Grant]:
     statement = select(Grant).order_by(Grant.name)
     return db.session.scalars(statement).all()
 
 
-def create_grant(name: str) -> Grant:
+def add_grant(name: str) -> Grant:
```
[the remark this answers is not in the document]
sfount: the HTTP handler is called `create_grant` 🫣 if we had a default handler suffix like `_handler` that would avoid this? happy for other suggestions but the data method and the handler can't have the same name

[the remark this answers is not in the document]
MarcUsher: Haha Sarah and I had this exact discussion when we set this up 😆 - our feeling was that `create_grant` felt more appropriate for data method that would actually create it in the DB. But also don't feel _that_ strongly about it, I think it's just important we be consistent with the naming

app/platform/forms.py
samuelhwilliams: Ok, maybe just have the label here as an empty string then?

sfount: Sounds good - as we recreate this pattern in various places I think I'd be in favour of explicit forms for creating and updating. I don't think there's going to end up being much overlap

samuelhwilliams: agree

app/common/data/interfaces/grants.py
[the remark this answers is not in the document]
sfount: I've gone with aliasing the interface as `data` - happy to revisit this but I think we're slightly on the side of avoiding `_handler` to make url_for a bit more intuitive

app/platform/__init__.py
```diff
@@ -4,13 +4,8 @@
 from werkzeug import Response
 from wtforms.fields.core import Field
 
+from app.common.data.interfaces import grants as data
```
samuelhwilliams, line 7: :sweating-towel-guy: not like this

samuelhwilliams: But I don't have a much better option 😭 

I think I was imagining it being more like `import app.common.data.interfaces as data` - but that means aggregating all the stuff in that module.

Otherwise this feels like maybe it should be `from app.common.data.interfaces import grants as grants_data`. Because we'll have a problem as soon as we need to import any interface module beyond the first with this setup.

But this all feels grim.

samuelhwilliams: So - yolo for now? 🫠

app/common/data/interfaces/grants.py
```diff
@@ -9,8 +9,8 @@
 from app.extensions import db
 
 
-def get_grant(grant_id: UUID4) -> Grant | None:
-    return db.session.get(Grant, grant_id)
+def get_grant(grant_id: UUID4) -> Grant:
+    return db.session.get_one(Grant, grant_id)
```
samuelhwilliams, line 13: 🤯 so unfamiliar with this syntax but it feels nice

app/platform/__init__.py
```diff
@@ -1,67 +1,61 @@
...
+@platform_blueprint.route("/grants/set-up", methods=["GET", "POST"])
 @auto_commit_after_request
-def add_grant() -> str | Response:
+def create_grant() -> ResponseReturnValue:
     form = GrantForm()
     if form.validate_on_submit():
         try:
-            create_grant(name=form.name.data)  # type: ignore
-            return redirect(url_for("platform.view_all_grants"))
+            assert form.name.data is not None
```
samuelhwilliams, line 21: Yeah I think this is interesting - it sort of works, but also in some optimised python runs I think asserts are stripped. I tend towards `cast(str, form.name.data)` but I'm very unconvinced which way is 'better'/'correct'.

samuelhwilliams: Might need to look into wtforms a bit more and understand if they have a typing-compatible answer to this.

samuelhwilliams (approved):

app/platform/__init__.py
sfount: Agree it would be good to have something at the wtforms/ serialise level -- this would be gnarly if there was a lot of values.

For now this is according to one of their docs
```python
# If you know a value can never be None due to some logic that mypy doesn't
# understand, use an assert
assert x is not None
print(x.upper())
```

https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

I guess it happening after forms validate means you're not relying on the runtime behaviour

samuelhwilliams: yeah I also wasn't thinking very hard and realise we could still theoretically run with the asserts stripped out because the type checker isn't runtime obviously 🤦

PR #27, opened 2025-04-04
BAU Tweak vite build settings

sfount:
Tweaks the config to:
- stop the docker compose logs being hijacked by vite
- have the vite styles ready to go when booting up the dev server for the first time
- use the flask interface rather than dipping into node

The intention was to have the bundler split out the CSS instead of having it load the SASS bundled in a dynamic JavaScript import. The dynamic JavaScript import isn't cached by the browser and is what's causing the flash of unstyled content. Splitting it out into a separate CSS file would allow us to continue to get watch/ live reload but avoid the FOUC. I wasn't able to appease it this time round.

### DIFF
--- a/app/vite/vite.config.mjs
+++ b/app/vite/vite.config.mjs
@@ -1,16 +1,22 @@
 /* jshint esversion: 6 */
-
 import { viteStaticCopy } from 'vite-plugin-static-copy';
 
 export default {
   server: {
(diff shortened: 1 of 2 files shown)


srh-sloan (approved):
> FIX: don't try to warmup the world
A good message for life in general!

PR #29, opened 2025-04-04
Send [non-functional] magic link emails

samuelhwilliams:
This patch integrates GOV.UK Notify and starts sending emails when you submit the 'sign in' page. The email does not yet contain an actually-usable magic link - just stub content. This PR is just about the integration with GOV.UK Notify.


.env.example
```diff
@@ -1,6 +1,2 @@
-DATABASE_HOST=localhost
-DATABASE_PORT=5432
-DATABASE_NAME=postgres
-DATABASE_SECRET='{"username":"postgres","password":"postgres"}'  # pragma: allowlist secret
-SERVER_NAME=funding.communities.gov.localhost:8080
-SECRET_KEY=pretend_secret_key
```
srh-sloan, line 6: Don't we still want these in here?

app/services/notify.py
```diff
@@ -0,0 +1,91 @@
...
+            return Notification(id=uuid.UUID(notification_data["id"]))
+        except (TokenError, APIError) as e:
+            raise NotificationError() from e
+
+    def send_magic_link(
+        self,
+        email_address: str,
+        magic_link_url: str,
+        magic_link_expires_at_utc: datetime.datetime,
+        request_new_magic_link: str,
```
srh-sloan, line 75:
```suggestion
request_new_magic_link_url: str,
```

docker-compose.yml
```diff
@@ -33,6 +33,7 @@ services:
       python -m flask db upgrade & \
       python -m debugpy --listen 0.0.0.0:8081 -m flask run --host 0.0.0.0 --port 8080 --reload --debug --cert=/app/certs/cert.pem --key=/app/certs/key.pem
       "
+    env_file: .env
```
srh-sloan, line 36: If we are using .env can we add copying .env.example to the makefile, or the readme?

srh-sloan (commented):
Looking good! Just some minor comments 👍

.env.example
samuelhwilliams: Do we? I'm not sure what they're for to be honest?

srh-sloan: If you run stuff locally like `uv run flask db migrate` I think it needs them? ASsuming you are doing `source .env`

srh-sloan (approved):

app/common/auth/__init__.py
```diff
@@ -19,6 +23,13 @@ def sign_in() -> ResponseReturnValue:
         # TODO: all session stuff will be revisited as part of FSPT-334
         session["email_address"] = email
 
+        notification_service.send_magic_link(
+            email,  # type: ignore[arg-type]
+            "https://magic-link-tbd",
+            datetime.datetime.now(ZoneInfo("UTC")) + datetime.timedelta(minutes=15),
+            "https://new-magic-link-tbd",
+        )
```
gidsg, line 31: Could we name these arguments to make it clearer what's going on?

gidsg (approved):
👍 I think the .envrc stuff looked unrelated, so maybe clearer in a separate PR? But will approve this one for now.

PR #33, opened 2025-04-08
First pass environment that doesn't need db settings

srh-sloan:
Suggested fix for having to pass in dummy database values in various scripts, eg. pipelines, makefile. In this case you would set `FLASK_ENV=scripts` and then you don't need to pass any other env setup values.

Thoughts please!


sfount:
Closed by https://github.com/communitiesuk/funding-service/pull/61

PR #34, opened 2025-04-08
Check `unit_test` env for CI

samuelhwilliams:


sfount:
I suspect this can be merged following https://github.com/communitiesuk/funding-service/pull/36

sfount:
If we do that this closes https://github.com/communitiesuk/funding-service/pull/33 for the time being?

PR #35, opened 2025-04-08
FSPT-278: Create-view-edit grant e2e playwright test

srh-sloan:
This adds a comprehensive playwright end to end test covering the main Grant journey:

- All grants page
- create a new grant
- check expected validation errors
- navigate to individual grant dashboard
- navigate to grant settings
- change grant name
- check original grant name no longer appears in all grants list and new grant name does

Checks headers/page identifiers, buttons etc. are all present and work correctly.

E2E tests now also work against the dev environment in CI when a PR is merged - getting the e2e tests to run on a PR against 'local' docker compose will be addressed in a separate PR.

_NOTE_ - in the old copilot world, we knew when a deploy had finished as the copilot command and logs would tell us. In this new world we build and push an image to ECR and AppRunner sees that there's a new image with the `latest` tag and deploys it. That step happens outside of the GitHub workflow so we don't currently have an easy way of knowing when a deploy has finished and therefore a good time to start the e2e tests running against the newly deployed version of the service.

For now this part of the pipeline is commented out until we are able to address this properly - [see ticket covering this](https://mhclgdigital.atlassian.net/browse/FSPT-377)


README.md
```diff
@@ -45,8 +45,12 @@ Developers are expected to run the app locally using [docker-compose](https://do
 ## E2E Tests
 All E2E (browser) tests should live inside [/tests/e2e](./tests/e2e).
 
+As a one-off setup step, run `playwright install` to download and configure the browsers needed for these tests.
```
samuelhwilliams, line 48:
```suggestion
As a one-off setup step, run `uv run playwright install` to download and configure the browsers needed for these tests.
```

tests/e2e/conftest.py
```diff
@@ -0,0 +1,34 @@
+from dataclasses import dataclass
+from typing import Generator
+
+import pytest
+
+
+@dataclass
+class FundingServiceDomains:
+    landing_url: str
```
samuelhwilliams, line 9: We maybe don't need this dataclass (yet) given this app is probably only having a single domain?

tests/e2e/pages.py
```diff
@@ -0,0 +1,120 @@
...
+class ChangeGrantNamePage(GrantDashboardBasePage):
+    backlink: Locator
+    title: Locator
+
+    def __init__(self, page: Page, domain: str) -> None:
+        super().__init__(page, domain)
+        self.backlink = self.page.get_by_role("link", name="Back")
+        self.title = self.page.get_by_role("heading", name="Grant name")
+
+    T = TypeVar("T", bound=BasePage)
```
samuelhwilliams, line 112: I think this is redundant now?

tests/e2e/test_create_view_edit_grant.py
```diff
@@ -0,0 +1,67 @@
...
+from playwright.sync_api import Page, expect
+
+from tests.e2e.conftest import FundingServiceDomains
+from tests.e2e.pages import AllGrantsPage, ChangeGrantNamePage, GrantDashboardPage, GrantSettingsPage, NewGrantPage
+
+
+def test_create_view_edit_grant_success(page: Page, domains: FundingServiceDomains):
+    # Set a larger than default viewport size to run locally as otherwise Flask Debug Toolbar
+    # interferes with some elements
+    page.set_viewport_size({"width": 1600, "height": 1200})
```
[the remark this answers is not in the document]
MarcUsher: Ah noice yeah it does, will copy that implementation 👍 thanks current Sam and past Sam

tests/e2e/conftest.py
MarcUsher: Good shout, I've simplified this down

tests/e2e/pages.py
MarcUsher: Yeah we know what types of page should be returned in these instances so the generic thing is maybe not needed, good shout, I've simplified this 🙌

tests/e2e/test_create_view_edit_grant.py
```diff
@@ -0,0 +1,62 @@
...
+    all_grants_page.navigate()
+    expect(all_grants_page.title).to_be_visible()
+
+    # Set up new grant
+    new_grant_page: NewGrantPage = all_grants_page.click_set_up_new_grant()
+    expect(new_grant_page.title).to_be_visible()
+
+    new_grant_name = f"E2E {uuid.uuid4()}"
+    new_grant_page.complete_grant_name(new_grant_name)
+    all_grants_page: AllGrantsPage = new_grant_page.submit_new_grant_form()
```
samuelhwilliams, line 19:
```suggestion
    all_grants_page = new_grant_page.submit_new_grant_form()
```

I think these manual type hints shouldn't be needed, because the methods set their return types

tests/e2e/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+    page.set_viewport_size({"width": int(width), "height": int(height)})
+
+
+@pytest.fixture
+def get_e2e_params(request: pytest.FixtureRequest) -> Generator[dict[str, str], None, None]:
+    e2e_env = request.config.getoption("e2e_env", "local")
+    # vault_profile = request.config.getoption("e2e_aws_vault_profile", None)
+    # session_token_from_env = os.getenv("AWS_SESSION_TOKEN", None)
+    # if not session_token_from_env and e2e_env != "local" and not vault_profile:
+    #     sys.exit("Must supply e2e-aws-vault-profile with e2e-env")
```
sfount, line 20: Looks like these can be removed? I don't think we'll need to do the read-from-notify workaround when we fully control how the auth is working but could check in with the magic links work as it progresses

tests/e2e/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.fixture()
+def domain(request: pytest.FixtureRequest, get_e2e_params: dict[str, str]) -> str:
+    e2e_env = get_e2e_params["e2e_env"]
+
+    if e2e_env == "local":
+        return "https://funding.communities.gov.localhost:8080"
+    if e2e_env == "dev":
+        return "https://vniusepvmn.eu-west-2.awsapprunner.com/"
```
sfount, line 34: Something about this feels like it should come from the environment rather than a fixture/ map in the code

sfount (commented):
Looking good! I've not looked through the test cases in detail as I think Sam's got context but a few things that shouldn't block this PR

tests/e2e/conftest.py
MarcUsher: Yeah good shout I think these were leftover from copying over some of the set-up from post-award but we don't need that until we need to do things that access AWS secrets etc. Will remove and keep it clean for now!

tests/e2e/conftest.py
MarcUsher: Yeah this looks a little unnecessary for just the URL - I guess this kind of mapping is handy when there's a bit more to it like in pre/[post award](https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/7fc2533cabea324178d939bd428d959fda242953/tests/e2e_tests/conftest.py#L23) set-up but we're unlikely to have more domains in the new world

tests/e2e/test_create_view_edit_grant.py
```diff
@@ -0,0 +1,32 @@
...
+    # Change grant name
+    change_grant_name_page = grant_settings_page.click_change_grant_name(new_grant_name)
+    edited_grant_name = f"{new_grant_name} - edited"
+    change_grant_name_page.fill_in_grant_name(edited_grant_name)
+    grant_settings_page = change_grant_name_page.click_submit(edited_grant_name)
+
+    # Go back to Grants list and check new name appears/old name doesn't appear
+    all_grants_page = grant_settings_page.click_grants()
+    all_grants_page.check_grant_exists(edited_grant_name)
+    all_grants_page.check_grant_doesnt_exist(new_grant_name)
```
samuelhwilliams, line 32: :chefkiss:

samuelhwilliams (approved):
It's so fast and reliable 🥺 

image

PR #36, opened 2025-04-08
FSPT-311 Assets use vite directly

sfount:
Uses vite directly instead of using `flask-vite`, which is designed around Single Page Applications. This allows us to directly point to our SCSS and JS files and feels more like a traditional asset pipeline.

In no particular order: 
- remove the FOUC (Flash of unstyled content) that was happening locally as the vite dev server was previously serving a javascript file which then inlined styles into the page (as you would in a SPA) - now the JavaScript and CSS files are directly loaded and cached (or 304 not modified in the case of the dev server)
- include source maps for JS and SASS which allow any issues in either styles or JavaScript to point to the JS/ SASS rather than the output JS/ CSS. This works because we're pointing directly as the styles/ js from the dev server
- fixup logs warning about deprecations and copying external files
- simplify the configuration into a standard `assets` folder which looks a lot more familiar to Flask projects
- rely _entirely_ on the dev server when running locally (previously some assets came from the dev server and some relied in a production "build" having run). This works by updating the Jinja templates asset path to also point at the dev server and means that you don't even need to have node installed to run it locally if you're using docker (node will then be required to build the production assets)

Aiming to preserve:
- using modern frontend tooling that isn't deprecated or no longer well supported
- fetching and keeping up to date with GOV.UK Frontend in a reliable way
- HMR/ server reloads when styles or frontend JavaScript is changed

Try it using

```
make clean-build

# this will run the app and the dev server needed to serve (all) of the assets the app depends on
make up
```

Separately you can try out a static build

```
npm i
npm run build
```


sfount:
The amount of logs you should be seeing 

```
npm run build
```

```
> funding-service@1.0.0 build
> vite build

vite v6.2.5 building for production...
✓ 25 modules transformed.
app/assets/dist/manifest.json               0.27 kB │ gzip:  0.14 kB
app/assets/dist/assets/main-wrSXu5CS.css  126.77 kB │ gzip: 15.25 kB
app/assets/dist/assets/main-8UHBOLAm.js    48.34 kB │ gzip: 11.90 kB
[vite-plugin-static-copy] Copied 4 items.
✓ built in 623ms
```

docker-compose.yml
```diff
@@ -28,9 +29,8 @@ services:
       bash -c "
       cp /app/certs/rootCA.pem /usr/local/share/ca-certificates/rootCA.crt && \
       update-ca-certificates && \
-      uv run flask db upgrade
-
-      (uv run flask vite install && uv run flask vite start) &
+      npm run dev -- --host & \
+      uv run flask db upgrade & \
```
samuelhwilliams, line 33: Not sure if the reordering is intentional here? I think I still would like cp/update/db upgrade to be a single group that runs first and if it fails blocks everything, then npm run dev/python run server as two parallel processes? Is that a bad idea? 🤔

samuelhwilliams:
but in general chefskiss, this feels v v nice and much better, wonderful work 🙇

docker-compose.yml
sfount: Yeah good shout, I think this was just a slightly careless rebase that didn't catch the fix on main

app/config/__init__.py
```diff
@@ -98,15 +98,14 @@ def SQLALCHEMY_ENGINES(self) -> dict[str, str]:
...
-    VITE_AUTO_INSERT: bool = False
-    VITE_FOLDER_PATH: str = "app/vite"
-
     # GOV.UK Notify
     GOVUK_NOTIFY_DISABLE: bool = False
     GOVUK_NOTIFY_API_KEY: str
     GOVUK_NOTIFY_MAGIC_LINK_TEMPLATE_ID: str = "9773e73c-85a1-4c3f-a808-02b9623616a3"
 
+    ASSETS_VITE_BASE_URL: str = "http://localhost:5173"
+    ASSETS_VITE_LIVE_ENABLED: bool = False
```
samuelhwilliams, line 107: Could this be local-only config? I think the idea of this ever being able to be turned on in a deployed environment feels odd maybe.

samuelhwilliams: But if the principle we're sticking to is "all config must be declared on the base class" - I can also get behind that.

samuelhwilliams (approved):
Yep, lovely, very very nice - good work. Side quest complete ✅

PR #37, opened 2025-04-09
Fix some HTML and add HTML tag validation

samuelhwilliams:
Fixes two issues:
- We had two error summaries on the create grant page
- We had two `<main>` elements on the common/base template.

Also adds a meta test that will run all HTML responses from our app through an html5 validation lib and flag any structural errors. This didn't catch things like the double main or duplicate IDs, but did catch some unmatched closing tags so does provide _some_ value.


sfount (approved):
Nice! First reflection was that this felt like a linting thing but I guess this catches it after its been rendered where plenty of other things could have crept in. Very neat idea

PR #38, opened 2025-04-09
Update GOV.UK Notify service

samuelhwilliams:
We now have a (trial) service called 'The Funding Service'. Currently we're on 'Access Funding'.

This update points us at the magic link template in the new GOV.UK Notify service.

Environments will need their API key updating as this rolls out.


MarcUsher (approved):

tests/unit/app/services/test_notify.py
```diff
@@ -23,7 +23,7 @@ def test_send_magic_link(self, app):
                 matchers.json_params_matcher(
                     {
                         "email_address": "test@test.com",
-                        "template_id": "c19811c2-dc4a-4504-99b5-7bcbae8d9659",
+                        "template_id": "9773e73c-85a1-4c3f-a808-02b9623616a3",
```
sfount, line 26: Nit: but I'm not sure this is meaningfully asserting much I think it would fine for this just to check its an instance of a string or something?

Maybe there is use for one that checks the actual value is passed through appropriately but maybe not all of them

Have possibly gone back on my nit while writing it - happy either which way

PR #42, opened 2025-04-09
Autouse DB fixture; remove redundant fixtures

samuelhwilliams:
The `app` fixture was relying on `setup_db_container`, which spins up postgres but does not run migrations. Anything that needs the db to exist should be relying on the `db` fixture.

This patch changes app to rely on the `db` fixture, which gives it a migrated db.

I've also set the `db_session` fixture as an autouse one. This means every single test in the `integrations` folder will always have a transactionally-isolated session. It means a slight bit of waste/overhead on any 'integration' tests that don't need the db, but the confidence it gives us that future developers will always be writing tests with in a safe space feels worth this for now.

### DIFF
--- a/tests/integration/common/data/test_grants.py
+++ b/tests/integration/common/data/test_grants.py
@@ -16,12 +16,12 @@ def test_get_all_grants(factories):
     assert len(result) == 5
 
 
-def test_create_grant(db) -> None:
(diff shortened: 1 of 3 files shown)


sfount (approved):
Agree this should be set by default, nice tidy up 👍

PR #43, opened 2025-04-10
Generate and use magic links

samuelhwilliams:

## Summary
This PR adds basic User and MagicLink tables, and updates the sign in flow to generate a magic link and insert it into the email.

A new stub page is added to claim the magic link, but no auth session is actually created yet. This stub page will eventually be mostly-invisible to users, as we'll add JS to automatically submit the form and claim the link.

The redirected-to page after claiming a magic link is currently hard-coded to the list grants page; again, this to be updated separately.

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -16,9 +16,9 @@ auth_blueprint = Blueprint(
 )
 
 
-@auth_blueprint.route("/sign-in", methods=["GET", "POST"])
+@auth_blueprint.route("/request-a-link-to-sign-in", methods=["GET", "POST"])
 @auto_commit_after_request
-def sign_in() -> ResponseReturnValue:
+def request_a_link_to_sign_in() -> ResponseReturnValue:
     form = SignInForm()
     if form.validate_on_submit():
         email = cast(str, form.email_address.data)
@@ -34,7 +34,7 @@ def sign_in() -> ResponseReturnValue:
             email,
             magic_link_url=url_for("auth.claim_magic_link", magic_link_code=magic_link.code, _external=True),
             magic_link_expires_at_utc=magic_link.expires_at_utc,
-            request_new_magic_link_url=url_for("auth.sign_in", _external=True),
+            request_new_magic_link_url=url_for("auth.request_a_link_to_sign_in", _external=True),
         )
 
         return redirect(url_for("auth.check_email", magic_link_id=magic_link.id))
@@ -56,7 +56,7 @@ def check_email(magic_link_id: uuid.UUID) -> ResponseReturnValue:
 def claim_magic_link(magic_link_code: str) -> ResponseReturnValue:
     magic_link = interfaces.magic_link.get_magic_link(code=magic_link_code)
     if not magic_link or not magic_link.usable:
-        return redirect(url_for("auth.sign_in"))
+        return redirect(url_for("auth.request_a_link_to_sign_in"))
 
     form = ClaimMagicLinkForm()
     if form.validate_on_submit():
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -1,11 +1,13 @@
-import datetime
-from zoneinfo import ZoneInfo
+import uuid
+from typing import cast
 
-from flask import Blueprint, redirect, render_template, session, url_for
+from flask import Blueprint, abort, redirect, render_template, url_for
 from flask.typing import ResponseReturnValue
 
-from app.common.auth.forms import SignInForm
-from app.extensions import notification_service
+from app.common.auth.forms import ClaimMagicLinkForm, SignInForm
+from app.common.data import interfaces
+from app.common.data.interfaces.user import get_or_create_user
+from app.extensions import auto_commit_after_request, notification_service
 
 auth_blueprint = Blueprint(
     "auth",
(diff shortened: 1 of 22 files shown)


tests/integration/common/data/test_magic_link.py
```diff
@@ -0,0 +1,62 @@
...
+        retrieved_magic_link = interfaces.magic_link.get_magic_link(code=magic_link.code)
+
+        assert magic_link is retrieved_magic_link
+
+
+class TestClaimMagicLink:
+    def test_claim_magic_link(self, db_session, factories):
+        magic_link = factories.magic_link.create()
+        now = db_session.scalar(func.current_timestamp()).replace(tzinfo=None)
+        assert magic_link.claimed_at_utc != now
```
samuelhwilliams, line 58: probably should just be 

```suggestion
        assert magic_link.claimed_at_utc is None
```

tests/integration/common/auth/test_auth.py
```diff
@@ -1,55 +1,99 @@
...
 
-def test_sign_in_page_post_invalid_email(client):
-    response = client.post(url_for("auth.sign_in"), data={"email_address": "invalid-email"})
-    assert response.status_code == 200
-    soup = BeautifulSoup(response.data, "html.parser")
-    assert page_has_error(soup, "Enter an email address in the correct format")
 
+class TestClaimMagicLinkView:
+    def test_get(self, client, factories):
+        pass
```
samuelhwilliams, line 57: ohno

app/common/data/models/__init__.py
```diff
@@ -7,3 +12,25 @@ class Grant(BaseModel):
...
+
+
+class MagicLink(BaseModel):
+    __tablename__ = "magic_link"
+
+    code: Mapped[str] = mapped_column(unique=True, default=lambda: secrets.token_urlsafe(12))
+    user_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("user.id"))
+    redirect_to_path: Mapped[str]
+    expires_at_utc: Mapped[datetime.datetime]
+    claimed_at_utc: Mapped[datetime.datetime | None]
```
MarcUsher, line 33: Just a small naming thing question, here we're explicitly saying these columns are utc, which we don't say for the `created_at`/`updated_at` in the base model - for some people that might imply they're not necessarily utc? Do we want to always explicitly state that datetime columns are utc where we have them in models to avoid ambiguity?

MarcUsher (commented):
2 small comments/questions that are more for my own education than anything else, will peep at this again after lunch though with fresh eyes 👀

app/common/data/models/__init__.py
samuelhwilliams: yeah I think we should probably go back to created_at/updated_at and suffix them with `_utc` also

app/common/data/interfaces/magic_link.py
```diff
@@ -0,0 +1,36 @@
...
+    if id_ and code or not id_ and not code:
+        raise ValueError("Must provide exactly one of `id_` and `code`")
+    elif id_:
+        return db.session.scalar(select(MagicLink).where(MagicLink.id == id_))
+    else:
+        return db.session.scalar(select(MagicLink).where(MagicLink.code == code))
+
+
+def claim_magic_link(magic_link: MagicLink) -> None:
+    magic_link.claimed_at_utc = func.current_timestamp()
```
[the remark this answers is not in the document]
samuelhwilliams: current_timestamp and now are both in whatever the DB's server time is (which is commonly, and for us also, is UTC), yep.

tests/integration/common/auth/test_auth.py
```diff
@@ -1,55 +1,103 @@
...
+        )
 
-def test_check_email_page_without_session(client):
-    response = client.get("/check-your-email", follow_redirects=True)
-    assert response.status_code == 200
-    assert b"Request a link to sign in" in response.data
-    assert url_for("auth.sign_in") in response.request.path
+        assert response.status_code == 302
+        assert response.location == "/my-redirect"
+        assert magic_link.claimed_at_utc is not None
```
samuelhwilliams, line 103: Very open to thoughts on whether we should be poking the magic link directly or mocking out the interface call. Or something else.

app/common/auth/__init__.py
```diff
@@ -14,31 +16,52 @@
...
 
         notification_service.send_magic_link(
-            email,  # type: ignore[arg-type]
-            magic_link_url="https://magic-link-tbd",
-            magic_link_expires_at_utc=datetime.datetime.now(ZoneInfo("UTC")) + datetime.timedelta(minutes=15),
-            request_new_magic_link_url="https://new-magic-link-tbd",
+            email,
+            magic_link_url=url_for("auth.claim_magic_link", magic_link_code=magic_link.code, _external=True),
+            magic_link_expires_at_utc=magic_link.expires_at_utc,
+            request_new_magic_link_url=url_for("auth.request_a_link_to_sign_in", _external=True),
```
sfount, line 37: I think all of this can now be extracted from the magic link in the helper but that doesn't need to be done in this PR, feels like something that be done in a cleanup at the end once all the mechanics are in

app/common/auth/__init__.py
```diff
@@ -14,31 +16,52 @@
...
 
 
-@auth_blueprint.get("/check-your-email")
-def check_email() -> ResponseReturnValue:
-    if "email_address" not in session:
-        return redirect(url_for("auth.sign_in"))
+@auth_blueprint.get("/check-your-email/<uuid:magic_link_id>")
+def check_email(magic_link_id: uuid.UUID) -> ResponseReturnValue:
+    magic_link = interfaces.magic_link.get_magic_link(id_=magic_link_id)
+    if not magic_link or not magic_link.usable:
```
sfount, line 48: Link this to our convo on managing errors and default behaviour for things not being there/ having an incorrect ID etc. -- i.e in this case would it make sense for get magic link to only return usable valid links and otherwise not found is propagated and managed by the middleware (maybe not! having it explicit is fine too just another data point)

app/common/data/interfaces/magic_link.py
```diff
@@ -0,0 +1,37 @@
...
+import uuid
+
+from sqlalchemy import func, select, update
+
+from app.common.data.models import MagicLink, User
+from app.extensions import db
+
+
+def create_magic_link(user: User, *, redirect_to_path: str) -> MagicLink:
+    db.session.execute(update(MagicLink).where(MagicLink.user == user).values(expires_at_utc=func.current_timestamp()))
```
sfount, line 11: Not a suggestion I've thought through at all but I wonder if these could just be deleted which would be a form of keeping the table size ~limited to the number of active users

app/common/data/interfaces/user.py
```diff
@@ -0,0 +1,19 @@
+from sqlalchemy.dialects.postgresql import insert as postgresql_upsert
+
+from app.common.data.models import User
+from app.extensions import db
+
+
+def get_or_create_user(email_address: str) -> User:
+    # This feels like it should be a `on_conflict_do_nothing`, except in that case the DB won't return any rows
+    # So we use `on_conflict_do_update` with a noop change, so that this upsert will always return the User regardless
+    # of if its doing an insert or an 'update'.
```
sfount, line 10: 👍

app/common/data/models/__init__.py
```diff
@@ -7,3 +13,29 @@ class Grant(BaseModel):
...
+
+    user: Mapped[User] = relationship("User", back_populates="magic_links")
+
+    __table_args__ = (Index(None, code, unique=True, postgresql_where="claimed_at_utc IS NOT NULL"),)
+
+    @property
+    def usable(self) -> bool:
+        return self.claimed_at_utc is None and self.expires_at_utc > datetime.datetime.now(utc).replace(tzinfo=None)
```
sfount, line 41: Nice this feels like a clean abstraction

pyproject.toml
```diff
@@ -48,6 +48,8 @@ dev = [
     "types-wtforms>=3.2.1.20250304",
     "html5lib>=1.1",
     "types-html5lib>=1.1.11.20241018",
+    "types-pytz>=2025.2.0.20250326",
+    "freezegun>=1.5.1",
```
sfount, line 52: I might just be missing something but does this PR consume this new dependency?

tests/integration/common/auth/test_auth.py
sfount: I think this makes sense to me - if the code was redirecting but not appropriately setting this then the link could be claimed multiple times and it would impact the main purpose of the security, this feels more thorough than you might need to be in other instances but for this it's 👌? I'll have a bit more of a think

sfount (approved):
This looks great 👍 sorry my review is a bit scattered but no issues with the code and the coverage looks comprehensive

app/common/auth/__init__.py
```diff
@@ -14,31 +16,52 @@
...
+@auto_commit_after_request
+def request_a_link_to_sign_in() -> ResponseReturnValue:
     form = SignInForm()
     if form.validate_on_submit():
-        email = form.email_address.data
+        email = cast(str, form.email_address.data)
 
-        # TODO: all session stuff will be revisited as part of FSPT-334
-        session["email_address"] = email
+        user = get_or_create_user(email_address=email)
```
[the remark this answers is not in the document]
samuelhwilliams: Yeah definitely interesting. I guess we've got a `select user join magic link where magic link.user == user and magic link.claimed at is not null` as a quick check for real users.

But I did briefly mull over this question, and then quickly short-circuited it when I set up the magic_link.user relationship.

Happy to quickly chat this over if you'd like? Or I can just give it a bit more of a think and tweak it.

app/common/auth/__init__.py
samuelhwilliams: I did _start_ doing this, then stopped. We could pass in the magic link object and read magic_link.user.email from it, and magic_link.expires_at_utc. But if we want to generate the two URLs then this method will only work if there's a corresponding flask app/request context, which feels like maybe a bit of state I wanted to keep out of here.

Again, only very lightly held, but that was the reason I didn't do this refactor yet.

app/common/auth/__init__.py
samuelhwilliams: Yeah interesting, didn't think of having `get_magic_link` also do a usable check. That might be v sensible you're right 👍 

I will consider this for a followup PR and hopefully not forget.

We will need to do a scan over all the interfaces soon-ish and think about error propagation as you say, so maybe we just do it in one pass later.

app/common/data/interfaces/magic_link.py
samuelhwilliams: I quite like the idea of retaining them all for now as a bit of audit trail. I think not overly worried about table size yet, but it's good to question that thought...

pyproject.toml
samuelhwilliams: nope 🫠  I started going down that path and then stopped when I hit the db timestamp issue. Let me rip it out.

sfount (approved):

PR #45, opened 2025-04-10
Add Flask-Talisman

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-341

Set it up with some (hopefully sensible; please validate) defaults, and then customise for local/pullpreview envs so things like flask-debugtoolbar work.

### DIFF
--- a/app/config/__init__.py
+++ b/app/config/__init__.py
@@ -1,3 +1,4 @@
+import copy
 import os
 from enum import Enum
 from typing import Any, Tuple, Type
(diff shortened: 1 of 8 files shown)


app/config/__init__.py
```diff
@@ -22,6 +24,49 @@ class DatabaseSecret(BaseModel):
...
+    "img-src": ["'self'"],
+    "style-src": ["'self'"],
+}
+
+
+def make_development_csp() -> dict[str, list[str]]:
+    csp = copy.deepcopy(FS_CONTENT_SECURITY_POLICY)
+    csp["default-src"].extend(
+        [
+            "http://localhost:5173",  # Vite assets
```
sfount, line 39: Can this come from `app.config["ASSETS_VITE_BASE_URL"]`? Maybe needs a bit of thinking about how it can sensible build it for http and ws but could always have a URL and WSS config 

Probably not a deal breaker but something that might be a pain to track down for a port change or something like that

sfount (approved):
Small suggestion but nothing blocking

PR #46, opened 2025-04-10
Auto-claim magic link via JS

samuelhwilliams:
We don't need users to click a button manually when using their magic link. The reason we add a button is to prevent email clients, which may preload the magic link URL for various reasons, from expiring the magic link. Email clients won't run JS on the page, so this shouldn't undermine the reason for us having this interstitial page.

### DIFF
--- a/app/common/templates/common/auth/claim_magic_link.html
+++ b/app/common/templates/common/auth/claim_magic_link.html
@@ -16,3 +16,7 @@
         </div>
     </div>
 {% endblock content %}
+
+{% block bodyEnd %}
+  <script nonce="{{ csp_nonce() }}">document.getElementById('submit').click();</script>
+{% endblock bodyEnd %}
--- a/app/common/templates/common/auth/claim_magic_link.html
+++ b/app/common/templates/common/auth/claim_magic_link.html
@@ -7,9 +7,7 @@
 {% block content %}
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
-            {# TODO: Add JS to auto submit this form #}
-
-            <form method="post" novalidate>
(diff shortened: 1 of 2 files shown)


app/common/templates/common/auth/claim_magic_link.html
```diff
@@ -16,3 +16,7 @@
         </div>
     </div>
 {% endblock content %}
+
+{% block bodyEnd %}
+  <script nonce="{{ csp_nonce() }}">document.getElementById('submit').click();</script>
```
sfount, line 21: I'll give this a go in a bit to get a sense for the page loading/ user experience but if this visually loads the page before the JavaScript runs I think we could try putting the js snippet in the `<head>` tag and having it either update the `window.locationl.href` or possibly create the form/ button and click it (to keep it POST-y) before the page has been fully fetched/ rendered.

sfount (approved):
Looks good! A nice balance of reliability for scrapers w/ good user experience 🏅

app/common/templates/common/auth/claim_magic_link.html
samuelhwilliams: https://github.com/alphagov/notifications-admin/blob/e5a228cc8e4b49f2e2b32162d6b34e22bb6d347b/app/templates/views/email-link-interstitial.html

similar

sfount (approved):

app/common/templates/common/auth/claim_magic_link.html
sfount: https://github.com/communitiesuk/funding-service/pull/46/commits/5b257f8f83b2ca89961d859bbae45107608cc112

MarcUsher (approved):

PR #49, opened 2025-04-11
Handle sign in and sign out with Flask-Session

samuelhwilliams:

## Summary
When a user claims their magic link, let's actually persist that login session by logging them in with Flask-Login. This adds data to their session. When they return in subsequent requests, we can load their user directly and thereby give them persistent access to the service.

We add a 'Sign out' link to the header if a user is logged in, and remove the 'Sign in' link; users will get directed to log in whenever they try to access a restricted page (not yet implemented).

The session is being stored as a client-side signed cookie for now, in line with [option 1 of the RFC](https://github.com/communitiesuk/funding-service-requests-for-comments/discussions/27). This is because Flask-Session does not yet support a Flask-SQLAlchemy-Lite extension (only the Flask-SQLAlchemy extension), so using server-side sessions stored in postgres is a bit of extra work. We'll try and get a fix upstream; if we do, we can switch to postgres sessions.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -5,11 +5,13 @@ from govuk_frontend_wtf.main import WTFormsHelpers
 from jinja2 import ChoiceLoader, PackageLoader, PrefixLoader
 
 from app import logging
+from app.common.data.models import User
 from app.config import get_settings
 from app.extensions import (
     auto_commit_after_request,
     db,
     flask_assets_vite,
+    login_manager,
     migrate,
     notification_service,
     talisman,
@@ -42,6 +44,11 @@ def create_app() -> Flask:
     toolbar.init_app(app)
     notification_service.init_app(app)
     talisman.init_app(app, **app.config["TALISMAN_SETTINGS"])
+    login_manager.init_app(app)
+
+    @login_manager.user_loader  # type: ignore[misc]
+    def load_user(user_id: str) -> User | None:
+        return db.session.get(User, user_id)
 
     # This section is needed for url_for("foo", _external=True) to
     # automatically generate http scheme when this sample is
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -5,6 +5,7 @@ from govuk_frontend_wtf.main import WTFormsHelpers
 from jinja2 import ChoiceLoader, PackageLoader, PrefixLoader
 
 from app import logging
+from app.common.data import interfaces
 from app.common.data.models import User
 from app.config import get_settings
 from app.extensions import (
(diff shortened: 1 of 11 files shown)


app/__init__.py
```diff
@@ -38,6 +47,11 @@ def create_app() -> Flask:
     toolbar.init_app(app)
     notification_service.init_app(app)
     talisman.init_app(app, **app.config["TALISMAN_SETTINGS"])
+    login_manager.init_app(app)
+
+    @login_manager.user_loader  # type: ignore[misc]
+    def load_user(user_id: str) -> User | None:
+        return db.session.get(User, user_id)
```
sfount, line 54: It's (almost definitely) a nit but could this call out to `interfaces.users` so that anything that can query the db sits in one place?

app/common/templates/common/base.html
```diff
@@ -19,7 +19,7 @@
     "productName": "MHCLG Funding Service",
     "homepageUrl": "#",
     "classes": "govuk-header--full-width-border app-!-no-wrap",
-    "navigation": [{ "text": ("Sign in"), "href": url_for('auth.request_a_link_to_sign_in') }],
+    "navigation": [{ "text": ("Sign out"), "href": url_for('auth.sign_out') }] if current_user.is_authenticated else [],
```
sfount, line 22: I think the proto/ design I last saw had a border but this could all be cleaned up/ aligned at the end?

Is current_user always set because flask login puts something on the global context or should this be checking it exists before trying to access `is_authenticated`? (Jinja might do that safely anyway, not sure!)

tests/integration/common/auth/test_auth.py
```diff
@@ -101,3 +104,24 @@ def test_post_claims_link_and_redirects(self, client, factories):
...
+        # the user in the session
+        client.post(url_for("auth.claim_magic_link", magic_link_code=magic_link.code), json={"submit": "yes"})
+        with client.session_transaction() as session:
+            assert "_user_id" in session
+
+        response = client.get(url_for("auth.sign_out"), follow_redirects=True)
+        assert response.status_code == 200
+
+        with client.session_transaction() as session:
+            assert "_user_id" not in session
```
sfount, line 127: Nice 👍 feels OK to do this setting it up to claim it initially in the context of magic links + sessions (although I'm sure future tests will want a user fixture managed on the session somewhere)

sfount (approved):

app/__init__.py
samuelhwilliams: oh - yeah - good catch

app/common/templates/common/base.html
samuelhwilliams: yep current_user is always available, even before logging in

sfount (approved):

PR #70, opened 2025-04-11
BAU Constrain Postgres versions for tests and CI

sfount:
Our workflows/ CI/ test environment should aim to reflect the deployed environments as much as possible to give us confidence that automated checks are relevant to what will then be run.

This attempts to tell renovate to give us pinned patch upgrades but keep the major/ minor version of postgres the same as our deployed version, this should then be bumped when minor/ major upgrades are made (this needs to be sense checked against our DB upgrade policy, I believe its constrained to minor versions but if there's an automatic upgrade policy the rules here may need to change).

I'm not confident in the renovate syntax proposed here that may take some trial and error.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dev = [
     "responses==0.25.7",
     "ruff==0.11.5",
     "sqlalchemy-utils==0.41.2",
-    "testcontainers==4.10.0",
+    "testcontainers[postgres]==4.10.0",
     "types-flask-migrate==4.1.0.20250112",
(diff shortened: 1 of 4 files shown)


samuelhwilliams (commented):
I feel like this looks sensible. I think almost inevitable that it gets out of sync with our AWS DB, but better than leaving it as-is.

May require some tweaking/trial-and-erroring esp the renovate bits, so let's keep an eye, but it feels good.

samuelhwilliams (approved):

samuelhwilliams (approved):

PR #73, opened 2025-04-14
Redirect to `next` when claiming magic link

samuelhwilliams:

## Summary
The 'request a magic link' endpoint now knows about a `next` entry in the user's session, which will get validated+sanitised and then stored in on the MagicLink record. When the magic link is claimed, we'll redirect the user to that URL, so that they can pick up where they were when we made them sign in.

Another ticket will add the decorator and enforce auth on all the pages to make use of this functionality.

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -1,7 +1,7 @@
 import uuid
 from typing import cast
 
-from flask import Blueprint, abort, redirect, render_template, url_for
+from flask import Blueprint, abort, redirect, render_template, request, url_for
 from flask.typing import ResponseReturnValue
 from flask_login import login_user, logout_user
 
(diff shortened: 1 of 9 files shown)


app/common/auth/__init__.py
```diff
@@ -20,15 +20,13 @@
 @auth_blueprint.route("/request-a-link-to-sign-in", methods=["GET", "POST"])
 @auto_commit_after_request
 def request_a_link_to_sign_in() -> ResponseReturnValue:
-    form = SignInForm()
+    form = SignInForm(data={"redirect_to": request.args.get("next", url_for("index"))})
```
sfount, line 23: Does putting `next` in the tamper-proof session which is set by the handler that realises you don't have authentication leave a bit less scope for users accidentally or maliciously tampering with this? 

Looking over the method to sanitise the input I feel like it wouldn't be an issue if it was changed but good to understand

sfount (approved):
Code looks good and really nice coverage 👍 

Question on if we want the onward URL to be exposed to users at all but happy to either chat it through or leave it

app/common/auth/__init__.py
samuelhwilliams: That feels like a v sensible shout, let's do that 👍  (caveat: we don't want to bung everything into the session, but this feels fine)

sfount (approved):
Nice! Putting santise redirect URL in a security/util feels clean too 👌 

Question on if `next` is cleaned up after one creation of the magic link (here in the request a link to sign in handler) or if that happens somewhere else in the flow but that feels like it will come out of the next ticket/ PR described

PR #74, opened 2025-04-14
Require authentication for grant management views

samuelhwilliams:
A very basic decorator to require a logged-in communities.gov.uk user when accessing grant management pages.

This will need a lot of extending and further development when we come to do multiple user types, but for now this is sufficient.

### DIFF
--- a/app/common/auth/decorators.py
+++ b/app/common/auth/decorators.py
@@ -1,0 +1,37 @@
+import functools
+from typing import Callable, cast
+
+from flask import abort, redirect, request, session, url_for
+from flask.typing import ResponseReturnValue
+from flask_login import current_user
+
+from app.common.data.models import User
+
+
+def login_required[**P](
+    func: Callable[P, ResponseReturnValue],
+) -> Callable[P, ResponseReturnValue]:
+    @functools.wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> ResponseReturnValue:
+        if not current_user.is_authenticated:
+            session["next"] = request.full_path
+            return redirect(url_for("auth.request_a_link_to_sign_in"))
+
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def mhclg_login_required[**P](
+    func: Callable[P, ResponseReturnValue],
+) -> Callable[P, ResponseReturnValue]:
+    @functools.wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> ResponseReturnValue:
+        # This decorator is itself wrapped by `login_required`, so we know that `current_user` exists and is
+        # not an anonymous user (ie a user is definitely logged-in) if we get here.
+        if not cast(User, current_user).email.endswith("@communities.gov.uk"):
+            abort(403)
+
+        return func(*args, **kwargs)
+
+    return login_required(wrapper)
--- a/tests/e2e/config.py
+++ b/tests/e2e/config.py
@@ -1,0 +1,65 @@
+import json
+import re
+import subprocess
+from typing import Literal, Protocol, cast
+
+import boto3
+
+
+class EndToEndTestSecrets(Protocol):
+    @property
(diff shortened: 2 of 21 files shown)


tests/integration/conftest.py
```diff
@@ -183,3 +184,17 @@ def _track_notification(*args, **kwargs):  # type: ignore
     )
 
     yield calls
+
+
+@pytest.fixture()
+def authenticated_client(
+    anonymous_client: FlaskClient, factories: _Factories, request: FixtureRequest
+) -> Generator[FlaskClient, None, None]:
+    email_mark = request.node.get_closest_marker("authenticate_as")
```
sfount, line 193: Splitting out the user who will be authenticated as a marker feels nice as permissions and options grow!

app/common/auth/decorators.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def mhclg_login_required[**P](
+    func: Callable[P, ResponseReturnValue],
+) -> Callable[P, ResponseReturnValue]:
+    @functools.wraps(func)
+    def wrapper(*args: P.args, **kwargs: P.kwargs) -> ResponseReturnValue:
+        # This decorator is itself wrapped by `login_required`, so we know that `current_user` exists and is
+        # not an anonymous user (ie a user is definitely logged-in) if we get here.
+        if not cast(User, current_user).email.endswith("@communities.gov.uk"):
```
sfount, line 32: Something about having this wrapper doing this fixed check independently of sign up which will also be checking this feels like it might lead to behavior thats hard to track down?

I think this only seems confusing because there aren't permissions/ scope to check (i.e "Access to grant", "Update grant", "Read grant") but once those are in we should be able to trust that for them to have those permissions they would have had good reason (i.e had the appropriate email domain for access)

I don't think this needs to change this PR as this is basically a single combined implementation of the authentication and permissions that don't otherwise exist! Just prompted starting to think it through

sfount (approved):

app/common/auth/decorators.py
samuelhwilliams: Yep I think a good call out.

I'm fairly sure this will all come out in the wash when we start adding orgs and permissions properly, so confident to leave it in for now.

Fingers crossed I won't regret this arrogance.

sfount (commented):
Ah if this is one of the last PRs on this work  (very happy for it to be separate) two things it feels like it would be helpful to include: 
- remove the `next` property from the session when the user successfully authorises OR when the magic link is actually created (and sets the redirect in the DB) -- setting precedent for anticipating cleaning up any use of the flask sessino
- removing "Grants" from the sub-nav of the non-authorised view, for now you should see and be able to interact with nothing until you're authorised 

Something to consider if we should do it now or not might be to have a decorator or similar that forwards you from sign-in pages/ flows if you do happen to already be authorised so that we avoid unusual half states as much as possible

tests/e2e/config.py
```diff
@@ -0,0 +1,65 @@
...
+
+class EndToEndTestSecrets(Protocol):
+    @property
+    def GOVUK_NOTIFY_API_KEY(self) -> str: ...
+
+
+class LocalEndToEndSecrets:
+    @property
+    def GOVUK_NOTIFY_API_KEY(self) -> str:
+        with open(".env") as env_file:
```
sfount, line 17: I wonder if we want to use pydantic-settings to try and keep this consistent everywhere, I'm not too particular about it

sfount (approved):
Looks reasonable to me, I've not thoroughly read every line of the Notify access code as I know it comes from other places that it works and I'm sure its been tried! 

Good to consider how we'll expect this to work with SSO in end to end tests too, I guess following this principle there could be a test account that could actually log in generating its own OTP or we can consider setting up the session for tests using the environments secrets

tests/e2e/config.py
samuelhwilliams: interesting thought - may revisit 🤔

PR #75, opened 2025-04-15
Enable experimental logging to sentry

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-376

Sentry is exploring integrating with logging so that errors, traces and logs can all be seen through a single pane of glass.

The instrumentation is likely to be basic, but we've had a fairly poor logging story on Funding Service for a while, so if we can easily turn on a single integration and get some basic integrated logging in place, this feels like a quick win worth testing.

https://github.com/getsentry/sentry-python/discussions/4220

If we decide to turn this off later, we won't strictly be any worse off.

### DIFF
--- a/app/sentry.py
+++ b/app/sentry.py
@@ -34,7 +34,7 @@ def init_sentry() -> None:
         env = Environment(os.getenv("FLASK_ENV", Environment.PROD.value))
 
         sentry_sdk.init(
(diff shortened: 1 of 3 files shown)


gidsg (approved):
Nice to try this 🥼 👍

PR #78, opened 2025-04-15
Don't force HTTPS on all requests through Talisman

samuelhwilliams:
Our deployed environments are currently failing because the healthcheck won't go healthy. My hypothesis is that AppRunner is doing the healthcheck over HTTP locally, and Talisman is trying to force it to be over HTTPS, so it's never succeeding.

We can enforce HTTPS through other methods, so don't need talisman to do this for us (eg cloudfront, HSTS preload, etc).


MarcUsher (approved):
Let's see if it works!

MarcUsher (approved):

PR #80, opened 2025-04-15
Change docker login method for github actions

samuelhwilliams:
The docker CLI throws this lovely warning:

WARNING! Your password will be stored unencrypted in /home/runner/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -59,6 +59,11 @@ jobs:
           role-session-name: "funding-service_build_${{ steps.currentdatetime.outputs.datetime }}"
           aws-region: eu-west-2
 
+      - name: Login to ECR
+        uses: docker/login-action@v3
+        with:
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Test workflow run: https://github.com/communitiesuk/funding-service/actions/runs/14471734909/job/40587188779

It failed, but the ECR login succeeded so I think this is all good. ECR just doesn't like this branch name having a `/` in it.

MarcUsher:
> Test workflow run: https://github.com/communitiesuk/funding-service/actions/runs/14471734909/job/40587188779
> 
> It failed, but the ECR login succeeded so I think this is all good. ECR just doesn't like this branch name having a `/` in it.

Yeah I've tidied this up a bit in my PR so we're appending the sha rather than the branch name to avoid this error and so we're specific about the images

PR #81, opened 2025-04-15
FSPT-377: Make AppRunner deploy manual

MarcUsher:
We want visibility and control over the AppRunner deployments rather than relying on automatic deployments when a new image with the `latest` tag is pushed to ECR.

We're not sure if we'll continue to use AppRunner so for the time being we can use the AWS AppRunner deploy action and specify deploying the image we've just built in the pipeline. This allows us to then run the e2e tests after a successful deploy or catch any issues with deployments (which before now could fail silently). We're also now tagging images with the sha as well as `latest`, as the pipeline now only runs on merges to main.

### But also multiple envs now - what to do with ECR?

With multiple deployed environments now we run into the issue that each AWS account has its own ECR. We only want to build our images once and deploy this same image to each environment in turn, which becomes tricky when we push the built image to the ECR of the lowest environment first.

The team discussed allowing each account to have access to the ECR of the account/environment beneath it, but this would require updates to the terraform setup and injecting unique account IDs conditionally (as you'd only want test to be able to access dev, and prod to access test etc.), and feels like it could run into some rough edges.

We discussed having a separate AWS CI account which would host all the built images and then our pipeline would pull the image to that environment's ECR and deploy it (or deploy it directly from that AWS CI ECR). While we wait to have a separate AWS CI account set up, we agreed to use GitHub Container Registry as an interim solution, as it's what we already use in our existing services.

### Deprecation Warning ignored

The additional ignored `DeprecationWarning` is ignoring a warning from an implementation of a method in the botocore library, so not something we have any control over (see [failed run](https://github.com/communitiesuk/funding-service/actions/runs/14493055489/job/40653795696) for context)
```
self = <botocore.auth.SigV4Auth object at 0x7fc573fa7ed0>
request = <botocore.awsrequest.AWSRequest object at 0x7fc573fa65d0>

    def add_auth(self, request):
        if self.credentials is None:
            raise NoCredentialsError()
>       datetime_now = datetime.datetime.utcnow()
E       DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
```

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -16,16 +16,16 @@ on:
       TEMP_SLACK_CHANNEL_ID:
         required: true
     outputs:
-      image_location:
+      ghcr_image_location:
         description: "A URI pointing to the paketo-built image"
-        value: ${{ jobs.paketo_build.outputs.image_location }}
+        value: ${{ jobs.paketo_build.outputs.ghcr_image_location }}
 
 jobs:
   paketo_build:
     runs-on: ubuntu-latest
     name: Paketo
     outputs:
-      image_location: ${{ steps.build_and_publish.outputs.image_location }}
+      ghcr_image_location: ${{ steps.build_and_publish.outputs.ghcr_image_location }}
 
     steps:
       - uses: buildpacks/github-actions/setup-pack@d82294e03fd6b8e8484cde2ee166788876e366e3 # v5.8.10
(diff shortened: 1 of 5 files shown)


.github/workflows/package.yml
```diff
@@ -68,14 +70,20 @@ jobs:
         id: build_and_publish
         run: |
           IMAGE_ID=${{ secrets.AWS_ACCOUNT }}.dkr.ecr.eu-west-2.amazonaws.com/dev-funding-service
```
samuelhwilliams, line 72:
```suggestion
          IMAGE_ID=${{ secrets.AWS_ACCOUNT }}.dkr.ecr.eu-west-2.amazonaws.com/dev-funding-service
```

The ECR name needs to be scoped to the environment, I think?

pyproject.toml
```diff
@@ -113,9 +113,9 @@ markers = [
     "authenticate_as: Email address to use for `authenticated_client` fixture; default `test@communities.gov.uk`"
 ]
 
-
 filterwarnings = [
     "error",
     "ignore:Could not insert debug toolbar:UserWarning",
-    "ignore:cannot collect test class 'TestConfig' because it has a __init__ constructor:pytest.PytestCollectionWarning"
+    "ignore:cannot collect test class 'TestConfig' because it has a __init__ constructor:pytest.PytestCollectionWarning",
+    "ignore:datetime\\.datetime\\.utcnow\\(\\) is deprecated and scheduled for removal in a future version*:DeprecationWarning"
```
samuelhwilliams, line 120: A bit surprised that these are needed for this PR? 🤔

MarcUsher: Me too but kept getting failed e2e test runs when testing because of this error (I think we treat deprecation warnings as failures in the test setup currently?)

samuelhwilliams (approved):
Nice write up 🙇‍♂️

PR #91, opened 2025-04-15
FSPT-379 End to end tests authorise once across distributed workers

sfount:
This proposes setting up a file lock in the user auth fixture for end to end tests. This allows any number of workers to spin up and re-use a session which is correctly authorised once.

Currently all end to end tests that pull in `user_auth` will create a new random user in the system and take the time to go through the magic link process individually.

This change means even if running pytest with `-n auto` the same authorised session can be used across tests.

Note that this doesn't guard against the scenario when you're not running in parallel mode as the fixture is function scoped which would result in all tests authorising individually - this behaviour doesn't feel quite right so we should interrogate if we can reproduce this behaviour without manipulating the individual `page` fixture (it feels like its reproducing pytests session fixture caching rather than working with it).

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -53,7 +53,7 @@ jobs:
       - name: install dependencies
         run: uv sync --frozen
       - name: run unit tests
-        run: uv run pytest -nauto .
+        run: uv run pytest -n auto .
         env:
           FLASK_ENV: unit_test
           DATABASE_HOST: localhost
--- a/.github/workflows/run_e2e_tests.yml
+++ b/.github/workflows/run_e2e_tests.yml
@@ -60,7 +60,7 @@ jobs:
           aws-region: eu-west-2
 
       - name: run e2e tests
-        run: uv run --frozen pytest --e2e --e2e-env ${{ inputs.environment }} --tracing=retain-on-failure --browser chromium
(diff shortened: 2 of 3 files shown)


tests/e2e/helpers.py
```diff
@@ -28,10 +29,13 @@ def extract_email_link(email: dict[str, str]) -> str:
...
-    for email in emails:
-        if email["email_address"] == email_address:
-            print(email)
-            return extract_email_link(email)
+    # it looks like Notify is eventually consistent and this can fail when I'm running it locally
+    for _ in range(3):
+        emails = client.get_all_notifications(template_type="email", status="delivered")["notifications"]  # type: ignore[no-untyped-call]
+        for email in emails:
+            if email["email_address"] == email_address:
+                print(email)
```
samuelhwilliams, line 37: not from you but ❌

sfount:
just for myself in case I need to link back: it should be noted that parallelising tests while there are so few will actually take longer than just running them synchronously (spinning up workers, etc). 

I think this is a reasonable trade off though for having CI check that tests are reasonably isolated and it should catch failures that you might not get running pytest locally - there's an example of surfacing something useful here https://github.com/communitiesuk/funding-service/pull/118

samuelhwilliams:
Something weird going on here - the tests aren't actually running at the same time if you have multiple workers (eg `pytest --e2e --headed -n2`)? 🤔

sfount:
Closing this until we get to the bottom of why the `main` branch isn't running in parallel even with `-n 2` (this functionality should then be added making sure that doesn't regress)

PR #94, opened 2025-04-16
wip remove server name

samuelhwilliams:


sfount (approved):
Happy to approve this to get us and and running while its dev and domains are in flux, I think its worth recording it as a decision if we want to stick with it -- is there any spoofing of host headers that could cause it to behave in ways we haven't through etc. if the app isn't in control of this etc.

PR #95, opened 2025-04-16
Add ProxyFix config for AWS environments

samuelhwilliams:
We need to tell our app that it has a proxy in front of it, which will be inserting its own (eg) X-Forwarded-Host headers and overriding the initial Host header.

ProxyFix will pop off CloudFront's host header and then trust the next one in the stack, which is our intended 'external' domain. Otherwise Flask will always think it's being served on the AppRunner domain, and therefore build URLs that point at AppRunner instead of CloudFront.


samuelhwilliams:
Closing this for now as we will ignore the existence of cloudfront until we get proper domains in place

PR #102, opened 2025-04-17
Disable CI runs on main branch

samuelhwilliams:
We currently require branches to be up-to-date with main, so running CI again after merge is redundant.


MarcUsher (approved):
I approve of less unnecessary noise

PR #115, opened 2025-04-22
FSPT-379 E2E lookup magic links by reference

sfount:
Adds an ephemeral reference to the notification sent during the magic link auth to the page using a `[data-x]` attribute.

Remove the logic where we generate random unique email addresses to then be able to look them up in Notify. Instead directly tie the notification that was sent in that session to the page checking the email.

This will now let us use one email address that actually sends emails, while we're running end to end tests against actual environments this allows us to use an API key that sends email allowing pre-production users to not log into the Notify admin console.

The method should also work without a problem for test Notify keys that pretend to send emails.

We're making the choice to only include the `[data-x]` attribute if we're not in a production Flask environment but that can be removed if necessary - there should be nothing sensitive about the emails reference.

A follow up PR will then switch the fixture providing `user_auth` for end to end tests to share a single session rather than uniquely authenticating every test.

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,4 +1,4 @@
-from typing import Generator
+from typing import Generator, cast
 
 import pytest
 from playwright.sync_api import BrowserContext, Page
@@ -7,7 +7,7 @@ from pytest_playwright import CreateContextCallback
 
 from tests.e2e.config import AWSEndToEndSecrets, EndToEndTestSecrets, LocalEndToEndSecrets
 from tests.e2e.dataclasses import E2ETestUser
-from tests.e2e.helpers import generate_email_address, retrieve_magic_link
+from tests.e2e.helpers import retrieve_magic_link
 from tests.e2e.pages import RequestALinkToSignInPage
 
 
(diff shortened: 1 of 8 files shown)


tests/conftest.py
```diff
@@ -37,6 +37,8 @@ def pytest_addoption(parser: Parser) -> None:
         help="Change the viewport size of the browser window used for playwright tests (default: 1920x1080)",
     )
 
+    parser.addoption("--authenticate_as", action="store", help="Email address to authenticate as in e2e tests")
```
samuelhwilliams, line 40:
```suggestion
    parser.addoption("--e2e-authenticate-as", action="store", help="Email address to authenticate as in e2e tests")
```

tests/e2e/conftest.py
```diff
@@ -52,26 +53,29 @@ def e2e_test_secrets(request: FixtureRequest) -> EndToEndTestSecrets:
     raise ValueError(f"Unknown e2e_env: {e2e_env}.")
 
 
+@pytest.fixture
+def email(request: FixtureRequest) -> str:
+    email_env = os.getenv("E2E_GOVUK_NOTIFY_AUTHENTICATE_AS")
```
samuelhwilliams, line 58:
```suggestion
    email_env = os.getenv("E2E_AUTHENTICATE_AS")
```

probably nice to keep the env var and the option in sync name-wise.

I'm also a bit 🤷‍♂️  on introducing env vars to toggle these things, esp if we're doing it only partially, but not going to block.

samuelhwilliams: Also, this option strikes me more as one that might be configured per-test rather than per-session? Esp once we have multiple user types. This was supported previously with the user_auth fixture and the `user_domain` marker that we're now ripping out. I guess it's fine in the short term but I feel like I'll need to add it back in again soon 🤷

tests/e2e/conftest.py
```diff
@@ -52,26 +53,29 @@ def e2e_test_secrets(request: FixtureRequest) -> EndToEndTestSecrets:
...
+    email_option = request.config.getoption("authenticate_as")
+    email = email_option or email_env
+
+    if not email:
+        raise ValueError("Set E2E_GOVUK_NOTIFY_AUTHENTICATE_AS in the environment or use --authenticate_as")
```
samuelhwilliams, line 63: Controversial: if we're going this route, just put the email address in here as a default value? Yes we'd be leaking it, but does that matter (genuine question, not dismissive)? Would rather as few things as possible need configuring/specifying to run the e2e tests.

tests/e2e/conftest.py
sfount: 👍 yeah agree with all of this, based on your other thoughts I think we'll just remove it in favour of a mark and a default value which would encourage tests to override it if needed

I think it will work fine to be able to have different auth'd sessions per test (and should all work nicely with keeping one session per run if that fixture uses the mark as well) but I think we can keep an eye on and hopefully keep end to end tests to stick to as few users as possible and really cover user types & permissions in detail in integration tests

PR #118, opened 2025-04-23
Add basic auth support

samuelhwilliams:
We'll use in this in our non-prod environments to limit the potential for scraping/unintended access.

Once everything's behind CloudFront, and the app isn't publicly accessible, we should be able to do this through a cloudfront edge function or similar, rather than app code. But for now this is fine.


tests/utils.py
```diff
@@ -86,3 +88,22 @@ def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self._regex})"
+
+
+def build_db_config(setup_db_container: PostgresContainer | None) -> Dict[str, Any]:
+    if setup_db_container is None:
```
sfount, line 94: Nice this feels like a clean abstraction 👌

sfount (approved):
Have sense checked with docker compose locally and functionality looks good 👍 

Excellent chasing down of fixtures/ patching, that would have gotten a lot gnarlier further down the line. This has made me think of a few places I should go clean up tests that wouldn't stand up to multiple random runs -- good to keep on top of this from the beginning 🙇

sfount (approved):

PR #124, opened 2025-04-25
FSPT-341 Format HTML 

sfount:
This commit adds the prettier formatter to the node package manager
dependencies. This configuration for formatting jinja templates is the
only one we've found so far that feels like it does a consistent job.

As we're not yet convinced this commit doesn't include it in pre-commit
or CI breaking checks but just makes it available to developers to use
if they're working on templates and would like opinions.

This can be run with:
```
make format-html
```

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -75,6 +75,23 @@ jobs:
       - name: run type checking
         run: uv run mypy
 
+  lint_html_formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: Read .nvmrc
+        id: nvmrc
+        run: echo "nodeVersion=$(cat .nvmrc)" >> "$GITHUB_OUTPUT"
+      - name: Setup node ${{ steps.nvmrc.outputs.nodeVersion }}
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: ${{ steps.nvmrc.outputs.nodeVersion }}
+      - name: install dependencies
+        run: npm ci
+      - name: check HTML formatting
+        run: make check-html
+
 
   lint_imports:
     runs-on: ubuntu-latest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ci:
   autofix_prs: false
-  skip: ["mypy"]
+  skip: ["mypy", "format-html"]
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
@@ -29,6 +29,11 @@ repos:
           ]
         exclude: tests/pre_award/keys/rsa256
 
+  - repo: https://github.com/seddonym/import-linter
+    rev: v2.3
+    hooks:
+      - id: import-linter
+
   - repo: local
     hooks:
       - id: mypy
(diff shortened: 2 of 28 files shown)


samuelhwilliams:
I think it's probably worth enforcing this somehow - maybe we can go with doing it in CI only, but no pre-commit yet? 👀 I think 100%-optional is maybe not the right line to draw?

samuelhwilliams (approved):
LGTM

.github/workflows/build_and_test.yml
```diff
@@ -75,6 +75,23 @@ jobs:
...
+        id: nvmrc
+        run: echo "nodeVersion=$(cat .nvmrc)" >> "$GITHUB_OUTPUT"
+      - name: Setup node ${{ steps.nvmrc.outputs.nodeVersion }}
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: ${{ steps.nvmrc.outputs.nodeVersion }}
+      - name: install dependencies
+        run: npm ci
+      - name: check HTML formatting
+        run: make format-html
```
sfount, line 93: I think `--write` here will still return 0 if it comes across things it can fix (after dumping the files into an ephemeral void), we probably just want to run `--check`

I can double check the behviour as I've forgotten

sfount (commented):
Feeling like a good balance of checks + not too much tooling, nice investigation!

.github/workflows/build_and_test.yml
samuelhwilliams: yeah I think you're right - trying to work out how to get a 1 now 👀

samuelhwilliams: Fixed in 36911594cd40c8ac7c17b9d4d63210c3531defd2

app/common/templates/common/auth/sign_in.html
wjrm500, line 1: Addition of this file out of scope of the commit at least, if not the PR?

wjrm500 (commented):
When I bring the code in locally and run `make up` I end up with this:

image

Then I tried `make bootstrap` as figured that might do something, but I still end up with the same thing in the browser

Makefile
```diff
@@ -20,7 +20,11 @@ vite:
 	npm install
 	npm run build
 
-.PHONY: format-html
+.PHONY: check-html
+check-html:
+	npx prettier --plugin=prettier-plugin-jinja-template --parser=jinja-template --tab-width=2 --html-whitespace-sensitivity ignore --bracket-same-line=true --print-width=240 --check **/*.html
+
+.iHONY: format-html
```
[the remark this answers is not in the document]
samuelhwilliams: .iHONY is a dumb typo, thanks Marc 👀

.pre-commit-config.yaml
```diff
@@ -37,8 +42,12 @@ repos:
         types: [python]
         entry: uv run mypy
         pass_filenames: false
-
-  - repo: https://github.com/seddonym/import-linter
-    rev: v2.3
-    hooks:
-      - id: import-linter
+      - id: format-html
```
MarcUsher, line 45: One general question for me - here in pre-commit we're running the `format-html` command, which will fix any discrepancies it comes across, whereas in CI we're running `check-html` which just outputs if there are any discrepancies. 

This makes sense to me, but just curious what situation we'd foresee where anything would be caught by the `check-html` which hadn't been caught and fixed by the `format-html`?

MarcUsher (approved):
Approving, this is super helpful to have! Thanks both! Just leaving one last question for my own thinking.

.pre-commit-config.yaml
samuelhwilliams: pre-commit is to make developers lives better, and is in user world so you can't trust it's actually happened (git commit --no-verify, devs without precommit hooks installed etc)

CI is for actually making sure your standards are being met. It shouldn't catch anything not caught by format-html, but it catches situations where format-html hasn't been run for some reason.

PR #126, opened 2025-04-28
FSPT-363: production server

gidsg:
Switches to use gunicorn server on deployed envs.

Default config is based on that used by pre-award


samuelhwilliams:
I think we're missing some monkeypatching to actually make the gevent stuff work: 96a0c5b7e9f637f35d597b115f66db59d843aa72.

Will need to chat to understand how to reproduce your error case around AsyncToSync 👀

PR #129, opened 2025-04-29
FSPT-363: production server

gidsg:
Switches to use gunicorn server on deployed envs.

Default config is based on that used by pre-award

Note have tested using Pull Preview and locally but difficult to test on main pipeline as we don't have an easy way to deploy branches to the environment currently.


.github/workflows/deploy.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           image: ${{ steps.push_image.outputs.ecr_image_location }}
           access-role-arn: arn:aws:iam::${{ secrets.AWS_ACCOUNT }}:role/${{ inputs.environment }}-funding-service-app-runner-build-role
           region: eu-west-2
-          port: 5000
+          port: 8080
```
samuelhwilliams, line 111: Probably want to update terraform to point at 8080 as well, just to keep everything in sync.

Tiny49 (approved):
LGTM

PR #140, opened 2025-05-01
FSPT-388: Implement SSO for platform admins

MarcUsher:
Part of [FSPT-388](https://mhclgdigital.atlassian.net/browse/FSPT-388)

This is the initial work and scaffolding to allow SSO login to the new funding service platform for MHCLG users. 

The service uses the same app in Azure as the pre-award and post-award services, so we have added the redirect URI to match the pattern of those that are already registered in the Azure portal, with `/sso/get-token`.

In order to test the full SSO sign-in journey with end to end tests we need to be able to consistently use or replicate the SSO sign-in journey. For local e2e tests, we are doing this by implementing a stub service which implements the same SSO interface and flow as Microsoft and "returns" an authenticated user, with an interstitial "enter your email" page so that we can be confident that redirects are working as expected.

This is currently only used locally, as the SSO e2e test will need to be implemented differently for the deployed environments which will send users to the real Microsoft login page, for which we are looking into having a test user provisioned by IT that we can use to do 'real' logins during some browser-based flow.

Work still do to on SSO:
- Database changes to include a Role table and persisting roles after sign-in for MHCLG users in the `FSD_ADMINS group` (for now, any user with an MHCLG email address can log in and access the platform)
- Update e2e test to work on deployed environments with an actual MHCLG test account
- Skipping sign in pages if logged in
- Tidying up journeys and Auth choices

[FSPT-388]: https://mhclgdigital.atlassian.net/browse/FSPT-388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/config/__init__.py
```diff
@@ -228,6 +228,25 @@ def SQLALCHEMY_ENGINES(self) -> dict[str, str]:
...
+        return "https://login.microsoftonline.com/" + self.AZURE_AD_TENANT_ID
+
+    # The absolute URL must match the redirect URI you set
+    # in the app's registration in the Azure portal.
+    @property
+    def AZURE_AD_REDIRECT_URI(self) -> str:
+        # This obvs doesn't work right now for deployed envs - on the old apps we set 'AUTHENTICATOR_HOST' everywhere
+        # for each environment but as this is one app with one domain/host I'm not sure if we need to follow
+        # that pattern? And if not where we'd get the host value from for dev/test/prod.
+        return urllib.parse.urljoin("https://funding.communities.gov.localhost:8080", "/sso/get-token")
```
gidsg, line 248: Can we use `url_for` with `external=True`? This should generate an absolute URL for the current app.

stubs/sso/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+        "token_source": "identity_provider",
+    }
+
+
+@app.route("/<path:arbitrary_name>", methods=["GET", "POST"])
+def arbitrary_path(arbitrary_name):
+    print(request.data)
+    print(request.args)
+    print(request.path)
+    print(arbitrary_name)
```
samuelhwilliams, line 150: we can defo drop this now ha

stubs/sso/__init__.py
```diff
@@ -0,0 +1,150 @@
+import base64
```
samuelhwilliams, line 1: We probably want a decent comment at the top of this file, or a README for the sso stub directory, explaining what this is for and why

stubs/sso/__init__.py
MarcUsher: Oh yeah thanks good catch 😆

tests/e2e/pages.py
```diff
@@ -47,6 +47,34 @@ def click_request_a_link(self) -> None:
...
+
+    def navigate(self) -> None:
+        self.page.goto(f"{self.domain}/sso/sign-in")
+        expect(self.title).to_be_visible()
+
+    def click_sign_in(self) -> None:
+        self.sign_in.click()
+
+
+class SSOEmailLoginPage(BasePage):
```
samuelhwilliams, line 64: Maybe StubSSOEmailLoginPage?

app/config/__init__.py
```diff
@@ -246,6 +260,19 @@ class LocalConfig(_SharedConfig):
...
+    # can run correctly in CI but these should be overwritten in your local .env file with real values
+    # if you want to sign in with SSO locally
+
+    # Azure Active Directory Config
+    AZURE_AD_CLIENT_ID: str = "00000000-0000-0000-0000-000000000000"
+    AZURE_AD_CLIENT_SECRET: str = "incorrect_value"
+    AZURE_AD_TENANT_ID: str = "00000000-0000-0000-0000-000000000000"
+
+    @property
+    def AZURE_AD_AUTHORITY(self) -> str:
```
samuelhwilliams, line 273: Can this be overridden by setting an env var? Can we document how devs can use real Microsoft sso locally?

docker-compose.yml
```diff
@@ -42,11 +42,36 @@ services:
...
+    ports:
+      - "4005:4005" # application
+    stdin_open: true
+    tty: true
+    command: >
+      bash -c "
+      cp /app/certs/rootCA.pem /usr/local/share/ca-certificates/rootCA.crt && \
+      update-ca-certificates
+
+      uv run flask run --host 0.0.0.0 --port 4005 --reload --debug --cert=/app/certs/cert.pem --key=/app/certs/key.pem
```
samuelhwilliams, line 66: Did we sort hot reloading of the SSO stub server? If not maybe let's turn it off.

tests/e2e/pages.py
MarcUsher: Yeah good shout, helpful to differentiate it from anything on the main app

docker-compose.yml
MarcUsher: Yeah it was still aggressively reloading and breaking as I was working on it, I'll turn it off (won't be hugely used outside of this work so not the end of the world to not have it/easy to turn it back on)

app/config/__init__.py
MarcUsher: Thanks @gidsg, this ended up getting moved out of the config to the `sso` module in auth but following this pattern you suggested worked nicely 🙏

tests/integration/common/auth/test_auth.py
```diff
@@ -163,3 +163,10 @@ def test_get(self, anonymous_client, factories):
 
         with anonymous_client.session_transaction() as session:
             assert "_user_id" not in session
+
+
+class TestSSOSignInView:
+    def test_get(self, anonymous_client):
+        response = anonymous_client.get(url_for("auth.sso_sign_in"))
+        assert response.status_code == 200
+        assert b"Connected and consistent set of funding services" in response.data
```
samuelhwilliams, line 172: We could add a 🤏  test for the POST endpoint with `follow_redirects=False` to assert that the redirect goes to our configured redirect place?

stubs/sso/__init__.py
```diff
@@ -0,0 +1,156 @@
+"""
+This stub server is set up to implement the same interface as Microsoft SSO flow for local development and locally-run
+end to end tests.
```
samuelhwilliams, line 3:
```suggestion
This stub server is set up to (minimally) implement an oauth interface flow, as used by MIcrosoft AD, for local development and locally-run
end to end tests.
```

tests/conftest.py
```diff
@@ -43,10 +43,19 @@ def pytest_collection_modifyitems(config: Config, items: list[Any]) -> None:
     # If e2e tests are requested, skips everything not marked as e2e
     skip_e2e = pytest.mark.skip(reason="only running non-e2e tests")
     skip_non_e2e = pytest.mark.skip(reason="only running e2e tests")
+    skip_e2e_environment = pytest.mark.skip(reason="only run this e2e test in a specific environment")
```
samuelhwilliams, line 46: this is the reason shown within a specific pytest run, rather than the reason the market exists, so maybe a small tweak

```suggestion
    skip_e2e_environment = pytest.mark.skip(reason="test is configured not to run in this e2e environment")
```

app/config/__init__.py
```diff
@@ -246,6 +261,16 @@ class LocalConfig(_SharedConfig):
...
 
+    # The LocalConfig needs default values for these AZURE_AD variables so that the Check DB Migrations job
+    # can run correctly in CI but these should be overwritten in your local .env file with real values
+    # and setting AZURE_AD_LOGIN_DOMAIN to https://login.microsoftonline.com/ if you want to sign in with SSO locally
+
+    # Azure Active Directory Config
+    AZURE_AD_CLIENT_ID: str = "00000000-0000-0000-0000-000000000000"
+    AZURE_AD_CLIENT_SECRET: str = "incorrect_value"
+    AZURE_AD_TENANT_ID: str = "00000000-0000-0000-0000-000000000000"
+    AZURE_AD_LOGIN_DOMAIN: str = "https://sso.communities.gov.localhost:4005/"
```
samuelhwilliams, line 272: Ah, nice answer to this problem 👍  Can we also document that if a dev wants to use the real microsoft AD they should set this in `.env`? Maybe we add a commented out line in `.env.example`:

```
# AZURE_AD_LOGIN_DOMAIN=https://login.microsoftonline.com/
```

And then devs can just uncomment it as/when they want to use real SSO?

Also maybe we call it AZURE_AD_BASE_URL - it's more than just a domain because we've got the schema in there as well.

README.md
```diff
@@ -41,6 +41,11 @@ Developers are expected to run the app locally using [docker-compose](https://do
 * `make build` / `docker compose build` will rebuild the Funding Service image.
 * `make clean-build` / `docker compose build --no-cache` will rebuild the Funding Service image, bypassing caching. This should rarely be needed.
 
+### SSO
+By default, local development and locally-run end to end tests use a [stub server](./stubs/sso/) to implement the same  interface as Microsoft SSO flow. This is so we can reliably run e2e tests locally without hitting the real SSO endpoints.
```
samuelhwilliams, line 45: It's also so that devs don't need our test azure AD credentials to be able to use the app, making initial setup easier.

README.md
```diff
@@ -41,6 +41,11 @@ Developers are expected to run the app locally using [docker-compose](https://do
...
+
+If you need to allow real SSO login locally, you can set an environment variable of `AZURE_AD_LOGIN_DOMAIN=https://login.microsoftonline.com/` in your .env file, or comment out the LocalConfig version of this variable in the config so it falls back to the BaseConfig default. Be aware that the SSO e2e test will fail while this is enabled - don't forget to switch back to check this passes.
```
samuelhwilliams, line 47: If they need real SSO locally, they also need to get the 3 Azure AD creds from our 1password.

samuelhwilliams: When we update the e2e test to work with real MHCLG SSO we should remember to come back and remove this caveat.

app/config/__init__.py
samuelhwilliams: Ok sorry, you have added some of this info to the readme already ✅ 

I think maybe still nice to add a commented-out thing to .env.example

MarcUsher: Yeah I did consider this, good shout, will add it in

app/common/auth/__init__.py
```diff
@@ -72,6 +73,35 @@ def claim_magic_link(magic_link_code: str) -> ResponseReturnValue:
...
+def sso_sign_in() -> ResponseReturnValue:
+    form = SSOSignInForm()
+    if form.validate_on_submit():
+        session["flow"] = build_auth_code_flow(scopes=current_app.config["MS_GRAPH_PERMISSIONS_SCOPE"])
+        return redirect(session["flow"]["auth_uri"]), 302
+    return render_template("common/auth/sign_in_sso.html", form=form)
+
+
+@auth_blueprint.route("/sso/get-token", methods=["GET"])
+@auto_commit_after_request
```
samuelhwilliams, line 86: Created https://mhclgdigital.atlassian.net/browse/FSPT-409 as a mop-up to minimise future pain from missing this

samuelhwilliams (approved):
Feel free to get this in as is and follow up with the test separately.

nuwan-samarasinghe (approved):
Nice work !!

since after the vacation I have not updated my system and running this in local is a challenge so did a review by going through the code & it is a nice job

PR #141, opened 2025-05-02
Revert "FSPT-363: production server"

samuelhwilliams:
Reverts communitiesuk/funding-service#129


gidsg (approved):
Reverting as this was causing an exception
```
You cannot use AsyncToSync in the same thread as an async event loop - just await the async function directly.
```

PR #143, opened 2025-05-02
FSPT-396 Collections frontend scaffolding

sfount:
https://mhclgdigital.atlassian.net/browse/FSPT-396

This PR creates the basic pages for creating, editing and viewing a CollectionSchema. These pages are not intended to be user-facing, we expect users to complete this part of the journey in a different way but the design of that is still ongoing so these pages have been added to a 'Developer' tab in the left-nav. A [separate ticket](https://mhclgdigital.atlassian.net/browse/FSPT-402) will cover visually differentiating these pages to users.

The model for a CollectionSchema is also created, with a test factory


app/common/data/models/__init__.py
```diff
@@ -56,3 +58,20 @@ class MagicLink(BaseModel):
...
+    name: Mapped[str]
+    version: Mapped[int] = mapped_column(default=1)
+
+    grant_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("grant.id"))
+    grant: Mapped[Grant] = relationship("Grant", back_populates="schemas")
+
+    created_by_id: Mapped[uuid.UUID] = mapped_column(ForeignKey("user.id"))
+    created_by: Mapped[User] = relationship("User")
+
+    __table_args__ = (UniqueConstraint("name", "grant_id", name="uq_schema_name_grant_id"),)
```
samuelhwilliams, line 77: What happens when we create a new version?

app/platform/templates/platform/manage_grant_base.html
```diff
@@ -5,18 +5,39 @@
...
-                        <li class="fs-nav-item"><a href="{{ url_for('platform.view_grant', grant_id=grant.id) }}"
-                               class="{{ 'govuk-link govuk-link--no-visited-state govuk-link--no-underline ' + ('govuk-!-font-weight-bold' if active_sub_navigation_tab == 'grant_dashboard' else '') }}">Dashboard</a>
-                        </li>
-                        <li class="fs-nav-item"><a href="{{ url_for('platform.grant_settings', grant_id=grant.id) }}"
-                               class="{{ 'govuk-link govuk-link--no-visited-state govuk-link--no-underline ' + ('govuk-!-font-weight-bold' if active_sub_navigation_tab == 'grant_settings' else '') }}">Settings</a>
-                        </li>
+                        {% for item in settingsNav %}
+                            <li class="fs-nav-item">
+                                <a
+                                    href="{{ url_for(item.url_for, grant_id=grant.id) }}"
```
samuelhwilliams, line 35: Should we just have the `url_for`ing happen inside the settingsNav dict; this isn't reusable as soon as you have any other params in the URL?

app/common/data/interfaces/exceptions.py
```diff
@@ -9,7 +9,10 @@ class DuplicateValueError(Exception):
     field_name: str
     new_value: str
 
-    constraint_name_map: dict[str, str] = {"uq_grant_name": "name"}
+    # I think this should probably just wrap the IntegrityError and rely on its message for diagnosing
+    # presumably most things calling this will know what the integrity failure is and if they don't they could
+    # inspect the constraint name themselves (in the rare instance there might be multiple)
```
samuelhwilliams, line 14: I think this is a useful comment to put on a PR and then work through, but not useful to have in the code as an unresolved thought? Unless you're saying it's a decision and we should do it rather than just a musing.

sfount: Yeah agree this wouldn't be to stick here permanently but was just pushing this as WIP -- I think either the PR makes the actual change or its just removed entirely and we stick with it (so I put it somewhere I wouldn't forget - happy to avoid that if they start to make it in or we'd prefer not to).

I think I wasn't expecting to update an error signature while adding a database constraint but tbh this mapping has to happen somewhere and I don't have that strong an opinion on it

app/common/data/interfaces/collections.py
```diff
@@ -0,0 +1,32 @@
...
+        db.session.rollback()
+        raise DuplicateValueError(e) from e
+    return schema
+
+
+def get_collection_schema(collection_id: UUID4) -> CollectionSchema:
+    return db.session.get_one(CollectionSchema, collection_id)
+
+
+def update_collection_schema(name: str, collection: CollectionSchema) -> CollectionSchema:
```
samuelhwilliams, line 25: I think a nicer API for these is probably "give me the thing being updated, and then explicitly tell me what updates you're making"?

```suggestion
def update_collection_schema(collection: CollectionSchema, *, name: str) -> CollectionSchema:
```

app/common/data/interfaces/collections.py
```diff
@@ -0,0 +1,32 @@
+from pydantic import UUID4
+from sqlalchemy.exc import IntegrityError
+
+from app.common.data.interfaces.exceptions import DuplicateValueError
+from app.common.data.models import CollectionSchema, Grant, User
+from app.extensions import db
+
+
+def create_collection_schema(name: str, user: User, grant: Grant, version: int = 1) -> CollectionSchema:
```
samuelhwilliams, line 9:
```suggestion
def create_collection_schema(*, name: str, user: User, grant: Grant, version: int = 1) -> CollectionSchema:
```

Can we force kwargs here? I think there's no obvious ordering to the params, so making it clear at the call site might be nicer?

app/common/filters/time_and_date_filters.py
```diff
@@ -0,0 +1,7 @@
+from datetime import datetime
+
+from babel import dates
+
+
+def format_date(value: datetime) -> str:
```
sfount, line 6: It would be good at some point to have the various formats in the design system so that we don't have to keep adding or changing formats as we go - maybe we can make that a small BAU story to add them and give them good descriptive names https://www.gov.uk/guidance/style-guide/a-to-z#dates

app/common/data/migrations/versions/006_add_collection_schema.py
```diff
@@ -0,0 +1,40 @@
...
+from alembic import op
+
+revision = "006_add_collection_schema"
+down_revision = "005_add_utc_suffix"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
samuelhwilliams: agree they bug me as well. Unfortunately they're hardcoded in alembic so not easy to rip out which is annoying. I've considered adding a post-processor to strip them but it feels like almost overkill.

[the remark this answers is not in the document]
srh-sloan: They are gone

app/common/filters/time_and_date_filters.py
srh-sloan: Ticket created here https://mhclgdigital.atlassian.net/browse/FSPT-405

tests/integration/conftest.py
```diff
@@ -163,12 +163,17 @@ def db_session(app: Flask, db: SQLAlchemy) -> Generator[Session, None, None]:
             connection.close()
 
 
-_Factories = namedtuple("_Factories", ["grant", "user", "magic_link"])
+_Factories = namedtuple("_Factories", ["grant", "user", "magic_link", "collection_schema_factory"])
```
samuelhwilliams, line 166:
```suggestion
_Factories = namedtuple("_Factories", ["grant", "user", "magic_link", "collection_schema"])
```

tests/integration/models.py
```diff
@@ -48,3 +48,18 @@ class Meta:
...
+
+class _CollectionSchemaFactory(factory.alchemy.SQLAlchemyModelFactory):
+    class Meta:
+        model = CollectionSchema
+        sqlalchemy_session_factory = lambda: db.session  # noqa: E731
+
+    id = factory.LazyFunction(uuid4)
+    name = factory.Sequence(lambda n: "Collection %d" % n)
+
+    created_by_id = factory.LazyAttribute(lambda o: "o.user.id")
```
samuelhwilliams, line 61: I can see you've copied this from above, but I clearly made a mistake as this should be `o.created_by.id`. I'd expect this to error but we're actually just returning a string, not referencing attributes on the object via `o`. Can you fix this and the magiclink above please? 🙏 

```suggestion
    created_by_id = factory.LazyAttribute(lambda o: o.created_by.id)
```

app/platform/templates/platform/developers/create_collection.html
```diff
@@ -0,0 +1,34 @@
...
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-l">Set up a new collection</h1>
+            <form method="post" novalidate>
+                {{ form.csrf_token }}
+                {{ form.grant_id }}
+                {{ form.name }}
+                {{ form.submit(params={
+                    "text": "Set up collection"
+                }) }}
```
samuelhwilliams, line 30: Any reason we set the label in the template here rather than on the submit field of the form?

app/platform/__init__.py
```diff
@@ -65,3 +74,66 @@ def grant_change_name(grant_id: UUID4) -> str | Response:
...
+
+@platform_blueprint.route("/grants/<uuid:grant_id>/developers/collections/set-up", methods=["GET", "POST"])
+@mhclg_login_required
+@auto_commit_after_request
+def setup_collection(grant_id: UUID4) -> ResponseReturnValue:
+    grant = interfaces.grants.get_grant(grant_id)
+    form = CollectionForm(grant_id=grant.id)
+    if form.validate_on_submit():
+        try:
+            assert form.name.data is not None
```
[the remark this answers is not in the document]
srh-sloan: To fix a mypy error as it doesn't like the next line down - passing in `str | None` instead of `str`

app/platform/templates/platform/developers/create_collection.html
srh-sloan: I think because we use the same form for update or add and the text would be different for each

samuelhwilliams: Ahh ok ok

app/common/data/models/__init__.py
```diff
@@ -14,6 +14,8 @@ class Grant(BaseModel):
 
     name: Mapped[CIStr] = mapped_column(unique=True)
 
+    collection_schemas: Mapped[list["CollectionSchema"]] = relationship("CollectionSchema", lazy=True)
```
samuelhwilliams, line 17: SQLAlchemy has a bunch of different [lazy loading options](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#lazy-loading); `True` here I believe equates to `select`, which will do 1 additional query for each grant you've fetched when you touch the collection_schemas attribute. In most cases this will be fine, because we'll only be looking at one grant at a time, so you'll just have 2 SQL queries.

If we start building (probably admin) pages where we eg list all grants and how many collections are set up for them, this will start spitting out N+1 queries. We could flatten that down again to 2 queries by using `selectin` loading instead.

Maybe something here for us to discuss and decide on. We could have all relationships defined on the models as`lazy='raise'`, then whenever we query them in the interface layer we explicitly say what kind of lazy loading we want. Pros: clear, explicit, intentional, developer is forced to think about how the query will be used. Cons: have to specify the lazy loading type everywhere.

Alternatively, I think even if we don't go that far, I think I'd maybe prefer us to use the string description of the lazy loading method again just for readability. `lazy=True` doesn't help you glance at it and know what kind of approach is being taken; `lazy='select'` might be more explicit (although it still doesn't make the n+1 problem _immediately_ obvious).

samuelhwilliams (approved):
This looks lovely ✅   One comment for us to be thinking about as we get more models/relationships.

app/common/data/models/__init__.py
samuelhwilliams: Also annoyingly I've noticed that flask-debugtoolbar doesn't integrate with flask-sqlalchemy-lite so we can't see any SQL queries in the toolbar. :( Might need to do some PRs on both those libraries to add support, but definitely a big downside of using the -lite version instead of full fat flask-sqlalchemy.

PR #144, opened 2025-05-02
Fspt 366 freeze time

srh-sloan:
New fixture added to allow freezing time in both the code and the database.

- Works for freezing time with an initial date/time
- Works for updating time during the test using the TimeFreezer class

To Do:

- [x] Move TimeFreezer class into it's own file
- [x] Make the fixture test use a temp table for testing, not MagicLinks
- [x] Rebase off main


tests/integration/common/data/test_freeze_time_fixture.py
```diff
@@ -0,0 +1,55 @@
...
+import pytest
+from sqlalchemy import select
+
+from app.common.data.models import MagicLink
+
+
+@pytest.mark.freeze_time("2025-01-01 13:30:00")
+def test_create_magic_link_frozen_time(db_session, factories, time_freezer):
+    factory_ml = factories.magic_link.create()
+    magic_link = db_session.scalar(select(MagicLink).where(MagicLink.id == factory_ml.id))
```
samuelhwilliams, line 12: optional nit/suggestion for everywhere: 

```suggestion
    magic_link = db_session.get(MagicLink, factory_ml.id)
```

tests/integration/common/data/test_freeze_time_fixture.py
```diff
@@ -0,0 +1,55 @@
...
+    assert magic_link.created_at_utc == datetime.strptime("2025-01-01 13:30:00", "%Y-%m-%d %H:%M:%S")
```
samuelhwilliams, line 13: optional nit/suggestion (might just be my personal preference, do feel free to ignore): 

```suggestion
    assert magic_link.created_at_utc == datetime(2025, 1, 1, 13, 30, 0)
```

tests/integration/conftest.py
```diff
@@ -125,6 +128,67 @@ def _integration_test_timeout(request: FixtureRequest) -> None:
     request.node.add_marker(pytest.mark.fail_slow("250ms"))
 
 
+class TimeFreezer:
```
samuelhwilliams, line 131: As discussed on Slack - think it'd be nice to move this out to a separate file (tests/integration/time_freezer.py)

tests/integration/conftest.py
```diff
@@ -125,6 +128,67 @@ def _integration_test_timeout(request: FixtureRequest) -> None:
...
+@pytest.fixture(scope="function")
+def time_freezer(db_session: Session, request: FixtureRequest) -> Generator[TimeFreezer, None, None]:
+    marker = request.node.get_closest_marker("freeze_time")
+    if marker is None:
+        raise ValueError("No frozen time specified. Use pytest.marker.freeze_time to supply time")
+
+    fake_time = marker.args[0]
+    time_freezer = TimeFreezer(fake_time, db_session)
+    yield time_freezer
+    time_freezer.restore_actual_time()
```
samuelhwilliams, line 189: One thought, may have pros and cons:

If we make this an autouse fixture, and then only freeze time if the marker is set, you would only need to add the marker to the test in basic cases (as opposed to marker+signature).

samuelhwilliams (commented):
Really nice to have this functionality now - well done for working through the really gnarly DB overriding stuff. This feels reliable enough I think 👍 (Might need to keep an eye on what happens if tests error out and teardown fails - could cascade weirdly - but one to just watch).

samuelhwilliams (approved):
Really nice, this feels v clean 👏 Getting through the weird postgres behaviour around overriding the now function and how that interacts with eg default behaviour for columns was a lot of work 🙇

MarcUsher:
This is really great @srh-sloan! I briefly looked into this when I spotted the odd `updated_at` behaviour but it was waaaay more complicated than I thought to be able to freeze time for the DB in tests! 

Going to comb through this a bit more to get my head around how it all works but really great work on getting this working 🙌

PR #147, opened 2025-05-06
Change to use python time rather than db time

srh-sloan:
`func.now()` translates to using the database `now()` function which is not what we are trying to test in this case, so updated to use `datetime.now()`


samuelhwilliams:
Also on line 82 🙇

srh-sloan:
> Also on line 82 🙇

Fixed

samuelhwilliams (approved):

PR #148, opened 2025-05-06
FSPT-401: Run migrations conditionally on head file changing

gidsg:
This makes database migrations run conditionally in the main pipeline only if `.current-alembic-head` is in the changeset.

It also adds a standalone workflow for edge cases where we need to force migrations to run in other circumstances.

Note this is modelled as a [composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) to allow steps to be shared without a wholesale refactor of the main workflow.

I have tested this on the branch e.g.

https://github.com/communitiesuk/funding-service/actions/runs/14855949610 - standalone action 
https://github.com/communitiesuk/funding-service/actions/runs/14856870735 - main pipeline

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -82,23 +82,20 @@ jobs:
           echo "ecr_image_location=$ECR_IMAGE_LOCATION" >> $GITHUB_OUTPUT
 
       - name: checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
-      - name: Set up Python
-        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5
+      - name: check if changeset includes migrations
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        id: changes
         with:
-          python-version: 3.13.3
+          filters: |
+            alembic-head:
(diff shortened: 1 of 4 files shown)


srh-sloan (approved):
Nice! Like the splitting out so we can reuse in different circumstances :)

gidsg:
@srh-sloan sorry I made a small refactor, can you re-review?

.github/workflows/deploy.yml
```diff
@@ -82,23 +82,20 @@ jobs:
           echo "ecr_image_location=$ECR_IMAGE_LOCATION" >> $GITHUB_OUTPUT
 
       - name: checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
 
-      - name: Set up Python
-        uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5
+      - name: check if changeset includes migrations
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
```
samuelhwilliams, line 88: any concerns around using a full third-party action in a pipeline that has a significant amount of AWS access?

samuelhwilliams: We could exclude it from renovate bumps by adding it to renovate.json in the root of the repo, yep, if we're happy with that: https://docs.renovatebot.com/configuration-options/#ignoredeps

samuelhwilliams (approved):

PR #149, opened 2025-05-06
Use SQL params rather than interpolating directly

samuelhwilliams:
Although this is test code and there's no risk of SQL injection, I think we should prefer examples of writing SQL that use params over string interpolation (which if done without proper escaping and safety in real code could lead to vulnerabilities).


srh-sloan (approved):
Nice! Safer! 👍 🔒

PR #161, opened 2025-05-09
BAU: Allow @test.communities email domain

MarcUsher:
Now we've implemented SSO, people will be signing in with their `@test.communities.gov.uk` accounts through SSO. As we're only using the test key currently, this PR updates our form validation and route wrapper to allow logins with both `@communities` and `@test.communities` emails.

For consistency this has also now been applied to the magic link sign-in page, which previously validated to only allow `@communities` emails as well. It's likely we'll drop this validation in future once the platform is more built out, as this sign-in route will be the one that's open to and used by most external users, but for now internal users can continue to sign in with both methods.

Adding it as config in the main app and SSO stub seems slightly duplicative but as the SSO stub is only used locally and for locally-run end to end tests this seems like an ok compromise, and allows us to override the `INTERNAL_DOMAINS` variable for prod and not allow `test.communities` accounts.

### DIFF
--- a/stubs/sso/__init__.py
+++ b/stubs/sso/__init__.py
@@ -26,6 +26,7 @@ from stubs.sso.forms import SSOSignInForm
 
 app = Flask(__name__)
 app.config["SECRET_KEY"] = "dummy-value"  # pragma: allowlist secret
+app.config["INTERNAL_DOMAINS"] = ("@communities.gov.uk", "@test.communities.gov.uk")
 
 app.jinja_loader = ChoiceLoader(
     [
--- a/app/common/auth/decorators.py
+++ b/app/common/auth/decorators.py
@@ -1,7 +1,7 @@
 import functools
 from typing import Callable, cast
 
-from flask import abort, redirect, request, session, url_for
+from flask import abort, current_app, redirect, request, session, url_for
 from flask.typing import ResponseReturnValue
 from flask_login import current_user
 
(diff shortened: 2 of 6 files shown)


stubs/sso/__init__.py
```diff
@@ -26,6 +26,7 @@
 
 app = Flask(__name__)
 app.config["SECRET_KEY"] = "dummy-value"  # pragma: allowlist secret
+app.config["INTERNAL_DOMAINS"] = ["@communities.gov.uk", "@test.communities.gov.uk"]
```
srh-sloan, line 29: Can we make this a `tuple` here instead of casting it when we need it? Like 
`app.config["INTERNAL_DOMAINS"] = ("@communities.gov.uk", "@test.communities.gov.uk")`

MarcUsher: Ha yeah I do that originally and then thought there might be reasons (unknown future reasons) we'd want it as a list - will update and let the future us change it if they want to when it becomes necessary.

srh-sloan (approved):

sfount (approved):
Looks good 👌 

Copying over some thoughts from other threads for posterity (they've been followed up on here already) 

> I wonder if we move it from app constants to app config to let production override and not provide the test domain? (unless thats actually desired behaviour!)

> any advantage of an enum that gets turned into a list everywhere its consumed rather than just a lil list?

> all of these thoughts are nits, lmk if you just want a big green check

PR #162, opened 2025-05-09
FSPT-363: Use production server (with patch)

gidsg:
This restores the changes https://github.com/communitiesuk/funding-service/pull/129 but with an additional fix to workaround the exceptions we were seeing running sqlalchemy-lite with gevent under load as described in: https://github.com/pallets-eco/flask-sqlalchemy-lite/issues/11 and https://github.com/benoitc/gunicorn/issues/3070

 note I have not seen any adverse effect in testing from this monkeypatch but we should keep it under review.

Have tested locally, on Pull Preview and on dev, using`lots-of-curls.sh` script which reliably reproduces the issue.


.github/workflows/deploy.yml
```diff
@@ -105,7 +105,7 @@ jobs:
           image: ${{ steps.push_image.outputs.ecr_image_location }}
           access-role-arn: arn:aws:iam::${{ secrets.AWS_ACCOUNT }}:role/${{ inputs.environment }}-funding-service-app-runner-build-role
           region: eu-west-2
-          port: 5000
+          port: 8080
```
MarcUsher, line 108: If we're changing the port here do we need a corresponding PR on the terraform repo changing it from 5000 to 8080? I remember when we added this github action without the port initially it used some default value which didn't match what was being used elsewhere and the healthcheck would fail.

Might not be needed, asking for my own education more than anything!

gidsg: Yes, we do need to do that! Have restored a PR for this part now: https://github.com/communitiesuk/funding-service-terraform/pull/34

sfount (commented):
Change looks good, very small nit might be to link the extra context in the PR description in the commit message which patches the problematic SQL lite behaviour so that its explained when you're in the context of looking at the code

Quick question on if we need this as a committed script for the long term - otherwise happy to approve

docker-compose.yml
```diff
@@ -43,6 +43,7 @@ services:
       DATABASE_NAME: "postgres"
       DATABASE_SECRET: '{"username":"postgres","password":"postgres"}'  # pragma:allowlist secret
       REQUESTS_CA_BUNDLE: /etc/ssl/certs/ca-certificates.crt
+      GEVENT_SUPPORT: True
```
MarcUsher, line 46: Do we need this on the pullpreview docker-compose or is it just for local debugging?

gidsg: I think it only is needed if you're running in debug mode.

gidsg:
Good points @sfount have added an extended commit message for the patch and removed the curl script now.

MarcUsher (approved):
Approving now @sfount's comments and mine have been addressed

PR #171, opened 2025-05-14
BAU: Fix entrypoint on local docker runner

gidsg:
Flask development server [defaults to `wsgi.py`](https://github.com/pallets/flask/blob/a5f9742398c9429ef84ac8a57b0f3eb418394d9e/src/flask/cli.py#L351) if it exists. Following merge of #162 this meant code in this file was being erroneously executed when running the local development server.

 
This explicitly sets a different entrypoint to avoid erroneously running gevent code which is only used in deployed environments.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,13 +37,13 @@ services:
       "
     env_file: .env
     environment:
+      FLASK_APP: app/__init__.py
       FLASK_ENV: local
       DATABASE_HOST: "db"
       DATABASE_PORT: 5432
       DATABASE_NAME: "postgres"
       DATABASE_SECRET: '{"username":"postgres","password":"postgres"}'  # pragma:allowlist secret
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -37,13 +37,13 @@ services:
       "
     env_file: .env
     environment:
+      FLASK_APP: app/__init__.py
```
samuelhwilliams, line 40: Can just be:

```suggestion
      FLASK_APP: app
```

But no big problem either way

samuelhwilliams (approved):

docker-compose.yml
samuelhwilliams: The dockerfile also feeds into pullpreview, so we'll need to differentiate between these two regardless. I think it's fine to have in this docker compose file

srh-sloan (approved):
🚀

PR #164, opened 2025-05-12
FSPT-406: Section and Form Frontend Scaffolding

srh-sloan:
https://mhclgdigital.atlassian.net/browse/FSPT-406

This PR creates basic screens to allow adding and editing sections and forms within a data collection.

Migration `007_section_form` creates 2 new entitites:
- Section
- Form
These have skeleton fields as per the ticket, to just allow the screens to function. The `form` entity has a placeholder property for `questions` to just show where questions would appear on the page.

Adds new screens, similar to the sketches in the mural, to allow you to add, edit and reorder sections and forms. These screens have not been through a formal design process yet so are considered 'scaffolding' screens - just enough to illustrate the user journey.

- Screens don't yet show any marker that they are not 'real' screens, this is under a separate ticket: https://mhclgdigital.atlassian.net/browse/FSPT-402
- Back links are hard coded for (almost) every screen. Sorting this out will be covered in https://mhclgdigital.atlassian.net/browse/FSPT-417, again once we have more design and know how we want these links to work
- The `platform/__init__py` file feels like it's getting big, maybe think about splitting out
- Not sure on naming conventions for routes and pages yet, will wait til design done and potentially refactor. eg. Manage section, list sections, manage form, list forms - need to decide on what is 'manage', what is 'list' etc.

I will do a squash merge for this as the commit history isn't very useful and a bit messy!

PR #174, opened 2025-05-14
FSPT-407: Frontend scaffolding for questions

srh-sloan:
https://mhclgdigital.atlassian.net/browse/FSPT-407

Adds frontend screens (yet to go through full design, but based on [this mural](https://app.mural.co/t/mhclg2837/m/mhclg2837/1744304689877/1f2bc64bde530c8a210f3b91b8e9da458d0b594e) to do:
- Add a question
- Reorder questions in a form
- Edit a question

Adds skeleton database model, with test factory, for:
- Question

Depends on https://github.com/communitiesuk/funding-service/pull/164


sfount:
I've not looked through the code in detail yet but it would be good to talk about the add/ edit pattern for a more complicated page like adding a question. We cut a lot of corners in the prototype but I think it might be I'd be a bit worried about those two pages diverging on functionality if they're completely separate

app/common/data/interfaces/exceptions.py
```diff
@@ -17,6 +17,8 @@ class DuplicateValueError(Exception):
         "uq_form_slug_section": "title",
         "uq_section_slug_collection_schema": "title",
         "uq_collection_slug_grant_id": "name",
+        "uq_question_slug_form": "text",
+        "uq_question_text_form": "text",
```
samuelhwilliams, line 21: Also missing `uq_question_name_form`

app/common/data/interfaces/collections.py
```diff
@@ -135,3 +139,43 @@ def update_form(form: Form, *, title: str) -> Form:
...
+
+def move_question_down(question: Question) -> Question:
+    swap_elements_in_list_and_flush(question.form.questions, question.order, question.order + 1)
+    return question
+
+
+def update_question(question: Question, *, text: str, hint: str | None, name: str) -> Question:
+    question.text = text
+    question.hint = hint
+    question.name = name
```
samuelhwilliams, line 174: We're missing an update on the `slug` here if ... text or name(?) has changed.

app/platform/__init__.py
```diff
@@ -373,3 +378,155 @@ def edit_form(grant_id: UUID4, collection_id: UUID4, section_id: UUID4, form_id:
...
+            field_with_error.errors.append(f"{field_with_error.label.text} already in use")  # type:ignore[attr-defined]
+
+    return render_template(
+        "platform/developers/edit_question.html",
+        grant=question.form.section.collection_schema.grant,
+        collection=question.form.section.collection_schema,
+        section=question.form.section,
+        form=question.form,
+        question=question,
+        wt_form=wt_form,
```
samuelhwilliams, line 530:
```suggestion
        db_form=question.form,
        question=question,
        form=wt_form,
```

I added a meta test I hoped would catch this but I clearly did it wrong - can you fix the validation please? 😓 https://github.com/communitiesuk/funding-service/blob/1cc47d972ac39e625a59dcdb86640e91bac00286/tests/integration/conftest.py#L98

app/platform/templates/platform/developers/manage_form.html
```diff
@@ -2,41 +2,39 @@
...
-  <div class="govuk-grid-row">
+</h1>
+{% set question_text %}
+{% trans count=form.questions | length %}
+    {{ count }}
+    question {% pluralize %}
+    {{ count }}
+    questions
+{% endtrans %}
+{% endset %}
```
samuelhwilliams, line 34: This doesn't seem to get shown on the page anywhere.

app/platform/__init__.py
```diff
@@ -373,3 +378,154 @@ def edit_form(grant_id: UUID4, collection_id: UUID4, section_id: UUID4, form_id:
...
+@auto_commit_after_request
+def move_question(
+    grant_id: UUID4, collection_id: UUID4, section_id: UUID4, form_id: UUID4, question_id: UUID4, direction: str
+) -> ResponseReturnValue:
+    question = get_question_by_id(question_id=question_id)
+
+    if direction == "up":
+        move_question_up(question)
+    elif direction == "down":
+        move_question_down(question)
```
samuelhwilliams, line 478:
```suggestion
    if direction == "up":
        move_question_up(question)
    elif direction == "down":
        move_question_down(question)
    else:
        abort(400)
```

Might help us catch if we ever mess up the thing we pass into `direction`

tests/integration/platform/test_routes.py
```diff
@@ -278,3 +278,200 @@ def test_move_section(authenticated_client, factories, db_session):
...
+        url_for(
+            "platform.move_form",
+            grant_id=section.collection_schema.grant.id,
+            collection_id=section.collection_schema.id,
+            section_id=section.id,
+            form_id=form1.id,
+            direction="down",
+        ),
+    )
+    assert result.status_code == 302
```
samuelhwilliams, line 297: If we're looking for redirects, can we assert on the redirect URL please? Just from reading the test I can't be confident where it's going - I could imagine (from experience sadly) that it might just redirecting us to some login page instead, which would undermine the test.

app/platform/__init__.py
srh-sloan: Ah, nothing wrong with the test, but you have to do something that actually triggers the template with an error, ie have a test for all the templates 🫣

app/common/data/migrations/versions/009_add_question.py
```diff
@@ -0,0 +1,48 @@
+"""Add question table
+
+Revision ID: 008_add_question
+Revises: 007_add_section_form
```
samuelhwilliams, line 4:
```suggestion
Revision ID: 009_add_question
Revises: 008_add_userrole_org_tables
```

app/common/data/models/__init__.py
```diff
@@ -190,3 +191,50 @@ class Form(BaseModel):
...
+        "Question", lazy=True, order_by="Question.order", collection_class=ordering_list("order")
+    )
+
+
+class QuestionDataType(enum.StrEnum):
+    # If adding values here, also update QuestionTypeForm
+    # and manually create a migration to update question_type_enum in the db
+    TEXT_SINGLE_LINE = "A single line of text"
+    INTEGER = "A whole number"
+    TEXT_MULTI_LINE = "Multiple lines of text"
```
samuelhwilliams, line 205:
```suggestion
    TEXT_SINGLE_LINE = "A single line of text"
    TEXT_MULTI_LINE = "Multiple lines of text"
    INTEGER = "A whole number"
```

Just so they order a bit more nicely on the 'what kind of question' page 👀

app/platform/__init__.py
```diff
@@ -373,3 +382,156 @@ def edit_form(grant_id: UUID4, collection_id: UUID4, section_id: UUID4, form_id:
...
+@platform_blueprint.route(
+    "/grants/<uuid:grant_id>/developers/collections/<uuid:collection_id>/sections/<uuid:section_id>/forms/<uuid:form_id>/questions/add",
+    methods=["GET", "POST"],
+)
+@mhclg_login_required
+@auto_commit_after_request
+def add_question(grant_id: UUID4, collection_id: UUID4, section_id: UUID4, form_id: UUID4) -> ResponseReturnValue:
+    form = get_form_by_id(form_id)
+    question_data_type_arg = request.args.get("question_data_type", None)
+    question_data_type_enum = QuestionDataType.coerce(question_data_type_arg)
```
samuelhwilliams, line 428: If the `question_data_type` is invalid I think this'll throw a 500, which ideally we wouldn't make super easy based on user input.

Probably a fallback value of text is easier, but the alternative is to render a 400 of some kind.

```suggestion
    question_data_type_arg = request.args.get("question_data_type", QuestionDataType.TEXT.name)
    question_data_type_enum = QuestionDataType.coerce(question_data_type_arg)
```

app/platform/forms.py
```diff
@@ -46,3 +49,40 @@ class FormForm(FlaskForm):
...
+        widget=GovTextArea(),
+        description="The question hint will be shown to users when they are answering the question.",
+    )
+    name = StringField(
+        "Question name",
+        validators=[DataRequired("Enter the question name")],
+        description="The question name will be shown when exporting submissions from your users.",
+        filters=[strip_string_if_not_empty],
+        widget=GovTextInput(),
+    )
```
samuelhwilliams, line 87: image

A similar issue here with using the question label for the error summary - I don't think this will scale.

The field label defined here should be the question, so I don't think that should change. But we need some way to build the correct error message; I think our formulaic approach is not working well.

srh-sloan: Fixed this elsewhere but just missed this one - done now

app/platform/templates/platform/developers/choose_question_type.html
```diff
@@ -0,0 +1,39 @@
...
+            <form method="post" novalidate>
+
+                {{ form.csrf_token }}
+                {{ form.form_id }}
+
+                {{ form.question_data_type }}
+
+                {{ form.submit(params={
+                "text": "Continue"
+            }) }}
```
samuelhwilliams, line 35: `What is the type of the question?` should be the label on the form field, and then for simple one-question pages like this, use the isPageHeading params when rendering the question. This will mean the question is properly linked up to the input box and there's no duplication/redundancy of text on the page.

```suggestion
            <form method="post" novalidate>

                {{ form.csrf_token }}
                {{ form.form_id }}

                {{ form.question_data_type (params={"label": {"isPageHeading": true, "classes": "govuk-label--l" } } ) }}

                {{ form.submit(params={
                "text": "Continue"
            }) }}
```

samuelhwilliams (approved):
One small thing to fix above but then ✅

samuelhwilliams (approved):

PR #175, opened 2025-05-14
FSPT-410: Implement user roles in the database

MarcUsher:
Creates UserRole table for storing a User's roles at all levels. By having nullable columns for Org and Grant Foreign Key references, we can infer the permission level for that user, eg. 
- if they have the `admin` role but org and grant are null, then they are a global/platform admin
- if they have the `admin` role and an org id, then they are an organisation-level admin who would be able to eg. add other users to the organisation
This also means we just have one table to check for roles and permissions rather than splitting it across multiple join tables. 

Creates a stub `role_enum` with some initial broad user roles, an some basic check constraints for now for obvious cases, and a stub Organisation table, which will certainly be extended in the future but means we can have the UserRole table in place.

Adds tests to ensure these constraints are working as expected, which we might want to do later for existing constraints already set up in our tables that we're not explicitly testing elsewhere.

### DIFF
--- a/app/common/data/models/__init__.py
+++ b/app/common/data/models/__init__.py
@@ -5,7 +5,7 @@ from enum import Enum
 
 from pytz import utc
 from sqlalchemy import CheckConstraint, ForeignKey, Index, UniqueConstraint
-from sqlalchemy import Enum as PgEnum
+from sqlalchemy import Enum as SqlEnum
 from sqlalchemy.ext.orderinglist import ordering_list
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
@@ -76,7 +76,7 @@ class UserRole(BaseModel):
     )
     grant_id: Mapped[uuid.UUID | None] = mapped_column(ForeignKey("grant.id", ondelete="CASCADE"), nullable=True)
     role: Mapped[RoleEnum] = mapped_column(
-        PgEnum(
+        SqlEnum(
             RoleEnum,
             name="role_enum",
             validate_strings=True,
--- a/app/common/data/models/__init__.py
+++ b/app/common/data/models/__init__.py
@@ -1,13 +1,25 @@
 import datetime
 import secrets
 import uuid
+from enum import Enum
 
 from pytz import utc
-from sqlalchemy import ForeignKey, Index, UniqueConstraint
+from sqlalchemy import CheckConstraint, ForeignKey, Index, UniqueConstraint
+from sqlalchemy import Enum as PgEnum
 from sqlalchemy.ext.orderinglist import ordering_list
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
 from app.common.data.base import BaseModel, CIStr
+
+
+class RoleEnum(str, Enum):
+    ADMIN = (
+        "admin"  # Admin level permissions, combines with null columns in UserRole table to denote level of admin access
+    )
+    MEMBER = "member"  # Basic read level permissions
+    EDITOR = "editor"  # Read/write level permissions
+    ASSESSOR = "assessor"  # Assessor level permissions
+    S151_OFFICER = "s151_officer"  # S151 officer sign-off permissions
 
 
 class Grant(BaseModel):
@@ -16,6 +28,16 @@ class Grant(BaseModel):
     name: Mapped[CIStr] = mapped_column(unique=True)
 
     collection_schemas: Mapped[list["CollectionSchema"]] = relationship("CollectionSchema", lazy=True)
+    roles: Mapped[list["UserRole"]] = relationship("UserRole", back_populates="grant", cascade="all, delete-orphan")
+
+
+class Organisation(BaseModel):
+    __tablename__ = "organisation"
+
+    name: Mapped[CIStr] = mapped_column(unique=True)
+    roles: Mapped[list["UserRole"]] = relationship(
+        "UserRole", back_populates="organisation", cascade="all, delete-orphan"
+    )
 
 
 class User(BaseModel):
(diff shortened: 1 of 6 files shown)


app/common/data/models/__init__.py
```diff
@@ -42,6 +64,29 @@ def get_id(self) -> str:
...
+        ),
+        nullable=False,
+    )
+
+    __table_args__ = (
+        UniqueConstraint("user_id", "org_id", "grant_id", "role", name="uq_user_org_grant_role"),
+        Index("ix_user_roles_user_id", "user_id"),
+        Index("ix_user_roles_org_id", "org_id"),
+        Index("ix_user_roles_grant_id", "grant_id"),
+    )
```
samuelhwilliams, line 87: Can we get some check constraints on here as well please?

See if you can think of any more, but off the top of my head at least:
- `role = 'member' AND NOT (org_id IS NULL AND grant_id IS NULL)` - we don't have a "member" role for "the entire platform"
- `role = 's151_officer' AND (org_id IS NULL OR grant_id IS NOT NULL)` - S151 officers can only be assigned for an entire org, not on a platform or per-grant basis.

MarcUsher: Yes good shout, thank you! Going to review and finish this off this morning as I threw it together quickly before finishing on thurs.

samuelhwilliams: I also might suggest proactively adding a few more composite index on `user_id + org_id + grant_id`. I expect this table will be rarely written and frequently read, so optimising for fast lookups feels very reasonable.

tests/integration/models.py
```diff
@@ -36,6 +47,40 @@ class Meta:
...
+
+    class Params:
+        has_organisation = factory.Trait(
+            organisation_id=factory.LazyAttribute(lambda o: o.organisation.id),
+            organisation=factory.SubFactory(_OrganisationFactory),
+        )
+        has_grant = factory.Trait(
+            grant_id=factory.LazyAttribute(lambda o: o.grant.id),
+            grant=factory.SubFactory(_GrantFactory),
+        )
```
samuelhwilliams, line 81: :chefkiss: very fancy

tests/integration/models.py
```diff
@@ -36,6 +47,40 @@ class Meta:
...
+        sqlalchemy_session_factory = lambda: db.session  # noqa: E731
+
+    id = factory.LazyFunction(uuid4)
+    user_id = factory.LazyAttribute(lambda o: o.user.id)
+    user = factory.SubFactory(_UserFactory)
+    organisation_id = None
+    organisation = None
+    grant_id = None
+    grant = None
+    role = RoleEnum.ADMIN  # Default to ADMIN role but can be overridden when initialising the factory
```
samuelhwilliams, line 71: Do we even just leave this as `None` maybe? 🤔  Make it explicitly defined everywhere we use/create a role?

tests/integration/common/data/db/test_constraints.py
```diff
@@ -0,0 +1,42 @@
...
+        )
+
+    @pytest.mark.parametrize("has_organisation, has_grant", [(True, True), (True, False)])
+    def test_assessor_role_grant_only(self, db_session, factories, has_organisation, has_grant):
+        user_role = factories.user_role.create(
+            has_organisation=has_organisation, has_grant=has_grant, role=RoleEnum.ASSESSOR
+        )
+        db_session.add(user_role)
+        with pytest.raises(IntegrityError) as error:
+            db_session.flush()
```
samuelhwilliams, line 38: hmm something about this feels off - I thought that `.create()` on a factory should already add and flush/commit it to the session 🤔

tests/integration/common/data/db/test_constraints.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+from sqlalchemy.exc import IntegrityError
+
+from app.common.data.models import RoleEnum
+
+
+class TestUserRoleConstraints:
```
samuelhwilliams, line 7: Nice tests, good to have these checks on constraints when we're relying on them for such important things 👍

tests/integration/models.py
MarcUsher: Haha thanks. The other way to achieve the same result was to use their [Maybe](https://factoryboy.readthedocs.io/en/stable/reference.html#maybe) but I felt like that just made it a bit more cluttered to read, at least this is very directly `None` unless you change the boolean.

tests/integration/models.py
MarcUsher: Yeah that seems sensible, will update!

app/common/data/models/__init__.py
```diff
@@ -1,21 +1,43 @@
...
+    S151_OFFICER = "s151_officer"  # S151 officer sign-off permissions
+
+
 class Grant(BaseModel):
     __tablename__ = "grant"
 
     name: Mapped[CIStr] = mapped_column(unique=True)
 
     collection_schemas: Mapped[list["CollectionSchema"]] = relationship("CollectionSchema", lazy=True)
+    roles: Mapped[list["UserRole"]] = relationship("UserRole", back_populates="grant", cascade="all, delete-orphan")
```
samuelhwilliams, line 31: comment: let's keep an eye on if this is useful. I can see an indirect relationship to `users` being maybe useful, but I don't imagine we'll ever find it that useful to be like "what permissions do <people> have on this grant" 👀

samuelhwilliams: Same for organisation

MarcUsher: Yeah agree, added more for consistency than anything but not sure how often this will get used

MarcUsher:
> DB spec looks good, nice tests, good work 👍
> 
> One thought for the future that I should have flagged earlier - with the scope of this ticket being just the DB changes, it's hard to get much confidence that this is actually a useful and functioning set of changes. Maybe in the future we always do DB changes in unison with some actual product change (eg in this case, reading FSD_ADMIN and creating the Admin role in the DB when an MHCLG person logs in)?

Yeah that's fair - I think my rationale in chunking it up this way was to get something in that wouldn't be blocked by potentially slower progress on how we want to do seeding etc, but I can definitely see your point that this on its own doesn't really do much (and technically the scope of the ticket is to do the DB changes and the functional changes, I've just split it in the delivery to have smaller PRs)

PR #178, opened 2025-05-15
FSPT-402 - Watermark

srh-sloan:
Adds CSS to make a `div` behave as a watermark. This allows us to visually differentiate screens if we need to. Some example use cases for this are:
- Developers - for screens that won't be in the real use journey and are just developer facing while we work on the platform
- Scaffolding - for screens that allow us to see a user journey playing out, before the data model and/or journey design is complete

This PR also implements this 'developers' watermark on the relevant existing pages.

### DIFF
--- a/app/assets/main.scss
+++ b/app/assets/main.scss
@@ -6,39 +6,39 @@ $govuk-assets-path: "/static/assets/";
 @import "../../node_modules/govuk-frontend/dist/govuk";
 
 .app-\!-no-wrap {
-  white-space: nowrap;
+    white-space: nowrap;
 }
 
 // this class will only be applied if GOV.UK Frontend Jinja has set the `.js-enabled` class meaning
 // it could execute JavaScript
 .js-enabled .app-js-hidden {
-  display: none;
+    display: none;
 }
 
 .fs-nav-item > a {
-  display: block;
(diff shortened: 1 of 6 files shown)


app/assets/main.scss
```diff
@@ -6,17 +6,39 @@ $govuk-assets-path: "/static/assets/";
 @import "../../node_modules/govuk-frontend/dist/govuk";
 
 .app-\!-no-wrap {
-    white-space: nowrap;
```
sfount, line 9: 🤏 very minor nit but this changes the tab spacing to 2 where in most files we use 4, we should be consistent but not doing that would also help focus the review! We should get these formatted alongside HTML I'll go pick that PR back up soon

app/platform/templates/platform/developers/collection_details.html
```diff
@@ -17,6 +17,7 @@
 {% endblock beforeContent %}
 
 {% block content %}
+    <div class="fs-watermark">Developers</div>
```
sfount, line 20: adding `aria-hidden="true"` would ask screen readers to ignore this, we'll be removing it before any users see it so I don't think this should be too big of an issue

nuwan-samarasinghe (approved):
Nice change

PR #180, opened 2025-05-16
Add date/datetime formatters for consistent formatting

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-405

These will format dates/datetimes in accordance with the GOV.UK Style
Guide; having these defined in one place will make it easier to make
sure we format dates correctly and consistently across the board. Adding
more filters/variations should probably be scrutinised.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 from flask import Flask, Response, redirect, url_for
 from flask.typing import ResponseReturnValue
 from flask_babel import Babel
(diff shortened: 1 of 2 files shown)


app/common/filters.py
```diff
@@ -0,0 +1,39 @@
...
+        fmt = "%-I%p on %A %-d %B %-Y"
+
+    formatted_datetime = value.strftime(fmt)
+    formatted_datetime = formatted_datetime.replace("AM", "am").replace("PM", "pm")
+
+    return formatted_datetime
+
+
+def format_date_range(start: date, end: date) -> str:
+    from_, to = format_date(start), format_date(end)
```
srh-sloan, line 33: Could we have a docstring here like above please? I know you can scroll up if you're in this file, but I don't think the IDE will show you it based on the other functions?

app/common/filters.py
```diff
@@ -0,0 +1,39 @@
...
+    return f"{from_} to {to}"
+
+
+def format_datetime_range(start: datetime, end: datetime) -> str:
+    from_, to = format_datetime(start), format_datetime(end)
```
srh-sloan, line 38: And a docstring here too please :)

samuelhwilliams: Yep yep, fair shout 👍  I don't know why I didn't - I think just because it was mostly handing off to the other functions in a fairly trivial way, but still good to have an example of the format I think

srh-sloan (approved):

PR #185, opened 2025-05-21
Make test factories flush on create

MarcUsher:
When trying to use our Factory Boy factory `create` calls to trigger an Integrity Error we realised that the create calls weren't actually flushing or committing to the test DB, but just creating and adding the entity ephemerally in the session. This is the default behaviour [see docs](https://factoryboy.readthedocs.io/en/stable/orms.html#factory.alchemy.SQLAlchemyOptions.sqlalchemy_session_persistence)

When we call the `create` method with our factories this should perform a session flush, to differentiate it from the `build` method (when explicit adding and flushing/committing) is needed.

This sets the `sqlalchemy_session_persistence` to `flush` for all our factories, which should be our default going forward, so when we call `create` we know the object is created and a session flush is performed.

### DIFF
--- a/tests/integration/models.py
+++ b/tests/integration/models.py
@@ -22,6 +22,7 @@ class _GrantFactory(factory.alchemy.SQLAlchemyModelFactory):
     class Meta:
         model = Grant
         sqlalchemy_session_factory = lambda: db.session  # noqa: E731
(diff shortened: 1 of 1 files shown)


srh-sloan (approved):
Nice fix!

PR #187, opened 2025-05-21
FSPT-410: Add Platform Admin role for FSD_ADMIN members on SSO login

MarcUsher:
Draft PR - commits will be amended and new will be added!

Current work:
- make nulls not distinct for the UserRole table, for a reliable unique constraint on rows in that table (adds a test to check for this)
- sets the ADMIN role on SSO login for users who are in the `FSD_ADMIN` AD group
- creates a new route decorator to restrict routes to platform admins only

Still do to do:
- add custom exception for the `add_user_role` method for combinations which violate the unique and check constraints of the table
- add tests for the `add_user_role` interface
- add tests for the custom exception
- add 403 template
- actually add the decorator to relevant routes and update unit/e2e tests accordingly


app/common/data/interfaces/exceptions.py
```diff
@@ -31,3 +32,34 @@ def __init__(self, integrity_error: IntegrityError) -> None:
...
+        diagnostics = cast(CheckViolation, integrity_error.orig).diag
+        self.model_name = getattr(diagnostics, "table_name", None)
+        self.constraint_name = getattr(diagnostics, "constraint_name", None)
+
+        if self.constraint_name and self.constraint_name in self.constraint_message_map:
+            self.message = self.constraint_message_map[self.constraint_name]
+        else:
+            self.message = str(integrity_error)
+
+        current_app.logger.error(
```
samuelhwilliams, line 58: If we log this as an error, it will create an issue in Sentry. Do we definitely want this?

If we handle the exception and do something normal for the user (eg return a 400 or show an error message), do we still want to ping all developers with that?

app/common/templates/common/errors/403.html
```diff
@@ -0,0 +1,18 @@
...
+{% endblock pageTitle %}
+
+{% block content %}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-l">Sorry, you don't currently have permission to access page</h1>
+            <p class="govuk-body">
+                If you believe this is incorrect, please email us on
+                <a class="govuk-link govuk-link--no-visited-state"
+                    href="mailto:test@communities.gov.uk">test@communities.gov.uk</a>.
```
samuelhwilliams, line 14: `test@communities.gov.uk` 😅  this feels like we might want to tweak this? Probably add a config for FUNDING_SERVICE_EMAIL_ADDRESS or something?

samuelhwilliams:
Tests look nice 👏

app/common/templates/common/errors/403.html
MarcUsher: Haha yes this was a placeholder that I need to go back and update 😆 but also knowing the content on this page will prob change as we agreed with Kate and Brian that we can use dummy content/similar content to existing services for now and tweak later

app/common/data/interfaces/exceptions.py
MarcUsher: Yeah this is a fair point, we're not logging much in the new app yet so wasn't sure what level we wanted to log things at. Maybe this can switch this to be a logger.warning or .info and return a 400 for now (with a generic 400 page)? 

The logging was also me thinking we could log useful things like the values that had tried to be added in the row in case we needed to debug but your clever logging setup blocks you from logging any UUIDs so without that the logging might even be a bit redundant.

app/common/auth/__init__.py
```diff
@@ -94,6 +96,12 @@ def sso_get_token() -> ResponseReturnValue:
 
     user = get_or_create_user(email_address=sso_user["preferred_username"])
 
+    if "FSD_ADMIN" in sso_user["roles"]:
+        try:
+            add_user_role(user_id=user.id, role=RoleEnum.ADMIN)
+        except InvalidUserRoleError as e:
+            abort(400, description=str(e))
```
samuelhwilliams, line 103: I'm not sure we want to catch this error, do we?

The user isn't in control of the role that gets added, so a 400 (client error) feels inappropriate - there's nothing they can do to fix it?

We should probably just not catch the error and let it bubble up to a 500.

MarcUsher: Yeah the error won't ever really get surfaced here as we're always passing a correct combo, but the thinking behind adding it here was as a bit of an eg. for future uses of the `add_user_role` interface - happy to remove it and just raise the error though.

app/common/templates/common/errors/403.html
```diff
@@ -0,0 +1,20 @@
...
+{% endblock pageTitle %}
+
+{% block content %}
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">Sorry, you don't currently have permission to access this page</h1>
+        <p class="govuk-body">
+            If you believe this is incorrect, please contact us through our
+            <a class="govuk-link govuk-link--no-visited-state"
+                href="https://mhclgdigital.atlassian.net/servicedesk/customer/portal/5/group/68">support desk</a> if you
```
samuelhwilliams, line 14: Can we move this to something in our config object, just so it's not hardcoded in a template? Feels a bit magic-string'y to be here randomly.

samuelhwilliams (approved):

PR #191, opened 2025-05-22
FSPT-477 pullpreview sso stub implementation

nuwan-samarasinghe:
**Ticket : https://mhclgdigital.atlassian.net/browse/FSPT-477**

Following are the implementation steps that I have followed 

1. First I have converted the sso stub app into a another gunicorn service because this sso service if going to pretend like Microsoft login so to do that we have to have this SSO service as a separate service that is the reason behind this change

2. Adding some Network configuration in to the pull preview and in the caddy service I have implemented a subdomain and from this subdomain it will reverse proxy the requests into the correct containers if it has sso.<> then goes to sso stub server here I have added the `AZURE_AD_BASE_URL: https://sso.${PULLPREVIEW_PUBLIC_DNS}/` as a env since for given pullpreview has a different url

Also 

`extra_hosts`: This allows to define additional hostname-to-IP mappings for the container.
`host-gateway`: This tells Docker to resolve the hostname to the IP address of the host machine's gateway. This is particularly useful when you want a container to access services running on the host.

So this will mimic the Microsoft login behaviour 

3. SSO should have some static css, js files with this change it enables to get the static content for the sso service

### DIFF
--- a/.pullpreview/Caddyfile.template
+++ b/.pullpreview/Caddyfile.template
@@ -5,3 +5,7 @@ ${PULLPREVIEW_PUBLIC_DNS}:443 {
 
     reverse_proxy web:8080
 }
+
+sso.${PULLPREVIEW_PUBLIC_DNS}:443 {
+    reverse_proxy sso:8080
+}
--- a/app/config.py
+++ b/app/config.py
@@ -336,7 +336,7 @@ class PullPreviewConfig(_SharedConfig):
     AZURE_AD_CLIENT_ID: str = "00000000-0000-0000-0000-000000000000"
     AZURE_AD_CLIENT_SECRET: str = "incorrect_value"
     AZURE_AD_TENANT_ID: str = "00000000-0000-0000-0000-000000000000"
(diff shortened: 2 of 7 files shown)


samuelhwilliams (commented):
Really nice work on this Nuwan - looks great in pullpreview 👏 

Only finding one problem at the moment - using the `FlaskAssetsViteExtension` extension here has a hard requirement on `FLASK_ENV` being set, which we don't configure on the SSO stub server on local machines. So I think we need to add that.

Semi-interestingly, vite's hot reloading doesn't work for the SSO stub server locally. I think this is fine and we don't need to fix it, but I don't immediately understand why.

After adding that locally, it all looks good to me 👍

samuelhwilliams (approved):
Really nice work on this @nuwan-samarasinghe - threading the needle to get this all to work in the pullpreview environment, with the caddy proxy, and the vite server is super impressive. And it didn't take you long at all - v impressive as always 🙇 The `host-gateway` docker-compose extra host is clean, and switching the SSO stub server over to a factory is nicely consistent with how we do the main app.

PR #194, opened 2025-05-22
Use composite PK on CollectionSchema

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-437

The CollectionSchema table will (at some future time) support versioning. When a new version of the schema is created, we want a stable identifier that will let us track versions over time. This will be the `id` column. The `version` column will get incremented each time.

At the moment the `id` column is the sole PK, which would require later versions to have a new ID, and make it hard to track versions over time.

This patch switches us to using a composite PK (id, version) and cascades that appropriately to the related `sections` table's FK.

Rather than writing a migration that applies this cleanly, I think it's OK for now to manually wipe out our databases and then re-build them from scratch.

## Note for devs
You'll need to `make down` and wipe your docker volume that contains the funding-service data; should be `docker volume rm funding-service_postgres_data`


app/common/data/interfaces/collections.py
```diff
@@ -23,8 +23,16 @@ def create_collection_schema(*, name: str, user: User, grant: Grant, version: in
     return schema
 
 
-def get_collection_schema(collection_id: UUID4) -> CollectionSchema:
-    return db.session.get_one(CollectionSchema, collection_id)
+def get_collection_schema(collection_id: UUID4, version: int | None = None) -> CollectionSchema:
```
srh-sloan, line 26: Can we add a docstring that says this defaults to the latest version if `version` is `None` please?

srh-sloan (commented):
Tiny comment but otherwise makes sense and nicely kept to a smaller PR by not addressing all the issues versioning brings in at the same time 👍

srh-sloan (approved):

srh-sloan (approved):

PR #197, opened 2025-05-22
Move developer routes to their own package

samuelhwilliams:
We're now scaffolding pages that straddle the 'setting up collections', 'providing data', and soon 'reviewing data' areas. These sit across both the deliver grant funding and access grant funding domains/services.

To keep all of the work-in-progress/not-for-public-consumption pages in one place, we'll move these routes to a developers package and namespace them all with a `/developers` URL prefix.

This should make it clear that all of these pages have not been through a final design process, and consideration hasn't yet been given to eg public URLs, design+product decisions might not have been totally firmed up, etc.

We should be able to have a lighter touch review process on anything going into the developers package, and a more strict+stringent review process for anything moving out of it.

This doesn't mean that we're throwing things together totally willy-nilly in the developers package, but it does slightly reduce the overhead and mean that we can prioritise moving quickly and proving ourselves.


MarcUsher (approved):
No sneaky changes eh? 👀

PR #203, opened 2025-05-23
Create collection instances and view a tasklist

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-437

## Summary
These patches add a button on the 'manage schema' page that allows you to create a collection instance (which is the data a user is submitting against a schema) so that we can start to preview/test the schema as it's built.

The tasklist is currently non-functional, but will list the sections and forms in the correct order.

---

I've added a `CollectionHelper` class. The name here is intentionally a bit bad, because naming it feels hard.

This helper class is intended to wrap all of the business logic for
future features such as conditional questions/routing, injecting
contextual/dynamic information, etc. This should make it easy and safe
to reuse the logic across the platform without having to reimplement it
in multiple places.

Most pages that interact with collections/sections/forms/questions
should go through this (and perhaps some additional helper classes we
add later) instead of using the DB models directly.

## Acceptance criteria from ticket
 - [ ] There is a button labelled “Preview collection” on the “Manage collection” page which
 - [ ] Deletes all existing collections by that user for the collection schema
   - [ ] Creates a new empty one
   - [ ] Redirects the user to the collection tasklist page
 - [ ] The collection tasklist page shows all of the sections and forms in the schema, in the correct order.
 - [ ] The status for each section/form is just ‘not started'
 - [ ] There is a collections helper class that
   - [ ] Accepts a collection instance on instantiation
   - [ ] Has helper methods for manipulating/processing the collection
   - [ ] Get me the relevant ordered sections for this collection
   - [ ] Get me the relevant ordered questions for this section
   - [ ] Working out the status of a section / form. This can just hardcode ‘Not started’ for now, logic to be added later.
 - [ ] The section names in the tasklist are dead links (ie href=#)

### DIFF
--- a/app/common/data/interfaces/temporary.py
+++ b/app/common/data/interfaces/temporary.py
@@ -1,0 +1,30 @@
+"""
+This module contains temporary functions that are used to support the scaffolding of the core platform functionality.
+
+We anticipate that they should not be used by any of the real domains themselves (eg apply, assess, monitor), but are
+required to support the technical build-out of the platform.
+
+This file should be removed once the scaffolding is complete and some domain skins are in place.
+
+The only place that should import from here is the `app.developers` package.
+"""
+
+from pydantic import UUID4
+from sqlalchemy import delete
+
+from app.common.data.models import (
+    Collection,
+    CollectionSchema,
+)
+from app.extensions import db
+
+
+def delete_collections_created_by_user(*, grant_id: UUID4, created_by_id: UUID4) -> None:
+    db.session.execute(
+        delete(Collection).where(
+            Collection.collection_schema_id == CollectionSchema.id,
+            CollectionSchema.grant_id == grant_id,
+            Collection.created_by_id == created_by_id,
+        )
+    )
+    db.session.flush()
--- a/app/common/data/interfaces/temporary.py
+++ b/app/common/data/interfaces/temporary.py
@@ -9,7 +9,8 @@ This file should be removed once the scaffolding is complete and some domain ski
 The only place that should import from here is the `app.developers` package.
 """
 
-from pydantic import UUID4
+from uuid import UUID
+
 from sqlalchemy import delete
 
 from app.common.data.models import (
@@ -19,7 +20,7 @@ from app.common.data.models import (
 from app.extensions import db
 
 
-def delete_collections_created_by_user(*, grant_id: UUID4, created_by_id: UUID4) -> None:
+def delete_collections_created_by_user(*, grant_id: UUID, created_by_id: UUID) -> None:
     db.session.execute(
         delete(Collection).where(
             Collection.collection_schema_id == CollectionSchema.id,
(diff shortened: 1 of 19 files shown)


app/common/data/models.py
```diff
@@ -155,6 +157,32 @@ class CollectionSchema(BaseModel):
...
 
 
+class CollectionStatusEnum(enum.StrEnum):
+    NOT_STARTED = "Not started"
+
+
+class Collection(BaseModel):
+    __tablename__ = "collection"
+
+    data: Mapped[json_scalars]
```
sfount, line 167: I can't think of anything else this would be, tiny simplifier for the interface
```suggestion
    data: Mapped[json_scalars] = mapped_column(default={})
```

app/common/data/interfaces/temporary.py
```diff
@@ -0,0 +1,30 @@
...
+
+We anticipate that they should not be used by any of the real domains themselves (eg apply, assess, monitor), but are
+required to support the technical build-out of the platform.
+
+This file should be removed once the scaffolding is complete and some domain skins are in place.
+
+The only place that should import from here is the `app.developers` package.
+"""
+
+from pydantic import UUID4
```
sfount, line 12: Are we intentionally using pydantics UUID4 vs. the uuid lib? Sorry if I've missed something and thats obvious

app/common/data/models.py
```diff
@@ -49,6 +50,7 @@ class User(BaseModel):
     magic_links: Mapped[list["MagicLink"]] = relationship("MagicLink", back_populates="user")
 
     roles: Mapped[list["UserRole"]] = relationship("UserRole", back_populates="user", cascade="all, delete-orphan")
+    collections: Mapped[list["Collection"]] = relationship("Collection", back_populates="created_by")
```
sfount, line 53: Edit: this is the instances a user has created, not the collection schema

Original: I can't anticipate a read model where we'll ever need this back populated when fetching a user - maybe it can't hurt to have it but it would probably be good to be explicit about it being lazy as we'll _always_ need to fetch a user and very rarely care about what collections they've set up

sfount (approved):
A few very small comments, this looks sensible and all good for it to be brought in - if we're calling the instances of data "collections" I'm going to need to do some mindfulness and rewrite my thinking

app/common/data/interfaces/temporary.py
samuelhwilliams: Hm good query. I've seen this done elsewhere and didn't query it 🤦‍♂️  I don't think using this in type hints actually gives us anything over `uuid.UUID` outside of DB models/pydantic classes. Might be one to sweep up later.

app/common/data/types.py
```diff
@@ -1 +1,3 @@
-json_scalars = str | int | float | bool
+from typing import Dict
+
+json_scalars = Dict[str, str | int | float | bool]
```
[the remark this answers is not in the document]
samuelhwilliams: Very much shouldn't have touched that in this PR, leave it for later when we have concrete ideas 👀

PR #205, opened 2025-05-23
Lint imports to maintain separation

samuelhwilliams:
We have some rules about how parts of the app should interact with other parts of the app. At the moment these are mostly in our head:

- Database models should not be accessed directly in the majority of the app; queries should be written and exposed as interfaces which eg HTTP handlers can use.
- Our `developers` package is intended to be thrown away in the future; we don't want anything there leaking out into the rest of the codebase.

I've had to move the User models out into a separate file, because we _do_ import these in a few places across the codebase - not to use the models directly, but to cast Flask's `current_user` to our `User` object so that we get type hinting and intellisense. This is sub-optimal, but I think we can live with it for now...

This patch integrates [import-linter](https://import-linter.readthedocs.io/en/stable/readme.html) in our pre-commit and CI checks to detect any imports that violate the above rules.

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -1,8 +1,11 @@
+from typing import TYPE_CHECKING
 from uuid import UUID
 
 from app.common.data.interfaces.collections import get_collection
-from app.common.data.models import Collection, CollectionSchema, Form, Grant, Question, Section
 from app.common.data.types import CollectionStatusEnum
+
+if TYPE_CHECKING:
+    from app.common.data.models import Collection, CollectionSchema, Form, Grant, Question, Section
 
 
 class CollectionHelper:
@@ -13,7 +16,7 @@ class CollectionHelper:
     conditionals, routing, storing+retrieving data, etc in one place, consistently.
     """
 
-    def __init__(self, collection: Collection):
+    def __init__(self, collection: "Collection"):
         self._collection = collection
 
     @classmethod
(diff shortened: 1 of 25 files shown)


pyproject.toml
```diff
@@ -97,6 +97,31 @@ layers = [
...
+    "app.config",
+    "app.healthcheck",
+    "app.logging",
+    "app.monkeypatch",
+    "app.sentry",
+    "app.types",
+]
+# Unable to do `exhaustive` here unfortunately, so we won't get told if new modules are added and we forget to put them
+# in here. C'est la vie.
+# Importantly we don't block `app.common.data.models_user` because this is used in a lot of places to cast Flask's
```
MarcUsher, line 117: V small thing but the reason to not block `models_user` could be helpful to bump up to the commit message for future us?

app/common/helpers/collections.py
```diff
@@ -1,7 +1,8 @@
 from uuid import UUID
 
 from app.common.data.interfaces.collections import get_collection
-from app.common.data.models import Collection, CollectionSchema, CollectionStatusEnum, Form, Grant, Question, Section
+from app.common.data.models import Collection, CollectionSchema, Form, Grant, Question, Section
```
MarcUsher, line 4: Do we want to say anything explicit in commit/a comment as to why this module is allowed to import models (for type checking, it looks like)?

MarcUsher (approved):
This looks really useful, nice addition Sam! Will be super helpful as the codebase grows and a good thing for us to remember to check this in pyproject.toml as new modules get added and keeping these lists up to date.

Have just added 2 v small queries but they're not blockers, more general musings, so approving.

pyproject.toml
samuelhwilliams: v good call - let me do that 🙇

app/common/helpers/collections.py
samuelhwilliams: ah this mostly has slipped through as a consequence of how the import linter deals with things in the same hierarchy (app.common); I didn't set up any restrictions on anything in app.common using app.common.data.models, so that's all unchecked.

I can add that in but I think we'll hit a few more wanting-to-import-for-type-hint issues. Might need to see how this develops 👀

MarcUsher (approved):

PR #207, opened 2025-05-27
FSPT-412: Add redirect on auth routes if user is authenticated

MarcUsher:
Quality of life improvement to redirect authenticated users away from magic link and sso login routes.

This may need revisiting once we start accommodating more user types on the platform (eg. where to redirect different user types) but for now this is a helpful step to avoid confusion for authenticated users.


samuelhwilliams (approved):
Noice :)

PR #208, opened 2025-05-27
Render question pages for a form

samuelhwilliams:
## Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-435

## Description
Instrument the base logic for stepping through the questions in a form. The scope of this work is intentionally limited to focus mostly on building a framework for the business logic (eg conditionals, routing, validation, etc) to sit in later. All I'm doing here is letting you take a form, link to the first question page, render that, and then step through to the next question. Other than validation provided by WTForms natively (eg for numeric fields), we don't do anything special. Later tickets will focus on the validation stream. Persisting submitted data and cascading this through to the collection/form status is also part of future work.

### DIFF
--- a/app/common/collections/forms.py
+++ b/app/common/collections/forms.py
@@ -1,0 +1,52 @@
+from typing import cast
+
+from flask_wtf import FlaskForm
+from govuk_frontend_wtf.wtforms_widgets import GovSubmitInput, GovTextArea, GovTextInput
+from wtforms.fields.numeric import IntegerField
+from wtforms.fields.simple import StringField, SubmitField
+
+from app.common.data.models import Question
+from app.common.data.types import QuestionDataType
+
+_accepted_fields = StringField | IntegerField
+
+
+# FIXME: Ideally this would do an intersection between FlaskForm and QuestionFormProtocol, but type hinting in
+#        python doesn't currently support this. As of May 2025, it looks like we might be close to some progress on
+#        this in https://github.com/python/typing/issues/213.
+# This is a bit of a hack so that we have an externally-accessible type that represents the kind of form returned
+# by `build_question_form`. This gives us nicer intellisense/etc. The downside is that this class needs to be kept
+# in sync manually with the one inside `build_question_form`.
+class DynamicQuestionForm(FlaskForm):
+    question: _accepted_fields
+    submit: SubmitField
+
+
+def build_question_form(question: Question) -> DynamicQuestionForm:
+    # NOTE: Keep the fields+types in sync with the class of the same name above.
+    class _DynamicQuestionForm(FlaskForm):  # noqa
+        question: _accepted_fields
+        submit = SubmitField("Continue", widget=GovSubmitInput())
+
+    field: _accepted_fields
+    match question.data_type:
+        case QuestionDataType.TEXT_SINGLE_LINE:
+            field = StringField(label=question.text, description=question.hint or "", widget=GovTextInput())
+        case QuestionDataType.TEXT_MULTI_LINE:
+            field = StringField(
+                label=question.text,
+                description=question.hint or "",
+                widget=GovTextArea(),
+            )
+        case QuestionDataType.INTEGER:
+            field = IntegerField(
+                label=question.text,
+                description=question.hint or "",
+                widget=GovTextInput(),
+            )
+        case _:
+            raise Exception("Unable to generate dynamic form for question type {_}")
+
+    _DynamicQuestionForm.question = field
+
+    return cast(DynamicQuestionForm, _DynamicQuestionForm())
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -3,6 +3,7 @@ from uuid import UUID
 
 from sqlalchemy import select, text
 from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm import joinedload
 
 from app.common.data.interfaces.exceptions import DuplicateValueError
 from app.common.data.models import (
@@ -60,11 +61,27 @@ def update_collection_schema(schema: CollectionSchema, *, name: str) -> Collecti
     return schema
 
 
-def get_collection(collection_id: UUID) -> Collection:
-    return db.session.get_one(
-        Collection,
-        collection_id,
-    )
+def get_collection(collection_id: UUID, with_full_schema: bool = False) -> Collection:
+    options = []
+    if with_full_schema:
+        options.append(
+            joinedload(Collection.collection_schema)
+            .selectinload(CollectionSchema.sections)
+            .selectinload(Section.forms)
+            .selectinload(Form.questions)
+        )
+
+    # We set `populate_existing` here to force a new query to be emitted to the database. The mechanics of `get_one`
+    # relies on the session cache and does a lookup in the session memory based on the PK we're trying to retrieve.
+    # If the object exists, no query is emitted and the options won't take effect - we would fall back to lazy loading,
+    # which is n+1 select. If we don't care about fetching the full nested schema then it's fine to grab whatever is
+    # cached in the session alright, but if we do specifically want all of the related objects, we want to force the
+    # loading options above. This does mean that if you call this function twice with `with_full_schema=True`, it will
+    # do redundant DB trips. We should try to avoid that. =]
+    # If we took the principle that all relationships should be declared on the model as `lazy='raiseload'`, and we
+    # specify lazy loading explicitly at all points of use, we could potentially remove the `populate_existing`
+    # override below.
+    return db.session.get_one(Collection, collection_id, options=options, populate_existing=bool(options))
 
 
 def create_collection(*, schema: CollectionSchema, created_by: User) -> Collection:
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -1,11 +1,13 @@
-from typing import TYPE_CHECKING
+import uuid
+from itertools import chain
+from typing import TYPE_CHECKING, Optional
 from uuid import UUID
 
 from app.common.data.interfaces.collections import get_collection
 from app.common.data.types import CollectionStatusEnum
 
 if TYPE_CHECKING:
-    from app.common.data.models import Collection, CollectionSchema, Form, Grant, Question, Section
+    from app.common.data.models import Collection, Form, Grant, Question, Section
 
 
 class CollectionHelper:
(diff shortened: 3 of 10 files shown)


app/developers/templates/developers/ask_a_question.html
```diff
@@ -0,0 +1,39 @@
...
+
+{% block pageTitle %}
+  {{ question.text }} - MHCLG Funding Service
+{% endblock pageTitle %}
+
+{% block beforeContent %}
+  {{
+    govukBackLink({
+        "text": "Back",
+        "href": url_for("developers.manage_schema", grant_id=collection_helper.grant.id, schema_id=collection_helper.schema.id)
```
wjrm500, line 16: When I clicked "Back" on the second question page of a form I half-expected to go back to the first question page and was then mildly taken aback at landing on the schema page. Not sure if this is an issue

app/developers/routes.py
```diff
@@ -504,5 +506,35 @@ def edit_question(
...
-    collection_helper = CollectionHelper.load(collection_id)
+    collection, *_ = get_full_collection_schema(collection_id=collection_id)
+    collection_helper = CollectionHelper(collection)
     return render_template("developers/collection_tasklist.html", collection_helper=collection_helper)
+
+
+@developers_blueprint.route("/collections/<uuid:collection_id>/<uuid:question_id>", methods=["GET", "POST"])
+@platform_admin_role_required
+def ask_a_question(collection_id: UUID, question_id: UUID) -> ResponseReturnValue:
+    collection, schema, section, db_form, question = get_full_collection_schema(
```
wjrm500: Most of these are unused variables (inc. `schema`, `section` and `db_form`), would it not suffice to simply call `get_collection` then `get_question`, or are we thinking about future-proofing for breadcrumbs or the like? But then you can access higher levels of the hierarchy via the `Question` model e.g., `question.form.section.schema` (nice and readable obvs 😆)

app/common/helpers/collections.py
```diff
@@ -16,33 +18,61 @@ class CollectionHelper:
...
+            )
+        except StopIteration as e:
+            raise ValueError(f"Could not find a form with id={form_id} in schema={self.schema.id}") from e
+
+    def get_question(self, question_id: uuid.UUID) -> "Question":
+        try:
+            return next(
+                filter(
+                    lambda q: q.id == question_id,
+                    list(
```
wjrm500, line 66: Any reason why this is wrapped in `list` and the output from `chain.from_iterable` in `get_form` isn't?

app/common/helpers/collections.py
```diff
@@ -57,3 +87,34 @@ def get_ordered_visible_forms_for_section(self, section: "Section") -> list["For
...
         """Returns the visible, ordered questions for a given form based upon the current state of this collection."""
         return sorted(form.questions, key=lambda q: q.order)
+
+    def get_first_question_for_form(self, form: "Form") -> Optional["Question"]:
+        questions = self.get_ordered_visible_questions_for_form(form)
+        if questions:
+            return questions[0]
+        return None
+
+    def get_form_for_question(self, question_id: UUID) -> "Form":
```
wjrm500, line 97: Quick Q - how do we decide when to put something in this helper class vs in `data/interfaces/collections.py`? Conscious that we are including new methods in this class that aren't `Collection`-specific, such as this one

app/developers/routes.py
```diff
@@ -511,3 +512,28 @@ def edit_question(
 def collection_tasklist(collection_id: UUID) -> ResponseReturnValue:
     collection_helper = CollectionHelper.load(collection_id)
     return render_template("developers/collection_tasklist.html", collection_helper=collection_helper)
+
+
+@developers_blueprint.route("/collections/<uuid:collection_id>/<uuid:question_id>", methods=["GET", "POST"])
+@platform_admin_role_required
+def ask_a_question(collection_id: UUID, question_id: UUID) -> ResponseReturnValue:
+    collection_helper = CollectionHelper(collection_id)
```
wjrm500, line 520: Should we use the new `load` method to avoid going directly via the constructor?

app/developers/routes.py
```diff
@@ -511,3 +512,28 @@ def edit_question(
...
+    question = collection_helper.get_question(question_id)
+    form = build_question_form(question)
```
wjrm500, line 522: This bakes in the one question per page thing, just checking that is a concrete design decision. I mean it keeps things straightforward which is nice, so maybe regardless we can just move forward and face up to the pain of a change later if needed. The alternative is designing it in a more flexible way (possibility of multiple questions per page) from the off, and then having some "max questions per page" variable or the like set to 1 to effectively enforce that design constraint while providing a simple way to drop it in future

app/developers/templates/developers/ask_a_question.html
```diff
@@ -0,0 +1,39 @@
...
+
+{% block content %}
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <div class="app-body__watermark" aria-hidden="true">Developers</div>
+
+    <form method="post" novalidate>
+      {{ form.csrf_token }}
+
+      {% if question.data_type == question_types.TEXT_SINGLE_LINE or question.data_type == question_types.TEXT_MULTI_LINE %}
```
wjrm500, line 29: Obviously this `if` could get quite long, although I suppose length isn't the same as complexity. And we have to do this conditional switch _somewhere_, so it may as well be in this pretty bare template? This is just me rambling, feel free to ignore

wjrm500 (commented):
I've had a good look and left some thoughts but overall looks polished, definitely agree it's good to lay down some foundational code here to support future feature work. Exactly what we need to be doing right now to push on

app/common/helpers/collections.py
samuelhwilliams: My finger in the air:
- `interfaces` for retrieving the data from the DB
- this helper class for doing work on the retrieved data to do useful things - eg navigating across relationships, filtering collections for specific instances, small bits of business logic potentially etc.

app/common/helpers/collections.py
```diff
@@ -16,33 +18,61 @@ class CollectionHelper:
...
     @property
     def name(self) -> str:
-        return self._collection.collection_schema.name
+        return self.schema.name
 
     @property
     def status(self) -> str:
         return CollectionStatusEnum.NOT_STARTED
 
+    def get_section(self, section_id: uuid.UUID) -> "Section":
```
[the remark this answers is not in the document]
samuelhwilliams: I think not - at least not yet. Readability/simplicity over a 🤏  bit of localised duplication.

app/common/helpers/collections.py
```diff
@@ -1,11 +1,10 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional
```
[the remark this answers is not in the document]
samuelhwilliams: (P.S. I _think_ this problem goes away in python 3.14: https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-pep649)

app/common/collections/forms.py
```diff
@@ -0,0 +1,52 @@
...
+from wtforms.fields.numeric import IntegerField
+from wtforms.fields.simple import StringField, SubmitField
+
+from app.common.data.models import Question
+from app.common.data.types import QuestionDataType
+
+_accepted_fields = StringField | IntegerField
+
+
+# FIXME: Ideally this would do an intersection between FlaskForm and QuestionFormProtocol, but type hinting in
```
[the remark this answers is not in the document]
sfount: Does WTForm type the `data` property that is used after a validate? I guess thats the part of the signature we'd be interested in outside of it being a flask form

Quite a few things crossing over here so happy to chat it through! definitely nothing blocking for this PR

app/common/data/interfaces/collections.py
```diff
@@ -60,11 +61,27 @@ def update_collection_schema(schema: CollectionSchema, *, name: str) -> Collecti
...
-    return db.session.get_one(
-        Collection,
-        collection_id,
-    )
+def get_collection(collection_id: UUID, with_full_schema: bool = False) -> Collection:
+    options = []
+    if with_full_schema:
+        options.append(
+            joinedload(Collection.collection_schema)
+            .selectinload(CollectionSchema.sections)
```
sfount, line 69: I'm sure you've tried a few options here but I think everything down to questions (maybe we'd want to stop at forms intentionally) here could be a single joined query 

I think there will be some changes around how sections are modeled and a few extra things to track as we factor in groups so no need to get this perfect from the get go - a good one to chat about when we get a minute though

app/common/helpers/collections.py
```diff
@@ -17,32 +19,61 @@ class CollectionHelper:
...
-        return self._collection.collection_schema.name
+        return self.schema.name
 
     @property
     def status(self) -> str:
         return CollectionStatusEnum.NOT_STARTED
 
+    def get_section(self, section_id: uuid.UUID) -> "Section":
+        try:
+            return next(filter(lambda s: s.id == section_id, self.schema.sections))
```
sfount, line 52: Just for my curiosity do you know if there's a performance or practical difference between `filter()` and lambda or list comprehension syntax? I've normally seen this written like

```python
next(s for s in self.schema.sections if s.id == section_id)
```

I'm fairly confident python is doing something clever behind the scenes between expanding the list and an iterable but I've not dug into it!

sfount (approved):
I believe the sql query counting assertions are being moved to another PR to work out the details so they'll either need to be removed from this PR and added in another or this can be rebased to that in order (https://github.com/communitiesuk/funding-service/pull/212)

Code looks good to me, a few thoughts throughout but this looks good to come in once that above is actioned and tests pass

app/common/data/interfaces/collections.py
```diff
@@ -60,11 +61,27 @@ def update_collection_schema(schema: CollectionSchema, *, name: str) -> Collecti
...
+        )
+
+    # We set `populate_existing` here to force a new query to be emitted to the database. The mechanics of `get_one`
+    # relies on the session cache and does a lookup in the session memory based on the PK we're trying to retrieve.
+    # If the object exists, no query is emitted and the options won't take effect - we would fall back to lazy loading,
+    # which is n+1 select. If we don't care about fetching the full nested schema then it's fine to grab whatever is
+    # cached in the session alright, but if we do specifically want all of the related objects, we want to force the
+    # loading options above. This does mean that if you call this function twice with `with_full_schema=True`, it will
+    # do redundant DB trips. We should try to avoid that. =]
+    # If we took the principle that all relationships should be declared on the model as `lazy='raiseload'`, and we
```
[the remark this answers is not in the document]
samuelhwilliams: Yep I think we could definitely have some metatest on query counts - will make a stub ticket for that.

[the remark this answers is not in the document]
sfount: 🙇

sfount (approved):
Looking good! Possibly a few small comments about typing and how question type drives forms + models + data and how they join together but I think a lot of that will come out of submitting data next

PR #213, opened 2025-05-28
FSPT-509 - Add a grant button and basic question flow

wjrm500:
### Description

This change changes the content on the list grants page in line with updated designs, and replaces the single grant name input page with a series of pages, asking for things like GGIS number, name, description, contact name and contact email. Data is saved in the session on 'Save and continue', meaning that the Grant is only persisted when the final form page is submitted.

### Divergence from designs and reason for this

- **Design**: on the final form page, the button says "Save and continue"
- **Implementation**: the button says "Add grant"
- **Reason for difference**: "Save and continue" only makes sense as part of the final flow including a "Check your answers" page
---
- **Design**: for the grants list page, there is no specific design given for a scenario where the user has already set up a grant
- **Implementation**: we list existing grants as hyperlinks under the "Add a grant" button, which is basically what we already do
- **Reason for difference**: need some way of showing existing grants and this seems an inoffensive placeholder solution

### Tech decisions taken and reasoning

- **Storing data in session then persisting on final form submit**: this means Grants are packaged as a single entity with end-state validation on all fields, and no leakage of unfinished or unwanted Grants
- **Minimal abstraction**: we touched on the level of abstraction we're looking for in a couple of chats about this work. I suggested that we might use the form runner tool currently under development to actually render this form, and while this remains an option, legitimate concerns have been raised around over-abstraction. In the interests of not duplicating abstraction work elsewhere and ending up with competing abstraction systems, I've gone the other way here with very individual templates, FlaskForm classes and route functions. If we ever want to abstract in the future, we can abstract based on the general form runner system
- **No unique constraint on GGIS number**: nothing specific was mentioned about this and validating uniqueness is slightly more involved, so just erred on the side of not doing this initially. Probably makes sense to do this though so happy to put up a second commit

### How to test

- Run app locally (ping me on Slack if need help with this)
- Sign in at https://funding.communities.gov.localhost:8080/sso/sign-in, you will be redirected to `/grants` on success
- Click "Add a grant" and follow the pages through

### Screenshots

image

image

image

image

image

image


app/common/data/migrations/versions/004_add_detailed_grant_fields.py
```diff
@@ -0,0 +1,38 @@
...
+from alembic import op
+
+revision = "004_add_detailed_grant_fields"
+down_revision = "003_nulls_not_distinct_user_role"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
samuelhwilliams, line 20: Can we drop out these comments please 🙏

app/common/data/models.py
```diff
@@ -21,6 +21,10 @@ class Grant(BaseModel):
     __tablename__ = "grant"
 
     name: Mapped[CIStr] = mapped_column(unique=True)
+    ggis_number: Mapped[str | None] = mapped_column(nullable=True)
+    description: Mapped[str | None] = mapped_column(nullable=True)
+    primary_contact_name: Mapped[str | None] = mapped_column(nullable=True)
+    primary_contact_email: Mapped[str | None] = mapped_column(nullable=True)
```
samuelhwilliams, line 27:
```suggestion
    ggis_number: Mapped[str | None]
    description: Mapped[str | None]
    primary_contact_name: Mapped[str | None]
    primary_contact_email: Mapped[str | None]
```

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/ggis_number.html
```diff
@@ -0,0 +1,75 @@
...
+  }}
+{% endblock beforeContent %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-xl">
+        <span class="govuk-visually-hidden">This section is: </span>Add a grant
+      </span>
+      <h1 class="govuk-heading-xl">Do you have a Government Grants Information System (GGIS) reference number?</h1>
```
samuelhwilliams, line 27: Please can we use the one-question-per-page/isPageHeading approach from the design system: https://design-system.service.gov.uk/components/text-input/

Rather than us duplicating the content for this question in three places 🙇

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/contact.html
```diff
@@ -0,0 +1,63 @@
+{% extends "common/base.html" %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{% from "govuk_frontend_jinja/components/error-summary/macro.html" import govukErrorSummary %}
+
+{% block pageTitle %}
+  Who is the main contact for this grant? - Add a grant - MHCLG Funding Service
```
samuelhwilliams, line 6: again this duplication - we have this canonically in the form field definition, please let's pull from there 🙇

app/deliver_grant_funding/routes.py
```diff
@@ -9,27 +9,114 @@
...
+        return redirect(url_for("deliver_grant_funding.grant_setup_intro"))
+
+    form_data = session["grant_setup"].get("ggis", {})
+    form = GrantGGISForm(data=form_data)
+
+    if form.validate_on_submit():
+        session["grant_setup"]["ggis"] = {
+            "has_ggis": form.has_ggis.data,
+            "ggis_number": form.ggis_number.data if form.has_ggis.data == "yes" else None,
+        }
```
samuelhwilliams, line 48: can we do something clever here with maybe a dataclass/Pydantic model that gives us confidence in the expected shape/structure of this `ggis` blob? I think having an arbitrary blob of data in the session creeps me out a bit, and knowing we have something nicely typed/defined that can (de)serialize this would be nice.

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -648,3 +641,113 @@ def test_edit_question_post(authenticated_platform_admin_client, factories, db_s
     assert question_from_db.text == "Updated Question"
     assert question_from_db.hint == "Updated Hint"
     assert question_from_db.name == "Updated Question Name"
+
+
+def test_grant_setup_flow_with_ggis_and_description_error(authenticated_client, db_session, templates_rendered):
```
samuelhwilliams, line 646: This does not feel like a nicely encapsulated integration test - this is much more of an e2e test.

The rest of the integration tests here are much more about load one page, assert on one page.

If we're stepping through entire journeys then I think we want to do that in e2e tests.

app/common/data/models.py
wjrm500: Done [here](https://github.com/communitiesuk/funding-service/pull/213/commits/12eb29ccf4ae482897d2ac667749b82a62a543c7)

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/ggis_number.html
wjrm500: Done [here](https://github.com/communitiesuk/funding-service/pull/213/commits/fcfa6f6b3cbec1e2afb67cc6a8205d42bc30b1d0)

app/deliver_grant_funding/forms.py
```diff
@@ -21,6 +21,79 @@ class GrantForm(FlaskForm):
...
+    submit = SubmitField("Continue", widget=GovSubmitInput())
+
+
+class GrantGGISForm(FlaskForm):
+    has_ggis = RadioField(
+        "Do you have a Government Grants Information System (GGIS) reference number?",
+        choices=[("yes", "Yes"), ("no", "No")],
+        validators=[DataRequired("Select yes if you have a GGIS reference number")],
+        widget=GovRadioInput(),
+        default="no",
```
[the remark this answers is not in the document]
wjrm500: Done [here](https://github.com/communitiesuk/funding-service/pull/213/commits/12eb29ccf4ae482897d2ac667749b82a62a543c7)

app/deliver_grant_funding/routes.py
```diff
@@ -9,27 +9,114 @@
...
 
 
-@deliver_grant_funding_blueprint.route("/grants/set-up", methods=["GET", "POST"])
+@deliver_grant_funding_blueprint.route("/grants/add", methods=["GET", "POST"])
+@mhclg_login_required
+def grant_setup_intro() -> ResponseReturnValue:
+    form = GrantSetupIntroForm()
+    if form.validate_on_submit():
+        session.pop("grant_setup", None)
+        session["grant_setup"] = {}
```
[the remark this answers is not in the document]
wjrm500: Indeed we do not, fixed as part of [this commit](https://github.com/communitiesuk/funding-service/pull/213/commits/3474f32f49829dae00bd71f14b6b6979c53d9a45)

app/common/data/models.py
```diff
@@ -21,6 +21,10 @@ class Grant(BaseModel):
     __tablename__ = "grant"
 
     name: Mapped[CIStr] = mapped_column(unique=True)
+    ggis_number: Mapped[str | None] = mapped_column(nullable=True)
```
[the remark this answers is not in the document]
wjrm500: Done [here](https://github.com/communitiesuk/funding-service/pull/213/commits/12eb29ccf4ae482897d2ac667749b82a62a543c7)

app/deliver_grant_funding/routes.py
```diff
@@ -9,27 +9,114 @@
...
+    GrantNameSetupForm,
+    GrantSetupIntroForm,
 )
 from app.extensions import auto_commit_after_request
 
 deliver_grant_funding_blueprint = Blueprint(name="deliver_grant_funding", import_name=__name__)
 
 
-@deliver_grant_funding_blueprint.route("/grants/set-up", methods=["GET", "POST"])
+@deliver_grant_funding_blueprint.route("/grants/add", methods=["GET", "POST"])
```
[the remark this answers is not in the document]
wjrm500: Done [here](https://github.com/communitiesuk/funding-service/pull/213/commits/12eb29ccf4ae482897d2ac667749b82a62a543c7)

app/deliver_grant_funding/routes.py
```diff
@@ -9,27 +9,114 @@
...
+    if "grant_setup" not in session:
+        return redirect(url_for("deliver_grant_funding.grant_setup_intro"))
+
+    form_data = session["grant_setup"].get("name", {})
+    form = GrantNameSetupForm(data=form_data)
+
+    if form.validate_on_submit():
+        session["grant_setup"]["name"] = {"name": form.name.data}
+        session.modified = True
+        return redirect(url_for("deliver_grant_funding.grant_setup_description"))
```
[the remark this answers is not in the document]
wjrm500: Very fair, fixed [here](https://github.com/communitiesuk/funding-service/pull/213/commits/edfad05622851e004e1ce03f387c3fb538e734d5)

app/deliver_grant_funding/routes.py
wjrm500: We can indeed, see [here](https://github.com/communitiesuk/funding-service/pull/213/commits/3474f32f49829dae00bd71f14b6b6979c53d9a45)

Ended up implementing this not just for GGIS data specifically but for all grant setup session data

tests/integration/deliver_grant_funding/test_routes.py
wjrm500: Happy to break this up, see this [commit](https://github.com/communitiesuk/funding-service/pull/213/commits/7514b4d9b2667b9dc3d9a4c7c6db42ee049c6939)

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/contact.html
wjrm500: We don't actually have this text in the FlaskForm definition

app/common/data/migrations/versions/004_add_detailed_grant_fields.py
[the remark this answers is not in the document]
wjrm500: Done [here](https://github.com/communitiesuk/funding-service/pull/213/commits/12eb29ccf4ae482897d2ac667749b82a62a543c7)

app/common/data/interfaces/grants.py
```diff
@@ -13,13 +13,32 @@ def get_grant(grant_id: UUID) -> Grant:
     return db.session.get_one(Grant, grant_id)
 
 
+def grant_name_exists(name: str) -> bool:
+    """Check if a grant with the given name exists."""
+    statement = select(Grant).where(Grant.name == name)
+    grant: Grant | None = db.session.scalar(statement)
```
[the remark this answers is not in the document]
wjrm500: You're right, have changed

app/common/data/migrations/versions/004_add_detailed_grant_fields.py
```diff
@@ -0,0 +1,32 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("grant", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("ggis_number", sa.String(), nullable=True))
+        batch_op.add_column(sa.Column("description", sa.String(), nullable=True))
+        batch_op.add_column(sa.Column("primary_contact_name", sa.String(), nullable=True))
+        batch_op.add_column(sa.Column("primary_contact_email", sa.String(), nullable=True))
```
nuwan-samarasinghe: Probably these cannot be nullable ? I think why we added this is since we going though a step by step process to update the grant details we are getting a non null error from the database, better to think about a temporary data table concept to store grants if that is not yet completed till the end 

@samuelhwilliams @sfount let me know your thoughts just my thinking

app/common/forms/validators.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, max_words: int, message: str | None = None) -> None:
+        self.max_words = max_words
+        self.message = message
+
+    def __call__(self, form: BaseForm, field: Field) -> None:
+        if not field.data:
+            return  # Don't validate empty fields - use DataRequired for that
+
+        words = field.data.split()
```
nuwan-samarasinghe, line 23: Here it will just check for spaces I guess what if a user types a sentence like this 

`What is your name?and please give your initials`

`What is your name.and please give your initials`

then word count might be different check this on all the validators that we have over here

app/deliver_grant_funding/forms.py
```diff
@@ -21,6 +40,71 @@ class GrantForm(FlaskForm):
...
+        validators=[DataRequired("Select yes if you have a GGIS reference number")],
+        widget=GovRadioInput(),
+    )
+    ggis_number = StringField(
+        "Enter your GGIS reference number",
+        validators=[Optional()],
+        filters=[strip_string_if_not_empty],
+        widget=GovTextInput(),
+    )
+    submit = SubmitField("Save and continue", widget=GovSubmitInput())
```
nuwan-samarasinghe, line 60: Since this submit button has on all the forms cant we have a parent.. just a thought if no need ignore

wjrm500:
Morning @samuelhwilliams and @sfount, I've addressed all of your comments and the work is now ready for re-review at your leisure 🙇

app/deliver_grant_funding/routes.py
```diff
@@ -1,35 +1,143 @@
...
+    return render_template("deliver_grant_funding/grant_setup/name.html", form=form)
+
+
+@deliver_grant_funding_blueprint.route("/grants/setup/description", methods=["GET", "POST"])
+@platform_admin_role_required
+def grant_setup_description() -> ResponseReturnValue:
+    if "grant_setup" not in session:
+        return redirect(url_for("deliver_grant_funding.grant_setup_intro"))
+
+    grant_session = GrantSetupSession.from_session(session["grant_setup"])
```
nuwan-samarasinghe, line 90: Since we do the same but slight diff is there cant we generalise this.. again a thought ignore if not needed

app/common/forms/validators.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class MaxWords:
+    """
+    Validates that a field contains no more than a specified number of words.
+
+    Args:
+        max_words (int): Maximum number of words allowed
+        message (str, optional): Custom error message. If not provided, a default message is used.
+    """
```
samuelhwilliams, line 13: Don't love this comment either - the args section is fully redundant with the type hints. We should rely on the type hint information.

tests/integration/models.py
```diff
@@ -37,6 +37,9 @@ class Meta:
 
     id = factory.LazyFunction(uuid4)
     name = factory.Sequence(lambda n: "Grant %d" % n)
+    description = factory.Faker("text", max_nb_chars=200)
+    primary_contact_name = factory.Faker("name")
+    primary_contact_email = factory.Faker("email")
```
samuelhwilliams, line 42: Can we add all of the new fields please, ie also ggis number?

app/deliver_grant_funding/forms.py
wjrm500: Ah fair, I think this comment and the below are thematically tied, you're right we can save on lines of code but perhaps at the expense of readability, I'd prefer to have a bit of duplication than start building ourselves a jungle

app/deliver_grant_funding/routes.py
wjrm500: Re abstraction, please see this comment: https://github.com/communitiesuk/funding-service/pull/213#discussion_r2114041556

Thank you kindly 🙏

app/common/data/interfaces/grants.py
```diff
@@ -13,13 +13,31 @@ def get_grant(grant_id: UUID) -> Grant:
...
 
-def create_grant(name: str) -> Grant:
-    grant: Grant = Grant(name=name)
+def create_grant(
+    name: str,
+    description: str,
+    primary_contact_name: str,
+    primary_contact_email: str,
+    ggis_number: str | None = None,
+) -> Grant:
```
samuelhwilliams, line 33:
```suggestion
def create_grant(
    *,
    name: str,
    description: str,
    primary_contact_name: str,
    primary_contact_email: str,
    ggis_number: str | None = None,
) -> Grant:
```

When there are more than 2-3 arguments for the interface, we've been tending to force them over to kwargs to make it more explicit that things are being passed into the right places.

app/common/forms/validators.py
```diff
@@ -0,0 +1,63 @@
...
+    def __call__(self, form: BaseForm, field: Field) -> None:
+        if not field.data:
+            return  # Don't validate empty fields - use DataRequired for that
+
+        words = field.data.split()
+        word_count = len(words)
+
+        if word_count < self.min_words or word_count > self.max_words:
+            message = self.message or f"Must be between {self.min_words} and {self.max_words} words"
+            raise ValidationError(message)
```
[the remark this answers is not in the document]
wjrm500: Would you like me to combine them?

app/deliver_grant_funding/forms.py
```diff
@@ -1,16 +1,46 @@
...
+class ConditionalGGISRequired:
+    """Validator to ensure GGIS number is provided when 'yes' is selected."""
+
+    def __init__(self, message: str | None = None):
+        self.message = message or "Enter your GGIS reference number"
+
+    def __call__(self, form: FlaskForm, field: StringField) -> None:
+        if form.has_ggis.data == "yes" and (not field.data or not field.data.strip()):
+            raise ValidationError(self.message)
+
```
samuelhwilliams, line 42: Let's do this on the specific form itself - as this isn't reusable across forms, it probes `has_ggis` directly which is not gonna exist always.

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/contact.html
```diff
@@ -0,0 +1,59 @@
...
+      <p class="govuk-body">This is the person that teams at MHCLG can contact if they have any questions about the grant.</p>
+
+      <form method="post" novalidate>
+        {{ form.csrf_token }}
+
+        {{
+          form.primary_contact_name(params={
+            "label": {
+              "text": form.primary_contact_name.label.text,
+            },
```
samuelhwilliams, line 37:
```suggestion
```

This is redundant - gets fed through by govuk-frontend-wtf automatically :)

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
```diff
@@ -0,0 +1,60 @@
...
+    })
+  }}
+{% endblock beforeContent %}
+
+{% set before_input_text %}
+  <p class="govuk-body govuk-!-margin-top-7">
+    Provide a brief description of the main purpose of the grant. This information will be seen by potential grant recipients.
+  </p>
+  <p class="govuk-hint">
+    Do not include personal information
```
samuelhwilliams, line 25:
```suggestion
    Do not include personal information.
```

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
```diff
@@ -0,0 +1,60 @@
...
+        <span class="govuk-visually-hidden">This section is: </span>Set up the grant
+      </span>
+
+      <form method="post" novalidate>
+        {{ form.csrf_token }}
+
+        {{
+          form.description(params={
+            "label": {
+              "text": form.description.label.text,
```
samuelhwilliams, line 42:
```suggestion
```

samuelhwilliams: I _think_

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
```diff
@@ -0,0 +1,60 @@
...
+              "isPageHeading": true,
+              "classes": "govuk-fieldset__legend--xl"
+            },
+            "fromGroup": true,
```
samuelhwilliams, line 46: Maybe a small comment and link to the PR fixing it?

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/ggis_number.html
```diff
@@ -0,0 +1,83 @@
+{% extends "common/base.html" %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{% from "govuk_frontend_jinja/components/error-summary/macro.html" import govukErrorSummary %}
+
+{% block pageTitle %}
+  Do you have a GGIS reference number? - Add a grant - MHCLG Funding Service
```
samuelhwilliams, line 6:
```suggestion
  {{ {{ form.has_ggis.label.text }} - Add a grant - MHCLG Funding Service
```

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/ggis_number.html
```diff
@@ -0,0 +1,83 @@
...
+  {{
+    govukBackLink({
+        "text": "Back",
+        "href": url_for("deliver_grant_funding.grant_setup_intro")
+    })
+  }}
+{% endblock beforeContent %}
+
+{% set before_inputs_text %}
+  <p class="govuk-body govuk-!-margin-top-7">
```
samuelhwilliams, line 21: Let's please remove the custom margins here - trust the stuff that gets spit out by GOV.UK Frontend and govuk-frontend-jinja/wtf natively unless there is a really clear directive that we need to tweak padding/margins/layout for some reason. If that's the case, let's probe that a little more and think about how we can make sure it's being applied consistently in the future.

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/ggis_number.html
```diff
@@ -0,0 +1,83 @@
...
+              },
+            },
+            "formGroup": {
+              "beforeInputs": {
+                "html": before_inputs_text
+              }
+            },
+            "items": [
+              {
+                "value": "yes",
```
samuelhwilliams, line 52: Can we add a comment to the form field for this, where we declare the choices in a somewhat-duplicative way - letting developers know they need to be kept in sync manually.

This is complex because govuk-frontend-wtf doesn't seem to support conditionals in a nice way.

Hardcoding the values here will affect the browser and what the users sends to the server, but the choices hardcoded in the form will affect validation and what we actually expect. There's a sync issue here.

app/common/forms/validators.py
```diff
@@ -0,0 +1,25 @@
...
```
samuelhwilliams, line 25: I maybe wasn't super clear, sorry - you asked a clarification question and I maybe misheard. I think this needs a bit more of an update to handle the cases that the individual MinWords and MaxWords handled - ie if you only provide `max_words` then the error message should be standardised for that case. The interface for it can have min and max as optional, and behaviour based on the combination of those (with neither provided being an error state). Also briefly mentioned the idea of taking a `thing` to inject into the message - and we can then probably just not allowed the message to be overridden for now.

So min = 5, max = None, thing = "grant name", leads to message = "Grant name must be at least 5 words".

Also let's update the error messages to match the suggested errors in the design system, since they provide some suitable ones:

https://design-system.service.gov.uk/components/text-input/#error-messages
image

app/deliver_grant_funding/routes.py
```diff
@@ -1,35 +1,127 @@
...
-            interfaces.grants.create_grant(name=form.name.data)
-            return redirect(url_for("deliver_grant_funding.list_grants"))
+            assert form.primary_contact_name.data, "Primary contact name must be provided"
+            assert form.primary_contact_email.data, "Primary contact email must be provided"
+
+            grant_session.primary_contact_name = form.primary_contact_name.data
+            grant_session.primary_contact_email = form.primary_contact_email.data
+
+            grant = interfaces.grants.create_grant(
+                name=grant_session.name or "",
```
samuelhwilliams, line 109: optional: Are there tweaks to the pydantic model we can do that make this `or` unnecessary? Eg giving the fields a default value of `""` or something?

app/deliver_grant_funding/routes.py
```diff
@@ -1,35 +1,127 @@
...
     if form.validate_on_submit():
         try:
-            assert form.name.data is not None
-            interfaces.grants.create_grant(name=form.name.data)
-            return redirect(url_for("deliver_grant_funding.list_grants"))
+            assert form.primary_contact_name.data, "Primary contact name must be provided"
+            assert form.primary_contact_email.data, "Primary contact email must be provided"
+
+            grant_session.primary_contact_name = form.primary_contact_name.data
+            grant_session.primary_contact_email = form.primary_contact_email.data
```
samuelhwilliams, line 106: Can we use `form.populate_obj(grant_session)` here? That feels like the inverse of `obj=grant_session` that we do on instantiation, so the symmetry feels like a useful signal - if it works.

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
```diff
@@ -0,0 +1,61 @@
+{% extends "common/base.html" %}
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
+{% from "govuk_frontend_jinja/components/error-summary/macro.html" import govukErrorSummary %}
+
+{% block pageTitle %}
+  {{ form.description.label.text }}
```
samuelhwilliams, line 6: One that we can address separately but the titles should keep the service name in. We're probably not consistent about this yet so I'll raise a separate ticket for it.

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
```diff
@@ -0,0 +1,61 @@
...
+  <p class="govuk-hint">
+    Do not include personal information.
+  </p>
+{% endset %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-xl">
+        <span class="govuk-visually-hidden">This section is: </span>Set up the grant
```
samuelhwilliams, line 33: Curious on the visually hidden text idea - are we dealing with 'sections' here? 👀

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
```diff
@@ -0,0 +1,61 @@
...
+      </span>
+
+      <form method="post" novalidate>
+        {{ form.csrf_token }}
+
+        <!-- Erroneous-looking fromGroup attribute currently required due to typo in library -->
+        <!-- See here: https://github.com/LandRegistry/govuk-frontend-jinja/blob/45849744d2bf058937402f291e4a2c7a44eb7e23/govuk_frontend_jinja/templates/components/character-count/macro.html#L107 -->
```
samuelhwilliams, line 40: I think a link to the PR would also be useful to help track 'solved'ness: https://github.com/LandRegistry/govuk-frontend-jinja/pull/96

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -22,8 +29,8 @@ def test_list_grants(app, authenticated_client, factories, templates_rendered, t
     assert result.status_code == 200
     assert len(templates_rendered[0][1]["grants"]) == 5
     soup = BeautifulSoup(result.data, "html.parser")
-    assert soup.h1.text == "My grants"
-    assert len(queries) == 3  # 1) select grant, 2) rollback
+    assert soup.h1.text == "Grants"
+    assert len(queries) == 3  # 1) select grants, 2) rollback
```
samuelhwilliams, line 33: Can we update the comment for the new query please 🙇

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -650,3 +643,152 @@ def test_edit_question_post(authenticated_platform_admin_client, factories, db_s
...
+    response = authenticated_platform_admin_client.post(
+        url_for("deliver_grant_funding.grant_setup_description"), data=desc_form.data, follow_redirects=False
+    )
+    assert response.status_code == 302
+    assert response.location == url_for("deliver_grant_funding.grant_setup_contact")
+
+
+def test_grant_setup_contact_get_with_session(authenticated_platform_admin_client):
+    with authenticated_platform_admin_client.session_transaction() as sess:
+        sess["grant_setup"] = {}
```
samuelhwilliams, line 756: Do we think it's worth updating these tests that set the session to `{}` to have representative data in them? I think useful to understand the flows as a developer reviewing the tests, but will leave to you on whether it's too much maintenance burden.

tests/unit/app/deliver_grant_funding/test_forms.py
```diff
@@ -0,0 +1,76 @@
...
+
+    with patch("app.deliver_grant_funding.forms.grant_name_exists", return_value=True):
+        with pytest.raises(ValidationError, match="Grant name already in use"):
+            validator(form, form.name)
+
+
+def test_grant_ggis_form_validates_when_no_selected():
+    app = Flask(__name__)
+    app.config["SECRET_KEY"] = "test-key"  # pragma: allowlist secret
+    app.config["WTF_CSRF_ENABLED"] = False  # Disable CSRF for testing
```
samuelhwilliams, line 43: Is this because you had CSRF issues with testing these forms otherwise?

samuelhwilliams (approved):
Looks good, a few more optionals/comments but I think nothing blocking if you want to get this in. Happy for you to fix any of these that feel important, and push back on others 🙇

tests/unit/app/deliver_grant_funding/test_forms.py
wjrm500: Correct yes, tests fail otherwise - the `FlaskForm` requires app request context, I did think about using integration tests for testing the GGIS form here, but decided to keep everything under unit tests

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/description.html
wjrm500: This was just part of the prototype HTML, see screenshot:

image

But yes defo doesn't make sense, can remove completely

tests/unit/app/deliver_grant_funding/test_forms.py
samuelhwilliams: I suppose my question is - why do we need to make and configure our own app instance here? Why can't we use our standard `app` fixture, which sets that up for us? 🤔

wjrm500: Yarp very true, tried this and can confirm it works - did originally look at `conftest.py` here for the `app` fixture but then forgot it existed

app/deliver_grant_funding/routes.py
wjrm500: Tried this and we get errors like `FAILED tests/integration/deliver_grant_funding/test_routes.py::test_grant_setup_description_post_valid - ValueError: "GrantSetupSession" object has no field "submit"`. Briefly hunted for an `exclude_fields` arg or some sort of thing for `populate_obj` but doesn't look like it exists, so I think we'll keep as-is

PR #215, opened 2025-05-29
FSPT-409 identify auto commit from test

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FSPT-409**

###In this PR we try to achieve following 

1. if there is a `POST` `PATCH` 'DELETE' endpoint which is missing the  `@auto_commit_after_request` and then this endpoint needs to update the db but since it is not committing it violates the data base integrity here we detect wether session is clean or not during the end of request (clean means is data saved to the db or not)
2. also this helps us to check miss use of 'GET' endpoint so in these we should not do any create, update, delete and if detects there is a uncommitted change that means a miss use 
3. This change is event driven & according to the sqlalchemy that is the best way to identify wether that is committed or not [Documentation](https://docs.sqlalchemy.org/en/20/orm/events.html#sqlalchemy.orm.SessionEvents.before_flush)

### How to test 

Remove the annotation in one of the POST endpoint & test and the annotation is `@auto_commit_after_request` 

### Screen shots 

After removing the annotation
image

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -18,7 +18,11 @@ from flask_sqlalchemy_lite import SQLAlchemy
 from flask_wtf import FlaskForm
 from jinja2 import Template
 from pytest_mock import MockerFixture
+from sqlalchemy import event
+from sqlalchemy.engine import Connection
 from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session as SASession
+from sqlalchemy.orm.session import SessionTransaction
 from sqlalchemy_utils import create_database, database_exists
 from testcontainers.postgres import PostgresContainer
 from werkzeug.test import TestResponse
@@ -110,6 +114,8 @@ def _validate_form_argument_to_render_template(response: TestResponse, templates
 
 @pytest.fixture()
 def anonymous_client(app: Flask, templates_rendered: TTemplatesRendered) -> FlaskClient:
+    _setup_session_clean_tracking()  # setting up listeners
+
     class CustomClient(FundingServiceTestClient):
         # We want to be sure that any data methods that act during the request have been
         # committed by the flask app lifecycle before continuing. Because of the way we configure
@@ -124,6 +130,17 @@ def anonymous_client(app: Flask, templates_rendered: TTemplatesRendered) -> Flas
 
             response = super().open(*args, **kwargs)
             _validate_form_argument_to_render_template(response, templates_rendered)
+
+            session = app.extensions["sqlalchemy"].session
+            if not session.info.get("clean", True):
+                method = kwargs.get("method") or (args[0] if args else "GET")
+                path = kwargs.get("path") or (args[0] if args else "/")
+
+                pytest.fail(
+                    f"Detected uncommitted changes in the SQLAlchemy session after handling "
+                    f"{method} request to {path}. "
+                    f"To ensure database consistency, wrap your request handler with @auto_commit_after_request."
+                )
 
             app.extensions["sqlalchemy"].session.rollback()
             return response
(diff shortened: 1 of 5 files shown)


nuwan-samarasinghe:
This PR failed due to a time constraint that we have for the tests it some some of the tests exceeds the given time limit & that test is not relates to this PR

tests/integration/conftest.py
```diff
@@ -18,7 +18,11 @@
 from flask_wtf import FlaskForm
 from jinja2 import Template
 from pytest_mock import MockerFixture
+from sqlalchemy import event
+from sqlalchemy.engine import Connection
 from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session as SASession
```
samuelhwilliams, line 24: Any reason why we reimporting this rather than just using the existing `Session` import above?

samuelhwilliams (commented):
Do you think we should exclude SAVEPOINTS from the `track_sql_queries` fixture? It would mean we don't need to think about them in the test body, and as they're a side-effect of how we set up the transactional test isolation, maybe that's a nice thing to not have to think/care about there.

This relies on us not using savepoints in real app code, but we don't do that and I don't think we plan to do that - so it might be ok for now.

tests/integration/conftest.py
```diff
@@ -290,6 +307,41 @@ def authenticated_platform_admin_client(
...
+    def after_begin(session: SASession, transaction: SessionTransaction, connection: Connection) -> None:
+        session.info["clean"] = True
+
+    # 2. Before flush — catch dirty state even if flush hasn't happened yet
+    @event.listens_for(SASession, "before_flush")
+    def before_flush(session: SASession, flush_context: Any, instances: object | None) -> None:
+        if session.new or session.dirty or session.deleted:
+            session.info["clean"] = False
+
+    # 3. After flush — useful if more changes happen in-process
```
samuelhwilliams, line 322: Can you give me some examples of the kind of changes that could happen between `before_flush` and `after_flush` that we're catching here?

samuelhwilliams:
This looks really nice @nuwan-samarasinghe - good work; annoying it was more complicated than I think we expected with the need to hook into events, but this isn't super complicated and does give us a bit more confidence. V nice :)

tests/integration/conftest.py
nuwan-samarasinghe: E.g. 

when you added some data into the ORM object & flushed that does not men that it is committed so here if some data is being added, updated, deleted then when you init the flush we know that it called the flush but did not call commit so data will not be written into the db 

I think after flush we don't need

tests/integration/conftest.py
nuwan-samarasinghe: 🙇  let me fix this

nuwan-samarasinghe: Fixed

tests/integration/conftest.py
```diff
@@ -125,6 +131,17 @@ def open(self, *args, **kwargs) -> TestResponse:  # type: ignore[no-untyped-def]
             response = super().open(*args, **kwargs)
             _validate_form_argument_to_render_template(response, templates_rendered)
 
+            session = app.extensions["sqlalchemy"].session
+            if not session.info.get("clean", True):
+                method = kwargs.get("method") or (args[0] if args else "GET")
+                path = kwargs.get("path") or (args[0] if args else "/")
```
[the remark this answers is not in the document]
nuwan-samarasinghe: Fixed

samuelhwilliams (approved):
Really nice 👍 

I'd like to see more thoughtful commit messages in the future - it would be great if the context in the PR description could be echoed appropriately in the commit messages as well, so that developers can use `git blame`, and IDE integrations to show why code has changed, more effectively :)

PR #216, opened 2025-05-29
FSPT-503 Persist submitted question data

sfount:
Add a collections interface method to update the `data` attribute and
store the latest submitted answer from the user.

Note that for now we're just copying the existing data to persuade the
ORM to issue an update on the model, in the future we might want to look
into the SQLAlchemy mutable extension.

This proposes using Pydantic to manage the question data types and give
them a consistent interface. All of the question types we have
right now are primitives so these work with the current method of
building dynamic forms with a single field without any changes.

### DIFF
--- a/app/common/collections/forms.py
+++ b/app/common/collections/forms.py
@@ -1,5 +1,3 @@
-from typing import cast
-
 from flask_wtf import FlaskForm
 from govuk_frontend_wtf.wtforms_widgets import GovSubmitInput, GovTextArea, GovTextInput
 from wtforms.fields.numeric import IntegerField
@@ -22,7 +20,7 @@ class DynamicQuestionForm(FlaskForm):
     submit: SubmitField
 
 
-def build_question_form(question: Question) -> DynamicQuestionForm:
+def build_question_form(question: Question) -> type[DynamicQuestionForm]:
     # NOTE: Keep the fields+types in sync with the class of the same name above.
     class _DynamicQuestionForm(FlaskForm):  # noqa
         question: _accepted_fields
@@ -49,4 +47,4 @@ def build_question_form(question: Question) -> DynamicQuestionForm:
 
     _DynamicQuestionForm.question = field
 
-    return cast(DynamicQuestionForm, _DynamicQuestionForm())
+    return _DynamicQuestionForm
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -64,9 +64,6 @@ def update_collection_schema(schema: CollectionSchema, *, name: str) -> Collecti
 
 
 def update_collection_data(collection: Collection, question: Question, data: BaseModel) -> Collection:
-    # copying the existing data will convince SQLAlchemy that something needs to be updated
-    # this could also be done using the mutability extension which may be more performant
-    # https://docs.sqlalchemy.org/en/20/orm/extensions/mutable.html#module-sqlalchemy.ext.mutable
     collection.data = copy.deepcopy(collection.data)
     collection.data[str(question.id)] = data.model_dump()
     db.session.flush()
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -1,6 +1,8 @@
+import copy
 from typing import Any
 from uuid import UUID
 
+from pydantic import BaseModel
 from sqlalchemy import select, text
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload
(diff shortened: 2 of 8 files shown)


sfount:
Reading it back, using the pydantic models while we only have primitive question types does feel like overkill. I'm happy to remove those and just set the validated result of the question form directly until we feel like we need something more managed.

app/common/data/interfaces/collections.py
```diff
@@ -61,6 +62,16 @@ def update_collection_schema(schema: CollectionSchema, *, name: str) -> Collecti
     return schema
 
 
+def update_collection_data(
+    collection: Collection, question: Question, data: str | int | float | bool | None
+) -> Collection:
+    collection.data = copy.deepcopy(collection.data)
```
samuelhwilliams, line 68: In my WIP branch on this, I did just pull in `sqlalchemy-json`. I think let's just do that now to avoid needing to do this.

tests/unit/app/common/collections/test_forms.py
```diff
@@ -31,7 +31,7 @@ def test_the_next_test_exhausts_QuestionDataType(self):
     def test_expected_field_types(self, app, data_type, expected_field_type, expected_widget):
         """Feels like a bit of a redundant test that's just reimplementing the function, but ... :shrug:"""
         q = Question(text="Question text", hint="Question hint", data_type=data_type)
-        form = build_question_form(q)
+        form = build_question_form(q)()
```
samuelhwilliams, line 34: Good catch 👍

app/common/collections/forms.py
```diff
@@ -22,7 +20,7 @@ class DynamicQuestionForm(FlaskForm):
     submit: SubmitField
 
 
-def build_question_form(question: Question) -> DynamicQuestionForm:
+def build_question_form(question: Question) -> type[DynamicQuestionForm]:
```
samuelhwilliams, line 23: Another good catch - sorry 👍

app/developers/routes.py
```diff
@@ -516,12 +516,16 @@ def collection_tasklist(collection_id: UUID) -> ResponseReturnValue:
...
 @developers_blueprint.route("/collections/<uuid:collection_id>/<uuid:question_id>", methods=["GET", "POST"])
 @platform_admin_role_required
+@auto_commit_after_request
 def ask_a_question(collection_id: UUID, question_id: UUID) -> ResponseReturnValue:
     collection_helper = CollectionHelper.load(collection_id)
     question = collection_helper.get_question(question_id)
-    form = build_question_form(question)
+    answer = collection_helper.get_answer_for_question(question.id)
+
+    form = build_question_form(question)(question=answer)
```
samuelhwilliams, line 525: This line looks grim `question=answer` - but it's fine 😅

samuelhwilliams (approved):
Looks good 👍 

I think we should go straight to using the pydantic approach you've investigated. Even if it feels a bit overkill now, it feels really clean and I think good to get in place now so that it's obvious how we want to move on.

tests/integration/common/data/interfaces/test_collections.py
```diff
@@ -311,3 +313,16 @@ def test_move_question_up_down(db_session, factories):
...
     assert q2.order == 0
     assert q3.order == 1
+
+
+def test_update_collection_data(db_session, factories):
+    form = factories.form.build()
+    question = factories.question.build(form=form)
+    collection = factories.collection.build(collection_schema=form.section.collection_schema)
+
+    assert collection.data.get(str(question.id)) is None
```
samuelhwilliams, line 323:
```suggestion
    assert str(question.id) not in collection.data
```

Unsure if intentional, but using `.get()` will return `None` if the key doesn't exist, so this is a bit of a weaker assertion than it probably needs to be?

Same for everywhere else we use `.get` - probably just do straight lookup with `[]`?

tests/integration/common/helpers/test_collections.py
```diff
@@ -243,3 +245,30 @@ def test_current_question_does_not_exist(self, db_session, factories):
                 r"Could not find form for question_id=00000000-0000-0000-0000-000000000009 "
                 r"in collection_schema=[a-z0-9-]+"
             )
+
+    class TestGetAndSubmitAnswerForQuestion:
+        def test_submit_valid_data(self, db_session, factories):
+            section = factories.section.build()
+            form = factories.form.build(section=section)
+            question = factories.question.build(form=form)
```
samuelhwilliams, line 253:
```suggestion
            question = factories.question.build()
```

The question will walk itself back up to form/section/schema for us - so can cut a bit of cruft here if you want.

samuelhwilliams (approved):
Looks v good, well done

tests/integration/common/helpers/test_collections.py
sfount: Thanks I did feel like I was overkilling the scenario but just ended up copying and pasting as it built - will make this change just so that doesn't propagate!

sfount: Thanks this was helpful have made the change https://github.com/communitiesuk/funding-service/pull/216/commits/2de422a82fe1bc2762a78ec5bb5288973be08311

samuelhwilliams (approved):

PR #220, opened 2025-05-30
FSPT-503 Calculate statuses based on collection

sfount:
Depends on https://github.com/communitiesuk/funding-service/pull/216

Replaces the stub methods for calculating form and collection statuses
with methods to iterate over questions/ forms and set not started/ in
progress/ completed statuses accordingly.

Note this commit intentionally out of scopes any template updates to set
an appropriate macro per status, we'll do that and pass in the status
enums in a separate commit.

Adds a status macro to map collection statuses to grey, blue and green
classes according to the current form/ collection status. This may later
need to be specific to different use cases but as its the only one for
now just call it status.

Video showing the form and collection statuses reflecting updated info:

https://github.com/user-attachments/assets/279b519b-0304-4835-bed5-3bfc9f8393cf


app/developers/routes.py
```diff
@@ -511,7 +511,9 @@ def edit_question(
 @platform_admin_role_required
 def collection_tasklist(collection_id: UUID) -> ResponseReturnValue:
     collection_helper = CollectionHelper.load(collection_id)
-    return render_template("developers/collection_tasklist.html", collection_helper=collection_helper)
+    return render_template(
+        "developers/collection_tasklist.html", collection_helper=collection_helper, statuses=CollectionStatusEnum
```
samuelhwilliams, line 515: possibly keep an eye on whether we inject CollectionStatusEnum - and some other enums - automatically in the future. Fine for now.

PR #221, opened 2025-05-30
Stop non-admin MHCLG users from logging in with SSO

samuelhwilliams:
## Description
We currently allow all MHCLG users to log in via SSO, but if they're not in our admin group then they won't have permission to do anything. Rather than designing around the edge cases in design this creates, we will just block them from logging in and direct them to support if needed.

## Show it
image

@Kateharries - please feel free to adjust content as you wish when you're back 🙇


nuwan-samarasinghe (approved):
🥇

PR #225, opened 2025-05-30
Disable slow test failures in CI

samuelhwilliams:
We're seeing this pop off a lot in github actions at the moment. This feels maybe suboptimal - but maybe better than turning it off altogether, or loosening it up temporarily and then forgetting to tighten it back up.


sfount (approved):
I wonder if put in a story to turn this into a Slack message? Just enough to keep us aware of it and keeping it 🤏 but not enough to block progress with outside factors

PR #226, opened 2025-05-30
FSPT-520 List collections

sfount:
Adds a page to list collections for a given schema, this is hooked up to
the manage a schema (building a tasklist) page.

For now it just uses the collections that are fetched and back
referenced from the collection schema - I think the ORM should only be
fetching the collection schema once for any number of collections which
is nice but should be validated (otherwise thats a very expensive
query!). We might need to revisit that if we want to limit or paginate
the number of collections we're returning.

Link from schema management to collections preview
Screenshot 2025-05-30 at 16 58 29

Empty collections page
Screenshot 2025-05-30 at 16 58 32

A preview collection exists
Screenshot 2025-05-30 at 16 58 48

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -4,7 +4,7 @@ from uuid import UUID
 from pydantic import BaseModel
 from sqlalchemy import select, text
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import joinedload
+from sqlalchemy.orm import joinedload, selectinload
 
 from app.common.data.interfaces.exceptions import DuplicateValueError
 from app.common.data.models import (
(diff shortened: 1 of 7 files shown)


app/common/data/interfaces/collections.py
```diff
@@ -33,22 +33,28 @@ def create_collection_schema(*, name: str, user: User, grant: Grant, version: in
...
+) -> CollectionSchema:
     """Get a collection schema by ID and optionally version.
 
     If you do not pass a version, it will retrieve the latest version (ie highest version number).
 
     Note: We may wish to change this behaviour to the latest 'published' version in the future, or some other logic.
     """
+    options = []
+    if with_full_schema:
+        options.append(selectinload(CollectionSchema.sections).selectinload(Section.forms).selectinload(Form.questions))
```
sfount, line 47: Going to add an integration test to make sure these options are appropriately applied and this is eagerly loaded as expected

PR #233, opened 2025-06-02
Move factoryboy fixtures out of the `integration` sub-directory

samuelhwilliams:
Unit tests may still find it useful to be able to generate instances of our models to help test certain functionality. We don't want to do this manually when we have configured factories to help us instantiate test data.

## Caveat
When using built-but-unflushed DB instances, we need to be careful not to rely on behaviour that SQLAlchemy provides for us. One example of this can be ordering on relationships. In places where we want to explicitly rely on that ordering, we either need to:

1) Write the tests as integration tests that flush/commit to the DB, so SQLAlchemy can sync the correct ordering
2) Rely on ordering in python code instead.


nuwan-samarasinghe:
So if we adding these into unit tests what will be the purpose of the unit test it seems if we add to the unit tests means we are more and more becoming integration test ? cant we use mocking instead of giving actual db behaviour to unit tests

samuelhwilliams:
The unit tests will not be able to talk to the DB with this set up - that is enforced through the top-level fixture that raises an error on access to `db.session`.

But some unit tests will still need to receive a SQLAlchemy instance/object, and I think that's a legitimate thing to unit test. For example, the `CollectionHelper` should _essentially_ just be receiving raw data - it's just that the data is wrapped in a SQLAlchemy object. So we can still unit test methods on that, but we need to give them a populated object.

tests/conftest.py
```diff
@@ -110,3 +125,57 @@ def _precompile_templates(app: Flask) -> None:
     # takes away a significant amount of the difference between the first and second pass.
     for template_name in app.jinja_env.list_templates():
         app.jinja_env.get_template(template_name)
+
+
+@pytest.fixture(scope="function", autouse=True)
+def db_session(app: Flask) -> Generator[None, None, None]:
```
MarcUsher, line 131: This is an interesting lil change! One I'll come back to for peeping

MarcUsher (approved):
Noice, this seems like a sensible change (and helpful to show how it works in practice with test_collections moving to unit tests)

PR #237, opened 2025-06-02
FSPT-521 Show grant list page based on the user roll platform admin or member

nuwan-samarasinghe:
**Ticket : https://mhclgdigital.atlassian.net/browse/FSPT-521**

In the following PR based on the user roll it will show the grants and following are the scenarios that can occur

1. If you have a one record of `ADMIN` in the roles table then we treat them as a `PLATFORM_ADMIN` and they can see all the grants available
2. If you have a one record with `MEMBER` then you should directly go to the grants detail page
3. If you have multiple grants with `MEMBER` then you will go to the grants list table page.

How to test 2 & 3 if you used SSO login remove the `ADMIN` record manually and add some grants & roles as `MEMBER` then you can test 2 and 3 

Grants list page for a `ADMIN` or a `MEMBER` with multiple grants

image

### DIFF
--- a/app/common/data/interfaces/grants.py
+++ b/app/common/data/interfaces/grants.py
@@ -1,11 +1,13 @@
-from typing import Sequence
+from typing import Sequence, cast
 from uuid import UUID
 
+from flask_login import current_user
 from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError
 
 from app.common.data.interfaces.exceptions import DuplicateValueError
 from app.common.data.models import Grant
+from app.common.data.models_user import User
 from app.extensions import db
 
 
@@ -17,6 +19,19 @@ def grant_name_exists(name: str) -> bool:
     statement = select(Grant).where(Grant.name == name)
     grant = db.session.scalar(statement)
     return grant is not None
+
+
+def get_all_grants_by_user() -> Sequence[Grant]:
+    user: User = cast(User, current_user)
+    if user.is_platform_admin:
+        statement = select(Grant).order_by(Grant.name)
+        return db.session.scalars(statement).all()
+    else:
+        grant_ids = [role.grant_id for role in user.roles]
+        if not grant_ids:
+            return []
+        statement = select(Grant).where(Grant.id.in_(grant_ids)).order_by(Grant.name)
+        return db.session.scalars(statement).all()
 
 
 def get_all_grants() -> Sequence[Grant]:
--- a/app/common/data/interfaces/grants.py
+++ b/app/common/data/interfaces/grants.py
@@ -1,7 +1,6 @@
-from typing import Sequence, Tuple, cast
+from typing import Sequence
 from uuid import UUID
 
-from flask_login import current_user
 from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError
 
@@ -21,17 +20,16 @@ def grant_name_exists(name: str) -> bool:
     return grant is not None
 
 
-def get_all_grants_by_user() -> Tuple[Sequence[Grant], bool]:
-    user = cast(User, current_user)
+def get_all_grants_by_user(user: User) -> Sequence[Grant]:
     if user.is_platform_admin:
         statement = select(Grant).order_by(Grant.name)
-        return db.session.scalars(statement).all(), user.is_platform_admin
+        return db.session.scalars(statement).all()
     else:
         grant_ids = [role.grant_id for role in user.roles]
         if not grant_ids:
-            return [], user.is_platform_admin
+            return []
         statement = select(Grant).where(Grant.id.in_(grant_ids)).order_by(Grant.name)
-        return db.session.scalars(statement).all(), user.is_platform_admin
+        return db.session.scalars(statement).all()
 
 
 def get_all_grants() -> Sequence[Grant]:
--- a/app/common/data/interfaces/grants.py
+++ b/app/common/data/interfaces/grants.py
@@ -22,7 +22,7 @@ def grant_name_exists(name: str) -> bool:
 
 
 def get_all_grants_by_user() -> Sequence[Grant]:
-    user: User = cast(User, current_user)
+    user = cast(User, current_user)
     if user.is_platform_admin:
         statement = select(Grant).order_by(Grant.name)
         return db.session.scalars(statement).all()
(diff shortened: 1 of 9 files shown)


app/deliver_grant_funding/routes.py
```diff
@@ -142,8 +145,13 @@ def grant_setup_contact() -> ResponseReturnValue:
 
 @deliver_grant_funding_blueprint.route("/grants", methods=["GET"])
 @mhclg_login_required
-def list_grants() -> str:
-    grants = interfaces.grants.get_all_grants()
+def list_grants() -> Response | str:
+    user: User = cast(User, current_user)
+    grants = interfaces.grants.get_all_grants_by_user(user)
+    if not grants:
+        return redirect(url_for("deliver_grant_funding.grant_setup_intro"))
```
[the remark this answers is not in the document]
nuwan-samarasinghe: oh I think I did a bad thing over here I don't know how I added adding this change over here I updated the ticket [FSPT-521](https://mhclgdigital.atlassian.net/browse/FSPT-521) with the details but it seems I did not mention this & added let me have a chat with you about this

nuwan-samarasinghe:
One issue is being noticed so please let me fix that issue is once you logged in and created a grant and your entire system has only one grant and now even platform admin cannot create a another grant

app/common/data/interfaces/grants.py
```diff
@@ -19,6 +21,19 @@ def grant_name_exists(name: str) -> bool:
...
+    user = cast(User, current_user)
+    if user.is_platform_admin:
+        statement = select(Grant).order_by(Grant.name)
+        return db.session.scalars(statement).all(), user.is_platform_admin
+    else:
+        grant_ids = [role.grant_id for role in user.roles]
+        if not grant_ids:
+            return [], user.is_platform_admin
+        statement = select(Grant).where(Grant.id.in_(grant_ids)).order_by(Grant.name)
+        return db.session.scalars(statement).all(), user.is_platform_admin
```
samuelhwilliams, line 34: Sorry, I don't like this interface, I don't think we need to be returning a fact about the user as part of this - just the list of grants itself. Adding a boolean about the user feels like it makes things less clear.

Can we keep the return type/value the same please? In the view we can just poke at current_user.is_platform_admin directly if we want some specific behaviour related to that there 🙏

nuwan-samarasinghe: Fixed

nuwan-samarasinghe: If we directly use the `current_user.is_platform_admin` it gives a type not found in the pre-commit and either we have to individually check is the attribute available or not I don't feel that is correct and also casting in each place is also not good so that why I made `interfaces.user.get_current_user()` which convert the current_user into User

app/common/data/interfaces/grants.py
```diff
@@ -19,6 +20,18 @@ def grant_name_exists(name: str) -> bool:
...
+def get_all_grants_by_user(user: User) -> Sequence[Grant]:
+    if user.is_platform_admin:
+        statement = select(Grant).order_by(Grant.name)
+        return db.session.scalars(statement).all()
+    else:
+        grant_ids = [role.grant_id for role in user.roles]
+        if not grant_ids:
+            return []
+        statement = select(Grant).where(Grant.id.in_(grant_ids)).order_by(Grant.name)
+        return db.session.scalars(statement).all()
```
samuelhwilliams, line 32: Please can you add some test coverage for this - anything in interfaces is essentially the core of our appliation, so definitely want to have high confidence in it all 🙏

app/common/data/interfaces/user.py
```diff
@@ -13,6 +15,11 @@ def get_user(id_: str | uuid.UUID) -> User | None:
     return db.session.get(User, id_)
 
 
+def get_current_user() -> User:
+    user = cast(User, current_user)
+    return user
```
samuelhwilliams, line 20: I got stuck on solving this problem for ages by trying to force Flask's actual `current_user` to have the User type, and never made any progress, which has led to us having to cast current_user to User in a load of places in the codebase.

Just doing this feels like a much more sensible and pragmatic solution to that problem - good idea 🙇 👍

It might be nice to do this as a separate PR and update everywhere in the codebase appropriately? It feels very much an isolated change/refactor that we could do by itself.

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -22,17 +25,64 @@
...
+    assert len(queries) == 2
+
+
+def test_list_grants_as_member_with_multiple_grants(
+    app, authenticated_member_client, factories, templates_rendered, track_sql_queries
+):
+    grants = factories.grant.create_batch(5)
+    user: User = cast(User, current_user)
+    for grant in grants:
+        factories.user_role.create(user_id=user.id, user=user, role=RoleEnum.MEMBER, grant=grant)
```
samuelhwilliams, line 71: just fyi for the future, you only need to pass the `user` as `user_id` is derived from that

```suggestion
        factories.user_role.create(user=user, role=RoleEnum.MEMBER, grant=grant)
```

app/deliver_grant_funding/templates/deliver_grant_funding/grant_list.html
```diff
@@ -1,27 +1,49 @@
...
+    {% if grants %}
+    {% set table_rows = namespace(items=[]) %}
+    {% for grant in grants %}
+    {% set grant_detail_link %}
+    <a class='govuk-link govuk-link--no-visited-state'
+       href={{ url_for("deliver_grant_funding.view_grant", grant_id=grant.id) }}>{{ grant.name }}</a>
+    {% endset %}
+    {% set table_rows.items = table_rows.items + [
+                [{"html": grant_detail_link},
+                 {"text": grant.ggis_number},
```
samuelhwilliams, line 31: image

Can we show an empty field instead of the string `None` here please? I think we're making GGIS number compulsory soon, but for the moment it still looks a bit weird.

```suggestion
                 {"text": grant.ggis_number or ''},
```

nuwan-samarasinghe: Fixed

app/common/data/interfaces/grants.py
nuwan-samarasinghe: Added some tests

samuelhwilliams (approved):
Nice 👍

PR #238, opened 2025-06-02
FSPT-520 Getting a collection schema loads everything from the database

sfount:
Ensures that the entire collection schema is loaded when a schema if fetched with the `with_full_schema` flag set to `True`.

We'll want to avoid n+1 queries and the ORM fetching properties that we'd expect to constantly be looping through in the code.


srh-sloan (commented):
Nice test, but should we have one where we explicitly set `with_full_schema=False` and count the queries then as well?

tests/integration/common/data/interfaces/test_collections.py
```diff
@@ -325,3 +325,33 @@ def test_update_collection_data(db_session, factories):
...
+    with track_sql_queries() as queries:
+        from_db = get_collection_schema(schema_id=schema.id, with_full_schema=True)
+    assert from_db is not None
+
+    # Expected queries:
+    # * Initial queries for schema and user
+    # * Load the sections
+    # * Load the forms
+    # * Load the question
+    assert len(queries) == 5
```
samuelhwilliams, line 347: Have I forgotten how to count 👀

tests/integration/common/data/interfaces/test_collections.py
```diff
@@ -325,3 +325,33 @@ def test_update_collection_data(db_session, factories):
...
+
+    # No additional queries when inspecting the ORM model
+    count = 0
+    with track_sql_queries() as queries:
+        for s in from_db.sections:
+            for f in s.forms:
+                for _q in f.questions:
+                    count += 1
+
+    assert queries == []
```
nuwan-samarasinghe, line 357: Don't we need to verify the counts are correct ?

sfount:
> Nice test, but should we have one where we explicitly set `with_full_schema=False` and count the queries then as well?

Yeah this feels like maybe something we could do once to catch false positives? The main goal of the test is just to make sure our configuration of the ORM doesn't make extra/ unexpected queries after loading which this gives us reassurance on but adding a small assertion to make sure that the opposite is true might be good once

tests/integration/common/data/interfaces/test_collections.py
sfount: This is just checking that there are no queries issued to the database even when you're stepping through all of the properties of the model - if it hadn't fetched all of the sections/ forms/ questions up front this query list would be full of extra (SELECT question FROM...) queries as you tried to load properties that it hadn't fetched yet

samuelhwilliams (approved):

samuelhwilliams (approved):

PR #240, opened 2025-06-02
FSPT-524 Form statuses model "Have you completed this section?"

sfount:
We've called this metadata here as its how we referred to it while
breaking down the story (the users submitted answers as the data of a
collection and what has happened to it as metadata that derives its
status). Another option here is "Collection events" or the things that
have happened to the collection.

SQLAlchemy does reserve the property "metadata" so a prefix was needed
for this naming scheme, happy to revise.

To try out this method we're just using a very simple method of removing events
that are no longer relevant (if a user decides its no longer complete). 

If we want to meet the theoretical ideal of having non-mutable
events that tell are replayable and describe how a status was reached we can
move this to just add new events undoing the "Form marked as complete" and pick
the latest status - a number of options for this described in the tickets.

Note: for a short period of time between this PR and the next one there
will be no "Complete" collections as the page which can actually
register this event hasn't been merged!


app/common/data/types.py
```diff
@@ -36,3 +36,7 @@ class CollectionStatusEnum(enum.StrEnum):
     NOT_STARTED = "Not started"
     IN_PROGRESS = "In progress"
     COMPLETED = "Completed"
+
+
+class MetadataEventKey(enum.StrEnum):
+    FORM_RUNNER_FORM_COMPLETED = "Form completed"
```
samuelhwilliams, line 42:
```suggestion
    FORM_RUNNER_FORM_MARKED_COMPLETE = "Form marked as complete"
    FORM_RUNNER_FORM_MARKED_INCOMPLETE = "Form marked as incomplete"
```

maybe?

app/common/helpers/collections.py
```diff
@@ -165,6 +169,11 @@ def submit_answer_for_question(self, question_id: UUID, form: DynamicQuestionFor
         data = _form_data_to_question_type(question, form)
         interfaces.collections.update_collection_data(self.collection, question, data)
 
+    def mark_form_as_complete(self, form: "Form", user: "User") -> None:
+        interfaces.collections.add_collection_metadata(
+            self.collection, MetadataEventKey.FORM_RUNNER_FORM_COMPLETED, user, form
+        )
```
samuelhwilliams, line 175: Should we also check (and below, test) that data is actually present for all relevant questions?

sfount: In the next PR I've added a helper method that calls the interfaces that does an integrity check to make sure all of the questions have been answers and raises a value error otherwise (with coverage!). I'll push that up and have to chat through how its feeling at the helper vs. interface level

app/common/data/types.py
samuelhwilliams: > To try out this method we're just using a very simple method of removing events
that are no longer relevant (if a user decides a form is no longer complete).

Ah I see your comment.

I don't love the idea of deleting audit-trail-esque events very much 👀

sfount:
Going to merge this for now and we can have a look at how its feeling after the final ticket or two for the forms baseline -- very happy for this model to be revised (or altered, reducing immutable entries does feel cleaner)

PR #244, opened 2025-06-03
FSPT-510 - Grant setup flow 'Check your answers' page

wjrm500:
### Ticket

["Add a grant" review / check your answers](https://mhclgdigital.atlassian.net/browse/FSPT-510)

### Description

Added a "Check your answers" page after the grant setup flow. This lists the data points submitted by the user during grant setup, with a "Change" link alongside each data point. Clicking "Change" takes you back to the relevant form page, and then submitting the form takes you straight back to "Check your answers". Also added a setup confirmation page after "Check your answers".

### Potentially contentious decisions / questions

- "None" shown on "Check your answers" summary if no GGIS number given - should this just be empty string instead?
- Duplicate value error on database grant creation not handled - wasn't sure what to do with this e.g., would we show the error on the "Check your answers" page (but not tied to a specific field) or redirect the user to the specific page the error occurred for (this would always be the form page for grant name at the moment as name is the only unique field, but GGIS number may be unique in future). We do validate name uniqueness when the grant name form is submitted
- Page title for "Check your answers" is literally just "Check your answers"
- On setup confirmation page, the "Continue to grant home" is a link styled as a button (with `govuk-button` CSS class) - felt like a FlaskForm here was overkill?

### Screenshots

image

image

### DIFF
--- a/app/deliver_grant_funding/routes.py
+++ b/app/deliver_grant_funding/routes.py
@@ -1,6 +1,6 @@
 from uuid import UUID
 
-from flask import Blueprint, redirect, render_template, session, url_for
+from flask import Blueprint, redirect, render_template, request, session, url_for
 from flask.typing import ResponseReturnValue
 from werkzeug import Response
 from wtforms.fields.core import Field
@@ -9,6 +9,7 @@ from app.common.auth.decorators import mhclg_login_required, platform_admin_role
 from app.common.data import interfaces
 from app.common.data.interfaces.exceptions import DuplicateValueError
 from app.deliver_grant_funding.forms import (
+    GrantCheckYourAnswersForm,
     GrantContactForm,
     GrantDescriptionForm,
     GrantForm,
@@ -20,6 +21,8 @@ from app.deliver_grant_funding.session_models import GrantSetupSession
 from app.extensions import auto_commit_after_request
 
 deliver_grant_funding_blueprint = Blueprint(name="deliver_grant_funding", import_name=__name__)
+
+CHECK_YOUR_ANSWERS = "check-your-answers"
 
 
 @deliver_grant_funding_blueprint.route("/grants/setup", methods=["GET", "POST"])
@@ -46,9 +49,15 @@ def grant_setup_ggis() -> ResponseReturnValue:
         grant_session.has_ggis = form.has_ggis.data
         grant_session.ggis_number = form.ggis_number.data if form.has_ggis.data == "yes" else None
         session["grant_setup"] = grant_session.to_session_dict()
+        if request.args.get("source") == CHECK_YOUR_ANSWERS:
+            return redirect(url_for("deliver_grant_funding.grant_setup_check_your_answers"))
         return redirect(url_for("deliver_grant_funding.grant_setup_name"))
 
-    return render_template("deliver_grant_funding/grant_setup/ggis_number.html", form=form)
+    return render_template(
+        "deliver_grant_funding/grant_setup/ggis_number.html",
+        form=form,
+        check_your_answers_source=CHECK_YOUR_ANSWERS,
+    )
 
 
 @deliver_grant_funding_blueprint.route("/grants/setup/name", methods=["GET", "POST"])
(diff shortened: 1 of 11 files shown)


app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/confirmation.html
```diff
@@ -0,0 +1,26 @@
...
+    </span>
+    <h1 class="govuk-heading-xl">{{ grant.name }} added</h1>
+
+    <p class="govuk-body">You can now:</p>
+    <ul class="govuk-list govuk-list--bullet">
+      <li>update or add more information about the grant</li>
+      <li>create forms</li>
+    </ul>
+
+    <a href="{{ url_for('deliver_grant_funding.view_grant', grant_id=grant.id) }}" class="govuk-button">Continue to grant home</a>
```
sfount, line 24: All good for this to just be a link styled like a button in my opinion 👍 I think it would be good to check-in on if we need this page or we can use the flash message as-in the "adding/ inviting a user" flow thats happening at the same time

sfount (commented):
Just had a quick look over and looks great! A few quick comments while @samuelhwilliams was afk but will defer to him for context on the last few PRs

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/check_your_answers.html
```diff
@@ -0,0 +1,74 @@
...
+    <span class="govuk-caption-xl">
+      <span class="govuk-visually-hidden">This section is:</span>
+      Set up the grant
+    </span>
+    <h1 class="govuk-heading-xl">Check your answers</h1>
+    <dl class="govuk-summary-list">
+      <!-- Grant GGIS -->
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__key">GGIS</dt>
+        <dd class="govuk-summary-list__value">{{ grant_session.ggis_number }}</dd>
```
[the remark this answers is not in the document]
samuelhwilliams: Agree, no nones please 🙇

app/deliver_grant_funding/templates/deliver_grant_funding/grant_setup/check_your_answers.html
```diff
@@ -0,0 +1,74 @@
...
+{% set active_item_identifier = "grants" %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <span class="govuk-caption-xl">
+      <span class="govuk-visually-hidden">This section is:</span>
+      Set up the grant
+    </span>
+    <h1 class="govuk-heading-xl">Check your answers</h1>
+    <dl class="govuk-summary-list">
```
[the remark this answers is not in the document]
wjrm500: Absolutely no reason whatsoever - just forgot / did not know that there was a macro for this! 😸

Will look to change now and let you know if there's any blockers, but I guess there won't be given the standard look of the page

app/deliver_grant_funding/routes.py
```diff
@@ -46,6 +47,8 @@ def grant_setup_ggis() -> ResponseReturnValue:
         grant_session.has_ggis = form.has_ggis.data
         grant_session.ggis_number = form.ggis_number.data if form.has_ggis.data == "yes" else None
         session["grant_setup"] = grant_session.to_session_dict()
+        if request.args.get("return_to_check"):
```
samuelhwilliams, line 50: I added some behaviour like this in form runner recently, but used the query arg `source`. Could we use that here as well please?

I think we'll soon enough need some service-wide way of handing/directing backlinks, which is bubbling in the back of my head slightly.

I don't think we need to do a service framework here (but we do have [a ticket for it](https://mhclgdigital.atlassian.net/browse/FSPT-417)), so I think let's just be manually consistent for now, I think?

I don't think you need the enum for values, maybe just define a CHECK_YOUR_ANSWERS='check-your-answers' constant or something. Or you could just use the presence of the `source` kwarg as an empty string, like you do here, but that feels a pinch less readable in this case.

app/deliver_grant_funding/routes.py
```diff
@@ -98,30 +104,47 @@ def grant_setup_contact() -> ResponseReturnValue:
...
+    if form.validate_on_submit():
+        grant = interfaces.grants.create_grant(
+            name=grant_session.name,
+            description=grant_session.description,
+            primary_contact_name=grant_session.primary_contact_name,
+            primary_contact_email=grant_session.primary_contact_email,
+            ggis_number=grant_session.ggis_number,
+        )
+        session.pop("grant_setup", None)
+        return redirect(url_for("deliver_grant_funding.grant_setup_confirmation", grant_id=grant.id))
```
samuelhwilliams, line 136: As you've flagged, we don't do any error handling here on things like the grant name unique constraint. We do have that on the grant name form page though.

I am broadly OK with this; it will be a low traffic page and the chance of meaningful errors that occur here but not on the original grant name page should be low.

If we start getting errors from this page in sentry, we can rethink.

samuelhwilliams (approved):
This looks and functions great - I'm happy, no small feat apparently (😭), well done 👏 

I have noticed one small separate bug that you could please take away to fix: if you're a non-platform-admin user and a member of multiple grants, when you visit the "list grants" page, you see the "Set up a grant" button. Clicking it takes you to a 403 page. We should hide the button on that page if you can't create grants.

PR #252, opened 2025-06-04
BAU: Change login required redirect to SSO

MarcUsher:
Unauthenticated users were previously redirected to the magic link sign in page and flow. Now that we've identified that our initial Deliver Grant Funding users will all be logging in with SSO, and now none of our routes or actions can be taken by non-MHCLG users logging in via magic link, it makes sense to switch which sign-in flow we push unauthenticated users to when they first try to login or hit a route that reqiures them to be authenticated and logged in.


samuelhwilliams (approved):

tests/e2e/pages.py
```diff
@@ -88,8 +88,8 @@ def navigate(self) -> None:
         self.page.goto(f"{self.domain}/grants")
         expect(self.title).to_be_visible()
 
-    def click_add_a_grant(self) -> GrantSetupIntroPage:
-        self.page.get_by_role("button", name="Add a grant").click()
+    def click_set_up_a_grant(self) -> GrantSetupIntroPage:
+        self.page.get_by_role("button", name="Set up a grant").click()
```
wjrm500, line 92: Huh, our E2E tests were failing then? Apologies 😸

wjrm500 (approved):
Nice wee change

tests/e2e/pages.py
MarcUsher: Yeah there were a couple of steps missing and a misnamed button, not much to update but thought it helpful to keep it up to date on spotting - annoying that these can't run in CI yet to catch things like this!

PR #253, opened 2025-06-04
FSPT-438 Submit collection backend

sfount:
Extends the collection status to only move to complete if a collection
has been submitted.

This will be plugged into a HTTP form that will submit collections in
a future commit.

(Note no preview collections will move to collect until the next story is in)

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -57,13 +57,15 @@ class CollectionHelper:
 
     @property
     def status(self) -> str:
+        submitted = MetadataEventKey.COLLECTION_SUBMITTED in [x.event_key for x in self.collection.collection_metadata]
+
         form_statuses = set(
             [
                 self.get_status_for_form(form)
                 for form in chain.from_iterable(section.forms for section in self.schema.sections)
             ]
         )
-        if {CollectionStatusEnum.COMPLETED} == form_statuses:
+        if {CollectionStatusEnum.COMPLETED} == form_statuses and submitted:
             return CollectionStatusEnum.COMPLETED
         elif {CollectionStatusEnum.NOT_STARTED} == form_statuses:
             return CollectionStatusEnum.NOT_STARTED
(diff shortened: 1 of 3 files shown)


app/common/helpers/collections.py
```diff
@@ -174,6 +176,23 @@ def submit_answer_for_question(self, question_id: UUID, form: DynamicQuestionFor
         data = _form_data_to_question_type(question, form)
         interfaces.collections.update_collection_data(self.collection, question, data)
 
+    def submit_collection(self, user: "User") -> None:
+        collection_complete = self.status == CollectionStatusEnum.COMPLETED
+        if collection_complete:
+            return
+
+        # TODO: is there a nice way to remove this duplication with the status check?
```
sfount, line 184: Note in the next PR I actually found two uses for this so separated it out into a method there https://github.com/communitiesuk/funding-service/pull/255/files#diff-6f42f8e6a224d41ee10b28d9a4433f0349f53be3adf572a9a3f7faeb01185241R129-R136

app/common/helpers/collections.py
```diff
@@ -57,13 +57,15 @@ def name(self) -> str:
 
     @property
     def status(self) -> str:
+        submitted = MetadataEventKey.COLLECTION_SUBMITTED in [x.event_key for x in self.collection.collection_metadata]
```
samuelhwilliams, line 60: I maybe have a _tiny_ feel that some of this could live on the model itself as a `submitted` property, then we can just proxy `self.collection.submitted` or something - but this is fine for now. I think some of those core/simple/status-y values feel useful possibly outside of the helper. Also fine to wait until that is the case.

app/common/helpers/collections.py
```diff
@@ -127,7 +129,7 @@ def get_all_questions_answered_for_form(
     def get_status_for_form(self, form: "Form") -> str:
         all_questions_answered, answers = self.get_all_questions_answered_for_form(form)
         marked_as_complete = MetadataEventKey.FORM_RUNNER_FORM_COMPLETED in [
-            x.event_key for x in self.collection.collection_metadata
+            x.event_key for x in self.collection.collection_metadata if x.form and x.form.id == form.id
```
samuelhwilliams, line 132: opt nit

```suggestion
            x.event_key for x in self.collection.collection_metadata if x.form_id == form.id
```

app/common/helpers/collections.py
```diff
@@ -57,13 +57,15 @@ def name(self) -> str:
...
+        submitted = MetadataEventKey.COLLECTION_SUBMITTED in [x.event_key for x in self.collection.collection_metadata]
+
         form_statuses = set(
             [
                 self.get_status_for_form(form)
                 for form in chain.from_iterable(section.forms for section in self.schema.sections)
             ]
         )
-        if {CollectionStatusEnum.COMPLETED} == form_statuses:
+        if {CollectionStatusEnum.COMPLETED} == form_statuses and submitted:
```
samuelhwilliams, line 68: Small edge case technically slipping through with submitted=True and form_statuses somehow not matching? Do we want to error something in that case?

app/common/helpers/collections.py
sfount: This is a good idea, I'd like all the "reducers" for these to live in one place they could be unit tested but didn't think quite far ahead enough for how to structure that

samuelhwilliams (approved):
Generally happy - optional comments - lovely jubbly.

PR #255, opened 2025-06-04
FSPT-438 Tasklist frontend for submitting collection

sfount:
Depends on https://github.com/communitiesuk/funding-service/pull/253

Adds a submit collection button to move it through to "Complete".

Tweaks the helper method for checking a collection form status so that
it can be used by the template.

The confirmation page and view only data will be added in a separate
commit.

Page showing the tasklist before all forms have been completed (submit not clickable and some text shown):
Screenshot 2025-06-04 at 15 50 21

Page showing the tasklist when all forms have been completed (submit is now clickable):
Screenshot 2025-06-04 at 15 48 39

Page showing the tasklist when all completed (note these commits don't add the confirmation page or the a view only mode for the forms)
Screenshot 2025-06-04 at 15 49 23

Edge case: if there's a race condition with another user changing questions or marking forms as not yet ready before you press submit the submission will not be allowed and it will be handled
Screenshot 2025-06-04 at 15 49 13

### DIFF
--- a/app/common/data/migrations/versions/006_metadata_enum.py
+++ b/app/common/data/migrations/versions/006_metadata_enum.py
@@ -1,0 +1,23 @@
+"""extends metadata enum
+
+Revision ID: 006_metadata_enum
+Revises: 005_collection_metadata
+Create Date: 2025-06-04 14:56:07.802724
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "006_metadata_enum"
+down_revision = "005_collection_metadata"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(sa.text("ALTER TYPE metadata_event_key_enum ADD VALUE 'COLLECTION_SUBMITTED'"))
+
+
+def downgrade() -> None:
+    pass
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -126,6 +126,15 @@ class CollectionHelper:
         answers = [answer for q in visible_questions if (answer := self.get_answer_for_question(q.id)) is not None]
         return len(visible_questions) == len(answers), answers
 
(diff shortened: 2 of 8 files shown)


app/common/data/migrations/versions/006_metadata_enum.py
```diff
@@ -0,0 +1,23 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(sa.text("ALTER TYPE metadata_event_key_enum ADD VALUE 'COLLECTION_SUBMITTED'"))
+
+
+def downgrade() -> None:
+    pass
```
samuelhwilliams, line 23: Hopefully we never need to/try to downgrade this 🤞

app/common/helpers/collections.py
```diff
@@ -126,6 +126,15 @@ def get_all_questions_answered_for_form(
         answers = [answer for q in visible_questions if (answer := self.get_answer_for_question(q.id)) is not None]
         return len(visible_questions) == len(answers), answers
 
+    def get_all_forms_completed(self) -> bool:
```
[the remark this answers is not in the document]
samuelhwilliams: Also 🤏  test for this please

samuelhwilliams (commented):
Looking nice 👍 Would like a test but otherwise all seems good

app/common/data/migrations/versions/006_metadata_enum.py
sfount: Yeah what I've done here lines up with what Sarah/ Marc looked into here https://mhclgdigital.atlassian.net/wiki/spaces/FS/pages/554139677/Enums#Adding-values

I figured if we're going to squash migrations before prod copying and altering and destroying was probably more effort/ error prone than its worth

samuelhwilliams (approved):

PR #257, opened 2025-06-04
FSPT-534 - Adding new grant fields to settings page to allow them to be updated after grant creation

wjrm500:
### Ticket

[View and edit the 3-4 current "grant profile" data points](https://mhclgdigital.atlassian.net/browse/FSPT-534)

### Description

This PR adds the new fields added as part of the new grant setup flow (GGIS number, description, primary contact details) to the "Details" page of the grant dashboard so that they can be changed after a grant is created. Currently only grant name can be changed via this page.

New integration tests have been added for new routes, and E2E tests updated to change the newly changeable fields.

### Decisions to flag

We mimic the "Check your answers" page for consistency, so do not provide separate table rows / change links for primary contact name and primary contact email - instead they are grouped under "Main contact".

### Show the thing

https://github.com/user-attachments/assets/11fc1bcb-1967-4f3b-89da-148e795c186f


app/deliver_grant_funding/forms.py
```diff
@@ -32,16 +32,6 @@ def __call__(self, form: FlaskForm, field: StringField) -> None:
             raise ValidationError(self.message)
 
 
-class GrantForm(FlaskForm):
```
wjrm500, line 35: This was previously used specifically for the change name form, but now we can re-use the form from the grant setup flow 🌟

app/deliver_grant_funding/templates/deliver_grant_funding/settings/grant_change_ggis.html
```diff
@@ -0,0 +1,57 @@
...
+            "text": "No",
+            "checked": true if form.has_ggis.data == "no" else false
+          }
+        ]
+      })
+    }}
+
+    {{
+      form.submit(params={
+        "text": "Change grant GGIS"
```
wjrm500, line 53: This feels a bit wrong so should seek content guidance, same with the content in other submit buttons on change pages

app/deliver_grant_funding/routes.py
```diff
@@ -194,18 +193,72 @@ def grant_settings(grant_id: UUID) -> str:
     return render_template("deliver_grant_funding/grant_settings.html", grant=grant)
 
 
+@deliver_grant_funding_blueprint.route("/grants/<uuid:grant_id>/change-ggis", methods=["GET", "POST"])
+@mhclg_login_required
+@auto_commit_after_request
+def grant_change_ggis(grant_id: UUID) -> str | Response:
+    grant = interfaces.grants.get_grant(grant_id)
+    form = GrantGGISForm(
+        has_ggis=("yes" if grant.ggis_number else "no"),
```
wjrm500, line 202: We can't use `GrantGGISForm(obj=grant)` here as it's missing `has_ggis`

MarcUsher: This is _so_ close to `haggis`

srh-sloan: I can't unsee this now 🐑

samuelhwilliams (approved):
Nice 🎉

PR #268, opened 2025-06-05
FSPT-438 Fix: helper refactor missed template reference

sfount:
This should be added to the end to end tests whatever state they're in.


sfount:
This might be dumb or half a rebase or something

PR #273, opened 2025-06-06
Read-only view of submissions for the "applier"

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-548

## Description
We want to let people who have submitted their information to see what they submitted any time after the fact. This adds a read-only view of all the form runner pages and switches to that when a collection has been submitted, and adds some sanity checks to the pages and our CollectionHelper to help enforce that data integrity.

Also fixes some missing captions on form runner pages.

### DIFF
--- a/app/developers/templates/developers/ask_a_question.html
+++ b/app/developers/templates/developers/ask_a_question.html
@@ -20,6 +20,11 @@
       <form method="post" novalidate>
         {{ form.csrf_token }}
 
+        {# https://github.com/alphagov/govuk-frontend/issues/1841 #}
+        {# GOV.UK Frontend macros don't let you add captions to label/legend-headings that are outside of the <h> element #}
+        {# We want the caption outside of the <h> element to keep things more concise and understandable to screenreader users #}
+        {# So we just manually add the caption here. #}
+        <span class="govuk-caption-l">{{ question.form.title }}</span>
         {% if question.data_type == question_types.TEXT_SINGLE_LINE or question.data_type == question_types.TEXT_MULTI_LINE %}
           {{ form.question(params={"label": {"classes": "govuk-label--l", "isPageHeading": true} }) }}
         {% elif question.data_type == question_types.INTEGER %}


app/common/helpers/collections.py
```diff
@@ -72,6 +72,10 @@ def status(self) -> str:
         else:
             return CollectionStatusEnum.IN_PROGRESS
 
+    @property
+    def is_submitted(self) -> bool:
+        return self.status == CollectionStatusEnum.COMPLETED
```
nuwan-samarasinghe, line 77: The method is named `is_submitted()`, feels like something has been "submitted." but `COMPLETED` feels like it is done. So `COMPLETED` says that it is submitted ?

app/common/helpers/collections.py
```diff
@@ -182,13 +186,18 @@ def get_answer_for_question(self, question_id: UUID) -> TextSingleLine | TextMul
         return _deserialise_question_type(question, serialised_data) if serialised_data else None
 
     def submit_answer_for_question(self, question_id: UUID, form: DynamicQuestionForm) -> None:
+        if self.is_submitted:
+            raise ValueError(
```
nuwan-samarasinghe, line 190: How the front end is handled with this are we showing any errors to the user ?

app/developers/templates/developers/check_your_answers.html
```diff
@@ -20,18 +20,20 @@
...
-      <h1 class="govuk-heading-l">Check your answers</h1>
+      <span class="govuk-caption-l">{{ schema_form.title }}</span>
+
+      <h1 class="govuk-heading-l">{{ "Your submitted answers" if collection_helper.is_submitted else "Check your answers" }}</h1>
 
       {% set rows = [] %}
       {% for question in collection_helper.get_ordered_visible_questions_for_form(schema_form) %}
         {% set answer = collection_helper.get_answer_for_question(question.id) %}
         {% if answer is not none %}
+          {% set actions = [] if collection_helper.is_submitted else [{"href": url_for('developers.ask_a_question', collection_id=collection_helper.collection.id, question_id=question.id, source=back_link_source_enum.CHECK_YOUR_ANSWERS.value), "text": "Change", "visuallyHiddenText": question.name}] %}
```
nuwan-samarasinghe, line 31:
```
{% set actions = [
    {
      "href": url_for(
        'developers.ask_a_question',
        collection_id=collection_helper.collection.id,
        question_id=question.id,
        source=back_link_source_enum.CHECK_YOUR_ANSWERS.value
      ),
      "text": "Change",
      "visuallyHiddenText": question.name
    }
  ] %}
```
  
  shall we expand this like this current is taken small space but seems readability is low

app/common/helpers/collections.py
samuelhwilliams: Yeah this was maybe a half-baked idea by me - I will switch back to `is_completed` for now so it's consistent. But I like something about "submitted".

app/common/helpers/collections.py
samuelhwilliams: If this error does get thrown, it won't be handled in the frontend. It should be very hard for a user to trigger it - we'll need multiple users doing things at the same time and getting well out of sync.

Happy to address this the first time we see it come up in sentry.

app/developers/templates/developers/check_your_answers.html
samuelhwilliams: Good call - done :)

MarcUsher (approved):
Looks good! I got some strange behaviour locally with filling in collection but not able to reproduce consistently so probably a local set-up thing (but will try to capture steps if it persists). Also that's not linked to this PR I don't think as it's in the initial preview & filling in answers flow. This looks great!

If I try to URL hack and paste in an old question form page I get an attribute error of `_DynamicQuestionForm' object has no attribute 'is_completed'` - is that expected or should we be handling this differently?

PR #286, opened 2025-06-09
Rename core collections tables

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-535

CollectionSchema->Collection
Collection->Submission
CollectionMetadata->SubmissionEvent

And cascade this through the entire codebase and (developer) UI designs.

Given this is an extremely widespread refactor, we have accepted that we will wipe the databases of all environments fully and lose the history of everything prior to this change for DB migrations. It does not feel relevant to retain, and with no environments that currently handle data that we need to keep around forever, we can simplify the change significantly by taking this hit.

A `collection` for now is the definition of the data that we want the user to provide.

A `submission` for now is the actual data the user provides, in line with the collection.

The goal is for the SubmissionEvent table to be an immutable log of what has happened to submissions; metadata implies a mutable point in time K:V representation so call them 'event's to represent immutability.

These terms are still up for grabs pending user research.

---

To update the Notify template interface I've made a [v2 of the existing email](https://www.notifications.service.gov.uk/services/239747da-5aa1-4fe0-85ab-39d0272ca5c8/templates/2ff34065-0a75-4cc3-a782-1c00016e526e); we should delete the old one when this has been deployed and rename the v2 to removing the versioning reference.

## For a future PR
- Renaming the template files themselves - this is to simplify some of @sfount's refactoring for a WIP branch.
- Handle the grossness of "Submission submitted"


app/developers/forms.py
```diff
@@ -8,7 +8,7 @@
 
 
 class PreviewCollectionForm(FlaskForm):
-    submit = SubmitField("Preview this collection", widget=GovSubmitInput())
+    submit = SubmitField("Preview a submission", widget=GovSubmitInput())
```
sfount, line 11: I think this continues to make sense in the context it was in and doesn't need changing - you're configuring a "collection" by setting up sections/ forms/ questions etc. and this button is allowing you to preview how that collection all comes together into a task list

sfount: I guess _technically_ it should previously have said "Preview this schema"

sfount: The fact that it creates a submission in order to allow you to preview the behaviour feels a bit like an underlying detail

app/developers/templates/developers/collection_submit_confirmation.html
```diff
@@ -2,21 +2,20 @@
 {% extends "developers/access_grant_funding_base.html" %}
 
 {% block pageTitle %}
-  Collection submitted {{ collection_helper.name }} - MHCLG Funding Service
+  Submission submitted {{ submission_helper.name }} - MHCLG Funding Service
```
sfount, line 5: Just for good measure i'll make a note that I'm happy to push for the non-domain specific "completed" language for now, I didn't have conviction earlier but it looks clear seeing it all in the round

The helper already uses langauage around the generic mechanism like `submission.is_completed` so lets line everything up with that language and use domain language (i.e competitive application for funding is submitted) when we come to it

"Submission completed"
`SUBMISSION_COMPLETED`

app/developers/templates/developers/collection_tasklist.html
```diff
@@ -74,18 +71,18 @@ <h2 class="govuk-heading-m">{{ section.title }}</h2>
       {% endfor %}
     </div>
   </div>
-  {% if collection_helper.status != statuses.COMPLETED %}
+  {% if submission_helper.status != statuses.COMPLETED %}
```
sfount, line 74: Doesn't need to happen here but just to remind myself:
```suggestion
  {% if not submission_helper.is_completed %}
```

app/developers/templates/developers/list_collections.html
```diff
@@ -22,33 +22,33 @@
...
   </h1>
-  <p class="govuk-body">View the information that has been provided by users for a schema.</p>
+  <p class="govuk-body">View the information that has been provided by users for a submission.</p>
 
   {% set rows=[] %}
-  {# collections is probably a list of collection helpers? to be decided #}
-  {% for collection in collections %}
-    {% set link_to_collection %}
-      <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('developers.manage_collection', collection_id=collection.id) }}">{{ collection.reference }}</a>
+  {# submissions is probably a list of submission helpers? to be decided #}
```
sfount, line 31: It is - this can probably be removed 🫣

app/services/notify.py
```diff
@@ -90,15 +90,16 @@ def send_magic_link(
...
         return self._send_email(
-            collection.created_by.email,
+            submission.created_by.email,
             current_app.config["GOVUK_NOTIFY_COLLECTION_SUBMISSION_TEMPLATE_ID"],
             personalisation={
-                "collection name": collection.collection_schema.name,
-                "collection reference": collection.reference,
-                "collection url": url_for(
-                    "developers.collection_tasklist", collection_id=collection.id, _external=True
+                # todo: rename to submission
```
sfount, line 98: `# todone`

tests/unit/app/services/test_notify.py
```diff
@@ -49,30 +49,29 @@ def test_send_magic_link(self, app):
 
     @responses.activate
     def test_send_collection_submission(self, app, factories):
-        collection_schema = factories.collection_schema.build(name="Submission collection")
-        collection = factories.collection.build(
+        submission = factories.submission.build(
             id=uuid.UUID("10000000-0000-0000-0000-000000000000"),
-            collection_schema=collection_schema,
+            collection__name="My test collection",
```
sfount, line 54: Does this `__` syntax let you set properties on a sub factory? 🤯 or am I missing something?

sfount (approved):
A lot of the code reads more clearly now and this feels very comprehensive well done 👍 

The tiniest nit might be to add one line into the commit message about renaming metadata -> events so that it shows up when blaming those models/ methods

A few other things scattered but nothing that couldn't be done in a follow up? I'll leave it up to you

I've tried to break this in most the ways I know how locally and it stands up to it

app/developers/templates/developers/collection_submit_confirmation.html
sfount: I was definitely fine with "submitted" as a neutral enough term in the before times but really reaching here

app/developers/forms.py
samuelhwilliams: ok ok fair, will revert 👍

app/developers/templates/developers/collection_submit_confirmation.html
samuelhwilliams: agree

PR #290, opened 2025-06-10
BAU Tweak developer pages for consistency

sfount:
This proposes some minor tweaks to the developer form building pages to make them consistent and line them up with the original sketches. The journey here will definitely be changed and refined a lot but lets start with something consistent

The main change is separating out the ordering and management of sections into the "Manage" sections pages and keeps all forms/ questions links in the main collection tasklist builder view.

The original intention of that was so that the collection overview/ "tasklist" page was kept clear of many "add" buttons and "move up/ move down" links as well as being the main entry point/ overview for managing forms and questions for your collection (that would be accessed all the time).

The hypothesis being the number of times you need to add/ order sections and forms would be a lot less frequent than adding/ managing questions.

This definitely results in a lot more clicks and possibly some unintuitive separation until you're used to it - we should revise this journey/ design as we need to but lets at least make it fully consistent as a base so that we can then change it intentionally.

This revises the current version where there's a hybrid method of editing a form which is to either click in from the collection detail/ tasklist page or to go through "list sections" -> "manage section" -> "manage form". Just in watching product sign offs I've seen that happen a number of times and it doesn't feel like good repetitive behaviour.

Smaller fixes: 
- make sure we have consistent back link behaviour between all of the pages
- the list/ "Add X" button pattern should be consistent between all pages, this should be that the secondary button to "Add X" is always shown and the page either shows the list of things or some text saying there are none
- headings and caption use should be consistent
- all pages should be 2/3 consistently
- formatting issues
- don't use a "tasklist" HTML component in the case where its not actually a task list to fill out - this could have some very strange effects when using assistive technologies
- make the question text the link in the questions list to visually differentiate it from other up/ down pages
- removes the large "edit question type" disclaimer to try and bring the add and edit question pages closer together - after seeing that for the hundredth time that inset text taking up 1/3 of the page would start to get distracting. I think the "Add question page" showing the question type but not letting it be changed describes this clearly enough - something that would be good to sense check when putting it in front of users!

Updates the end to end tests to understand the tweaked layout.

### DIFF
--- a/app/assets/main.scss
+++ b/app/assets/main.scss
@@ -142,3 +142,11 @@ $app-destructive-link-active-colour: $govuk-text-colour;
     margin-left: auto;
   }
 }
+
+.app-tasklist-builder {
+    // extends the existing - want to think this through more
+    .govuk-summary-list__key {
+        font-weight: normal;
+        width: 50%;
+    }
+}
--- a/app/common/templates/common/macros/move-up-down-table.html
+++ b/app/common/templates/common/macros/move-up-down-table.html
@@ -1,9 +1,17 @@
 {% from "govuk_frontend_jinja/components/button/macro.html" import govukButton %}
+{# todo: I think all of this can be accomplished by overriding some design system classes #}
+{#       which would allow us to use the standard summary list macro with a few style changes #}
 {% macro moveUpDownTable(table_rows) %}
   <dl class="govuk-summary-list app-move-up-down-table">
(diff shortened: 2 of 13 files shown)


samuelhwilliams (approved):
Really nice tidy up - thank you for getting this sorted and through 🙇 Feels much better (small caveat around managing section+form order from that separate 'manage section' bit, but fine for now).

PR #291, opened 2025-06-10
FSPT-535 Change template names to line up with models

sfount:
We changed the model names collection -> submission and schema ->
submission. We intentionally out of scoped template names to avoid
conflicts with existing change requests - hopefully we can safely make
this all consistent now.


sfount:
Note the diff is going to be a bit strange and might need a double take because (for example) "list_collections.html" already existed it just meant something different so it will now be fully diffing the two unrelated HTML files

PR #294, opened 2025-06-10
Change basic auth password for caddy (pullpreview)

samuelhwilliams:
Using `caddy hash-password`


samuelhwilliams:
Ok, we didn't use this "password" because it gives an error: "hashedSecret too short to be a bcrypted password". So let's stick with what we've got.

PR #296, opened 2025-06-10
Use the app fixture, remove unneeded context

samuelhwilliams:
The core app fixture provides a properly-configured Flask instance, so we don't need to set SECRET_KEY manually.

These tests also don't need an explicit test request context - we have a db_session fixture that wraps every test in its own app context, which comes with a test request context as well.


nuwan-samarasinghe (approved):
nice I think most of them are copy past issue probably now it is fixed so good to go

PR #301, opened 2025-06-11
BAU_bug fixes for the user management: Adding missing member login changes and adding unit tests

nuwan-samarasinghe:
In this PR I am fixing a issue I noticed under this ticket [FSPT-528](https://mhclgdigital.atlassian.net/browse/FSPT-528) some how the change I did was missing in the previous merged PR.

When a user logs in as a MEMBER without an invitation from a platform admin, the system should return a 403 Access Denied error. However, if the user has been invited to the grant by an admin, they should be able to log in successfully without encountering an error. This update includes the necessary fix to implement that functionality.

[FSPT-528]: https://mhclgdigital.atlassian.net/browse/FSPT-528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -108,7 +108,7 @@ def sso_get_token() -> ResponseReturnValue:
         )
         add_user_role(user_id=user.id, role=RoleEnum.ADMIN)
 
-    elif not user or not user.roles:  # TODO: also allow to log in if they're a member of a grant.
+    elif not user or not user.roles:
         return render_template(
             "common/auth/mhclg-user-not-authorised.html", service_desk_url=current_app.config["SERVICE_DESK_URL"]
         ), 403
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -9,7 +9,7 @@ from app.common.auth.decorators import redirect_if_authenticated
 from app.common.auth.forms import ClaimMagicLinkForm, SignInForm, SSOSignInForm
 from app.common.auth.sso import build_auth_code_flow, build_msal_app
 from app.common.data import interfaces
-from app.common.data.interfaces.user import add_user_role, get_or_create_user
+from app.common.data.interfaces.user import add_user_role, get_or_create_user, get_user_by_email
 from app.common.data.types import RoleEnum
 from app.common.security.utils import sanitise_redirect_url
 from app.extensions import auto_commit_after_request, notification_service
@@ -98,13 +98,17 @@ def sso_get_token() -> ResponseReturnValue:
         return abort(500, "Azure AD get-token flow failed with: {}".format(result))
 
     sso_user = result["id_token_claims"]
-
-    user = get_or_create_user(email_address=sso_user["preferred_username"], name=sso_user["name"])
+    user = get_user_by_email(email_address=sso_user["preferred_username"])
     redirect_to_path = sanitise_redirect_url(session.pop("next", url_for("index")))
 
     if "FSD_ADMIN" in sso_user.get("roles", []):
+        user = get_or_create_user(
+            email_address=sso_user["preferred_username"],
+            name=sso_user["name"],
+        )
         add_user_role(user_id=user.id, role=RoleEnum.ADMIN)
-    else:  # TODO: also allow to log in if they're a member of a grant.
+
+    elif not user or not user.roles:  # TODO: also allow to log in if they're a member of a grant.
         return render_template(
             "common/auth/mhclg-user-not-authorised.html", service_desk_url=current_app.config["SERVICE_DESK_URL"]
         ), 403
--- a/app/common/data/interfaces/user.py
+++ b/app/common/data/interfaces/user.py
@@ -3,7 +3,7 @@ from typing import Optional, Sequence, cast
 
 from flask_login import current_user
 from sqlalchemy.dialects.postgresql import insert as postgresql_upsert
-from sqlalchemy.exc import IntegrityError
+from sqlalchemy.exc import IntegrityError, NoResultFound
 from sqlalchemy.sql.expression import select
 
 from app.common.data.interfaces.exceptions import InvalidUserRoleError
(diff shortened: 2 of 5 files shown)


app/common/auth/__init__.py
```diff
@@ -30,6 +30,8 @@ def request_a_link_to_sign_in() -> ResponseReturnValue:
         email = cast(str, form.email_address.data)
 
         user = get_or_create_user(email_address=email)
+        if user is None:
+            abort(404)
```
samuelhwilliams, line 34: Is this possible with the default behaviour of `get_or_create_users`? It would always create a user; how would we get back None? I think you're addressing a type hinting issue here that doesn't reflect the reality of the code execution path - let's fix the problem, not the symptom. :)

app/common/data/interfaces/user.py
```diff
@@ -29,7 +29,12 @@ def get_platform_admin_users() -> Sequence[User]:
     ).all()
 
 
-def get_or_create_user(email_address: str, name: Optional[str] = None) -> User:
+def get_or_create_user(email_address: str, name: Optional[str] = None, read_only: bool = False) -> Optional[User]:
+    if read_only:
+        try:
+            return db.session.execute(select(User).where(User.email == email_address)).scalar_one()
+        except NoResultFound:
+            return None
```
samuelhwilliams, line 37: I don't think the interface here feels very sensible - the function is called "get or create user", but then you take an argument that turns off half of that (the ability to create).

Can we just keep this as-is and add a separate "get_user" method that is read-only? Then use the appropriate one in views please.

app/deliver_grant_funding/routes.py
```diff
@@ -209,6 +209,8 @@ def add_user_to_grant(grant_id: UUID) -> ResponseReturnValue:
             user = next((user for user in grant.users if user.email == form.user_email.data), None)
             if user is None:
                 created_user = interfaces.user.get_or_create_user(email_address=form.user_email.data)
+                if created_user is None:
+                    abort(404)
```
samuelhwilliams, line 213: Same here, this is a symptom of the underlying problem that creates noise and confusion - let's please remove it :)

tests/integration/common/auth/test_auth.py
```diff
@@ -192,6 +193,42 @@ def test_get_without_fsd_admin_role(self, app, anonymous_client):
         assert response.status_code == 403
         assert "https://mhclgdigital.atlassian.net/servicedesk/customer/portal/5" in response.text
 
+    def test_get_without_fsd_admin_role_and_with_invite(self, app, anonymous_client, factories):
```
samuelhwilliams, line 196:
```suggestion
    def test_get_without_fsd_admin_role_and_with_grant_member_role(self, app, anonymous_client, factories):
```

tests/integration/common/auth/test_auth.py
```diff
@@ -192,6 +193,42 @@ def test_get_without_fsd_admin_role(self, app, anonymous_client):
...
+            }
+
+            response = anonymous_client.get(url_for("auth.sso_get_token"), follow_redirects=True)
+
+        assert response.status_code == 200
+        soup = BeautifulSoup(response.data, "html.parser")
+        nav_items = [li.get_text(strip=True) for li in soup.find("ul", id="navigation").find_all("li")]
+        assert len({"Home", "Grant details", "Grant team"} - set(nav_items)) == 0
+
+    def test_get_without_fsd_admin_role_and_with_no_invite(self, app, anonymous_client):
```
samuelhwilliams, line 217:
```suggestion
    def test_get_without_fsd_admin_role_and_with_no_roles(self, app, anonymous_client):
```

tests/integration/common/auth/test_auth.py
```diff
@@ -192,6 +193,42 @@ def test_get_without_fsd_admin_role(self, app, anonymous_client):
...
+                    "roles": [],
+                }
+            }
+
+            response = anonymous_client.get(url_for("auth.sso_get_token"), follow_redirects=True)
+
+        assert response.status_code == 200
+        soup = BeautifulSoup(response.data, "html.parser")
+        nav_items = [li.get_text(strip=True) for li in soup.find("ul", id="navigation").find_all("li")]
+        assert len({"Home", "Grant details", "Grant team"} - set(nav_items)) == 0
```
samuelhwilliams, line 215: I don't think it's good to assert on the nav bar here - changes to that will break this test and it's slightly unrelated.

Maybe we just check that the endpoint sends a redirect to the expected place?

tests/integration/common/auth/test_auth.py
nuwan-samarasinghe: renamed

srh-sloan (commented):
Code and tests look good, but can you add either a ticket reference for the bugs or a description of what the problems were to the PR description please?

app/common/data/interfaces/user.py
```diff
@@ -29,6 +29,13 @@ def get_platform_admin_users() -> Sequence[User]:
     ).all()
 
 
+def get_user_by_email(email_address: str) -> Optional[User]:
+    try:
+        return db.session.execute(select(User).where(User.email == email_address)).scalar_one()
+    except NoResultFound:
+        return None
```
sfount, line 36: totally optional but I think this can more succinctly be written as this (I've not tried this so if doing please do check it passes tests and does what you'd expect! note: neither of these handle the case where there's for some reason too many users for the email which I think is valid as the DB should constrain that in theory)

```suggestion
return db.session.execute(select(User).where(User.email == email_address)).scalar_one_or_none()```

nuwan-samarasinghe: it works & I did not know that they have this in place thanks @sfount

samuelhwilliams (approved):

PR #310, opened 2025-06-13
FSPT-561: Evaluate (python-like) expressions

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-561

We're starting to bring in the idea of validation and conditional
routing to the form builder/runner. We need to be able to describe how
these rules work and want a simple, powerful, and flexible system that
will let us do straight-forward and complex evaluations. These might
vary from things like "is the answer bigger than 100", to "if the answer
to question A was X, then this answer must be lower than the sum of all
their previous reported spending".

We feel like it's possible to have a single unified system under the
hood that can handle all of validation, conditional question behaviour,
and injecting values into text (eg questions and error messages) and
think it would be simpler to manage and maintain than doing anything too
unique for each of those features.

The idea here is to allow writing (a subset of) Python statements and
evaluating those. Traditionally this was done with `eval()` but there
are many and varied security concerns with doing this. Using Python-like
language should simplify learning how to write expressions (at least for
developers), but as this will handle untrusted user input we need to be
extremely cautious in our approach so as not to create huge security
vulnerabilities. It absolutely must be impossible for expressions to
result in arbitrary code execution.

We use a library called `simpleeval` which takes an expression and
converts it to the abstract syntax tree (AST) and then manually
processes that tree. We can easily add or remove nodes that we want to
be allowable in expressions, which should provide a good level of
protection against dangerous operations, as long as we configure the
allowlist of nodes correctly.

With this system, an expression might look like any of these:

`answer < 100`
`answer == "hello"`

In the future we might be able to do more interesting things like:
`answer <= grant.allocated_funding`
`answer <= question_a.answer`
`answer <= sum(forecasted_spends)`

With this expression system we're able to define the context available
to the expression (eg the set of variables/functions that can be used).
This allows us to expose some data/context to the expression, such as
the question's answer, or answers to all other questions, or information
about the grant they're providing data for.

But a lot of this is left as an extension exercise for the future.

### DIFF
--- a/app/common/expressions/__init__.py
+++ b/app/common/expressions/__init__.py
@@ -1,0 +1,101 @@
+import ast
+import uuid
+from typing import Any
+
+import simpleeval
+
+from app.common.data.models import Expression
+
+
+class BaseExpressionError(Exception):
+    pass
+
+
+class UndefinedVariableInExpression(BaseExpressionError):
+    pass
+
+
+class DisallowedExpression(BaseExpressionError):
+    pass
+
+
+class InvalidEvaluationResult(BaseExpressionError):
+    pass
+
+
+def _evaluate_expression_with_context(
+    expression: Expression, context: dict[str, str | int | float | bool | None] | None = None
+) -> Any:
+    """
+    The base evaluator to use for handling all expressions.
+
+    This parses arbitrary Python-language text into an Abstract Syntax Tree (AST) and then evaluates the result of
+    that expression. Parsing arbitrary Python is extremely dangerous so we heavily restrict the AST nodes that we
+    are willing to handle, to (hopefully) close off the attack surface to any malicious behaviour.
+
+    The addition of any new AST nodes should be well-tested and intentional consideration should be given to any
+    ways of exploit or misuse.
+    """
+    expr_context = expression.context or {}
+    context = context or {}
+
+    if context_overlap := set(expr_context).intersection(set(context)):
+        raise ValueError(
+            f"Cannot safely evaluate with overlapping contexts. "
+            f"The following keys exist in both the expression.context and additional context: {context_overlap}."
+        )
+
+    merged_context = {**expr_context, **context}
+
+    # May want EvalWithCompoundTypes at some point, but for now simple+very limited is OK.
+    evaluator = simpleeval.SimpleEval(names=merged_context)  # type: ignore[no-untyped-call]
+
+    # Remove all nodes except those we explicitly allowlist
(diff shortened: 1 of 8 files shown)


app/common/expressions/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    that expression. Parsing arbitrary Python is extremely dangerous so we heavily restrict the AST nodes that we
+    are willing to handle, to (hopefully) close off the attack surface to any malicious behaviour.
+
+    The addition of any new AST nodes should be well-tested and intentional consideration should be given to any
+    ways of exploit or misuse.
+    """
+    expr_context = expression.context or {}
+    context = context or {}
+
+    if context_overlap := set(expr_context).intersection(set(context)):
```
sfount, line 42: This is a really great guard and feels like it will have a great developer experience in the failure if we come across it 👍 great forward thinking

app/common/expressions/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # do we want these to evalaute to non-bool types like int/str ever?
+    if not isinstance(result, bool):
+        raise InvalidEvaluationResult(f"Result of evaluating {expression=} was {result=}; expected a boolean.")
+
+    return result
+
+
+def mangle_question_id_for_context(question_id: uuid.UUID) -> str:
+    # todo: work out how to refer to questions in an expressions-compatible way without doing this mangling
```
sfount, line 99: 👍 a good thing for us to think through, the prefix and hex version feel good enough for us to get started with. Separately but possibly two-birds we need to think through how we identify the same question across versions

app/common/expressions/managed.py
```diff
@@ -41,5 +43,9 @@ def message(self) -> str:
 
     @property
     def expression(self) -> str:
-        # todo: are UUIDs parsable by the expression parser/ language
-        return f"(( {self.question_id} )) > {self.minimum_value}"
+        # todo: how much of these should be calculated from the context and run
+        #       through the eval vs. stored in the expression - this is leaning towards the latter
+        # todo: are UUIDs parsable by the expression language
```
sfount, line 48: I _think_ these have slipped back in from a rebase and are largely answered by this PR - ideally they could be removed. The final todo could either stay or leave up to yo

sfount (approved):
really nice coverage and the commit message is really thorough and helpful 👍 there's a few comments that have sneaked in from my previous commit which can be removed now or later

PR #311, opened 2025-06-13
FSPT-564 Add question condition pages

sfount:
Add a question condition pages and flow. Note this PR just uses the "greater than" number condition to prove the concept but it isn't set up to reject questions that are not the correct type.

This is largely to agree a flow and show how expressions will be consumed and probably needs more thought.

Clicking to add a condition to a question, prompts you to select which answer you will depend on - note that it shows the context is this collection and this form (showing all questions in that form) which in the future we'll allow you to change to refer to other data in your collections/ grants. For now thats read only as we've limited the scope to your current form
Screenshot 2025-06-13 at 12 27 19

Choose what the criteria for your question showing up is
Screenshot 2025-06-13 at 12 27 30

If the form has no other questions with a supported question type to let you depend on 
Screenshot 2025-06-18 at 08 19 54


app/common/expressions/forms.py
```diff
@@ -0,0 +1,27 @@
...
+from wtforms import IntegerField, RadioField, SubmitField
+from wtforms.validators import DataRequired, Optional
+
+from app.common.data.types import ManagedExpressions
+
+
+class AddNumberConditionForm(FlaskForm):
+    type = RadioField(
+        "Only show the question if the answer is",
+        choices=[(ManagedExpressions.GREATER_THAN, ManagedExpressions.GREATER_THAN.value)],
```
samuelhwilliams, line 14: Some kind of `enum_to_choices(EnumClass)` helper would be good for this soon enough.

tests/unit/common/test_expressions.py
samuelhwilliams, line 1: Test file paths ideally match the app file paths - ie `common/data/expressions/test_managed.py` for tests on things in common/data/expressions/managed.py

tests/unit/common/test_expressions.py
```diff
@@ -101,3 +108,38 @@ def test_additional_context(self):
     def test_raise_on_non_boolean_result(self):
         with pytest.raises(InvalidEvaluationResult):
             evaluate(Expression(statement="1"))
+
+
+class TestManagedExpressions:
+    def test_filter_supported_question_types(self, factories):
```
samuelhwilliams, line 114: Another test asserting that the same question isn't returned by `get_supported_form_questions` would be good 🙇

samuelhwilliams (approved):
nice one, good baseline 🙇

PR #319, opened 2025-06-16
Ignore case for email and updating some tests

nuwan-samarasinghe:


tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -965,71 +965,107 @@ def test_grant_check_your_answers_post_creates_grant(authenticated_platform_admi
...
     )
-    soup = BeautifulSoup(result.data, "html.parser")
-    button = soup.find("a", string=lambda text: text and "Add grant team member" in text)
-    assert button is not None, "'Add grant team member' button not found"
-    email_address = "test1@communities.gov.uk"
-    result = authenticated_platform_admin_client.post(
+    soup = BeautifulSoup(response.data, "html.parser")
+    assert soup.find("a", string=lambda text: text and "Add grant team member" in text), (
+        "'Add grant team member' button not found"
+    )
```
samuelhwilliams, line 978: There is a separate test covering this now - let's remove the assertion here

```suggestion
```

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -965,71 +965,107 @@ def test_grant_check_your_answers_post_creates_grant(authenticated_platform_admi
...
+    factories.user_role.create(user=user, role=RoleEnum.MEMBER, grant=grant)
+    response = authenticated_member_client.get(url_for("deliver_grant_funding.list_users_for_grant", grant_id=grant.id))
+    soup = BeautifulSoup(response.data, "html.parser")
+    # UI assertions
+    assert soup.find("a", string=lambda text: text and "Add grant team member" in text) is None
+    assert all(header in [th.get_text(strip=True) for th in soup.find_all("th")] for header in ["Email", "Role"])
+    assert "Grant team" in soup.h1.text
+    # BE assertions
+    users = templates_rendered[0][1].get("grant").users
+    assert users and len(users) == 1
```
samuelhwilliams, line 1071: You're a bit inconsistent about your assertions - sometimes you store the thing you're asserting against in a nice variable (eg `users`), and sometimes you don't.

I think we should prioritise readability in tests, so if we can get anything long and awkward not inlined, that'd be great. Example:

```suggestion
    # UI assertions
    add_grant_team_member_button = soup.find("a", string=lambda text: text and "Add grant team member" in text)
    headers = {th.get_text(strip=True) for th in soup.find_all("th")}
    assert add_grant_team_member_button is None
    assert headers == {"Email", "Name", "Role"}
    assert "Grant team" in soup.h1.text

    # BE assertions
    users = templates_rendered[0][1].get("grant").users
    assert users and len(users) == 1
```

PR #321, opened 2025-06-17
Adding fixture change to have proper key for the templates and ignore govuk templates

nuwan-samarasinghe:
Currently in the fixture we store the templates and context in an array and we have to access those using array element and it is bit confusing and here we have introduced to have endpoint and against that we have the template as well as the context & with in tests we should be able to identify the details based on the endpoint  

```
{'deliver_grant_funding.add_user_to_grant': TemplateRenderRecord(template=<Template 'deliver_grant_funding/grant_team/grant_user_add.html'>, context={'form': <app.deliver_grant_funding.forms.GrantAddUserForm object at 0x1117b5f90>, 'grant': <app.common.data.models.Grant object at 0x111946d50>, 'g': <flask.g of 'app'>, 'request': <Request 'http://funding.communities.gov.localhost:8080/grant/share/user/816d6862-a3de-4330-a05a-130f049053bc' [POST]>, 'session': <SecureCookieSession {}>, 'vite_asset': <function FlaskAssetsViteExtension.assets_processor.<locals>.vite_asset at 0x1119d9440>, 'current_user': <app.common.data.models_user.User object at 0x11187de50>, 'format_date': <function format_date at 0x106b300e0>, 'format_date_short': <function format_date_short at 0x106b30180>, 'format_datetime': <function format_datetime at 0x106b30220>, 'format_date_range': <function format_date_range at 0x106b302c0>, 'format_datetime_range': <function format_datetime_range at 0x106b30360>, 'submission_mode_enum': <enum 'SubmissionModeEnum'>})}
```

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -229,7 +229,8 @@ def templates_rendered(app: Flask) -> Generator[TTemplatesRendered]:
         if template.name and "govuk_frontend_wtf" not in template.name and request and request.endpoint:
             endpoint = request.endpoint
             entry = TemplateRenderRecord(template=template, context=context)
-            recorded.setdefault(endpoint, entry)
+            if endpoint not in recorded:
+                recorded[endpoint] = entry
 
     template_rendered.connect(record, app)
     try:


tests/integration/conftest.py
```diff
@@ -222,10 +222,14 @@ def db_session(app: Flask, db: SQLAlchemy) -> Generator[Session, None, None]:
...
-    recorded = []
+    recorded: TTemplatesRendered = {}
 
     def record(sender: Flask, template: Template, context: dict[str, Any], **extra: dict[str, Any]) -> None:
-        recorded.append((template, context))
+        request = context.get("request")
+        if template.name and "govuk_frontend_wtf" not in template.name and request and request.endpoint:
+            endpoint = request.endpoint
+            entry = TemplateRenderRecord(template=template, context=context)
+            recorded.setdefault(endpoint, entry)
```
samuelhwilliams, line 232: is `setdefault` an intentional choice here? Do you think that's better than just doing a set?

PR #326, opened 2025-06-18
Centralise and standardise page titles

samuelhwilliams:
Establish page titles in the base templates and extend these in sub-templates. The most derived template can just set its fragment in `page_title` and it will be injected into the standard format correctly, ensuring that "MHCLG Funding Service" is always part of the page title.


MarcUsher (approved):

samuelhwilliams:
pre-commit.ci run

PR #327, opened 2025-06-18
FSPT-534 - Delegating case-sensitive checks to database

wjrm500:
When we add or change a grant, we need to do a case-insensitive check to see if the grant name exists in the database. Currently we implement that check in the Python which means manually lowercasing the strings before comparison, and failing to leverage the CITEXT PostgreSQL datatype that handles case-insensitive checking for us. The same is true for user email address when we add new users to grants.

This change removes the Python lowercasing logic and delegates responsibility for the check to the database, where case insensitivity is handled out of the box, with refactoring.


tests/integration/common/data/interfaces/test_grants.py
```diff
@@ -107,3 +108,24 @@ def test_updated_grant_nothing_provided(factories) -> None:
...
     assert updated_grant.primary_contact_email == "Initial Email"
     assert updated_grant.ggis_number == "GGIS-123456"
+
+
+class TestGrantNameExists:
+    def test_grant_name_exists_true(self, factories):
+        factories.grant.create(name="Existing Grant")
+
+        result = grant_name_exists("Existing Grant")
+        assert result is True
```
samuelhwilliams, line 118: Shall we test case insensitivity here too?

app/common/data/interfaces/user.py
```diff
@@ -22,14 +22,6 @@ def get_current_user() -> User:
     return user
 
 
-def get_platform_admin_users() -> Sequence[User]:
```
samuelhwilliams, line 25: Are we just deleting this because it's now unused?

I can see it being useful, so don't think deleting it is super necessary, but easy enough to add back too I guess :)

samuelhwilliams (approved):
nice 👍

tests/unit/app/deliver_grant_funding/test_forms.py
```diff
@@ -69,25 +49,29 @@ def test_grant_ggis_form_fails_when_yes_selected_and_empty(app: Flask):
...
-        grant=grant,
-    )
-    assert form.validate() is False
-    assert "already is a member of" in form.user_email.errors[0]
+def test_user_already_in_grant_users(app: Flask, factories):
+    grant = factories.grant.build(name="Test Grant")
+    user = factories.user.build(email="test.user@communities.gov.uk")
 
+    form = GrantAddUserForm(grant=grant)
+    form.user_email.data = "test.admin@communities.gov.uk"
```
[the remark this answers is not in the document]
wjrm500: Just find the separate line version slightly more readable + aligns with how the form object is instantiated in the route

app/common/data/interfaces/user.py
wjrm500: Exactly yeah, deleting as unused

PR #329, opened 2025-06-18
Bugfix - grant setup name backlink goes to grant setup name (same page)

wjrm500:
Should go back to GGIS.


nuwan-samarasinghe (approved):
Shall we add some tests to cover back link behaviour ?

nice spotting

PR #330, opened 2025-06-18
FSPT-463: Implement AuthorisationHelper for Role-Based Access Control (Platform Admin & Grant Member)

nuwan-samarasinghe:
**Ticket: https://mhclgdigital.atlassian.net/browse/FSPT-463**

This PR introduces a centralised `AuthorisationHelper` class to manage role-based access control across the Funding Service platform. It replaces ad-hoc permission checks and supports a growing hierarchy of roles, starting with:

Platform Admins (`FSD_ADMIN` group members): Full read/write access across all resources

Grant Team Members: Read-only access to specific Grants they are assigned to

## Acceptance Criteria:
-  Platform Admins retain full platform access
-  Grant Members can view assigned grants and nothing more
-  Non-assigned users see nothing even if logged in
-  AuthorisationHelper supports secure, repeatable permission logic
-  All Grant-related views and routes enforce correct role-based access

### DIFF
--- a/app/common/auth/authorisation_helper.py
+++ b/app/common/auth/authorisation_helper.py
@@ -1,7 +1,7 @@
 from uuid import UUID
 
 from app.common.data.models_user import User
-from app.common.data.types import RoleEnum
+from app.common.data.types import GRANT_ROLES_MAPPING, RoleEnum
 
 
 class AuthorisationHelper:
@@ -23,6 +23,8 @@ class AuthorisationHelper:
 
     @staticmethod
     def is_grant_admin(grant_id: str, user: User) -> bool:
+        if AuthorisationHelper.is_platform_admin(user=user):
+            return True
         is_grant_admin = any(
             role.role == RoleEnum.ADMIN and role.organisation_id is None and role.grant_id == grant_id
             for role in user.roles
@@ -31,14 +33,29 @@ class AuthorisationHelper:
 
     @staticmethod
     def is_grant_member(grant_id: str | UUID, user: User) -> bool:
-        # TODO account for hierarchical roles
-        # MEMBER_ROLES = [RoleEnum.MEMBER, RoleEnum.ADMIN]
-        # platform admin would have RoleEnum.ADMIN and grant_id would be None so need to change 'any()' call
-        # role.role in MEMBER_ROLES
+        """
+        Determines whether a user has permissions to act as a grant member.
+        Platform admin overrides anything else.
+        Checks
+        :param grant_id:
+        :param user:
+        :return:
+        """
+
+        if AuthorisationHelper.is_platform_admin(user=user):
+            return True
+
         if isinstance(grant_id, str):
             grant_id = UUID(grant_id)
+
         is_grant_member = any(
-            role.role == RoleEnum.MEMBER and role.organisation_id is None and role.grant_id == grant_id
+            # role.organisation_id is None TODO don't think this is needed, as we only check grant_id, but confirm
+            role.grant_id == grant_id
+            and RoleEnum.MEMBER
+            in GRANT_ROLES_MAPPING.get(
+                role.role,
+                role.role,  # Use the list of roles this role implies, otherwise just use this role
+            )
             for role in user.roles
         )
         return is_grant_member
@@ -46,13 +63,15 @@ class AuthorisationHelper:
     @staticmethod
     def has_grant_role(grant_id: str, user: User, role: RoleEnum) -> bool:
         """
-        Will return True if the user has the specified role for the grant, (TODO: or a higher role in the hierarchy).
-        Does not work for platform admin role, as that is not a grant-specific role.
+        Will return True if the user has the specified role for the grant.
+        Platform admin overrides anything else.
         :param grant_id:
         :param user:
         :param role:
         :return:
         """
+        if AuthorisationHelper.is_platform_admin(user=user):
+            return True
         match role:
             case RoleEnum.ADMIN:
                 return AuthorisationHelper.is_grant_admin(grant_id, user)
--- a/app/common/auth/authorisation_helper.py
+++ b/app/common/auth/authorisation_helper.py
@@ -1,3 +1,5 @@
+from uuid import UUID
+
 from app.common.data.models_user import User
 from app.common.data.types import RoleEnum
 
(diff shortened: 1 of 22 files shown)


app/common/auth/authorisation_helper.py
```diff
@@ -0,0 +1,82 @@
+from uuid import UUID
+
+from app.common.data.models_user import User
+from app.common.data.types import GRANT_ROLES_MAPPING, RoleEnum
+
+
+class AuthorisationHelper:
+    @staticmethod
+    def has_logged_in(user: User) -> bool:
```
samuelhwilliams, line 9: question/idea: what do you think about all of these methods have a default user value of the current user?

Then for 99% of cases you don't need to pass the user in and can just do `authorisation_helper.is_platform_admin()`. On the rare occassion you want to check _someone else's_ permission you can pass it in explicitly, but I think nearly always you'll be checking the permissions of the person making the request.

It's a bit more magical and a bit less explicit, but a bit more concise. Pros and cons. Happy for you to decide 🙇

app/common/auth/decorators.py
```diff
@@ -67,11 +69,35 @@ def platform_admin_role_required[**P](
...
+) -> Callable[[Callable[P, ResponseReturnValue]], Callable[P, ResponseReturnValue]]:
+    def decorator(func: Callable[P, ResponseReturnValue]) -> Callable[P, ResponseReturnValue]:
+        @functools.wraps(func)
+        def wrapped(*args: P.args, **kwargs: P.kwargs) -> ResponseReturnValue:
+            current_user = interfaces.user.get_current_user()
+
+            if AuthorisationHelper.is_platform_admin(user=current_user):
+                return func(*args, **kwargs)
+
+            if "grant_id" in kwargs:
```
samuelhwilliams, line 93: If we add the `has_grant_role` decorator to an endpoint that defines the variable as anything other than `grant_id` (accidentally or maybe for some unexpected reason), this check would get skipped altogether.

That feels like an insecure default.

Should we raise an error if we don't find a `grant_id` in the kwargs? That will help ensure consistency and always be a more "fail closed" secure default.

app/common/auth/authorisation_helper.py
```diff
@@ -0,0 +1,82 @@
...
+    @staticmethod
+    def is_platform_admin(user: User) -> bool:
+        is_platform_admin = any(
+            role.role == RoleEnum.ADMIN and role.organisation_id is None and role.grant_id is None
+            for role in user.roles
+        )
+        return is_platform_admin
+
+    @staticmethod
+    def is_grant_admin(grant_id: str, user: User) -> bool:
```
samuelhwilliams, line 25: Please we can make sure that UUIDs are typed as UUIDs everywhere, and that we don't pass them around as strings.

Although we're used to seeing UUIDs as strings, UUIDs are 128-bit values. Their 36-character alphanumeric representation is a stringified version of a UUID, but it is not the same as the UUID itself.

You can get some subtle bugs if you rely on treating them as strings, because things like case-sensitivity come into the equation.

All routes that take in UUIDs should type them properly as UUIDs in the decorator and so you should receive proper UUID objects in the view.

If we can feed this through everywhere it would be great and will prevent us having some strange problem in the future. For example: https://github.com/alphagov/notifications-admin/pull/4822

app/common/auth/decorators.py
srh-sloan: Nice idea, yes lets do this

nuwan-samarasinghe: yeah since this explicitly check for grant role before that what my idea was different let me update

nuwan-samarasinghe: Fixed

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -22,6 +23,8 @@
     SectionForm,
 )
 
+logger = logging.getLogger(__name__)
```
samuelhwilliams, line 26: This left in from some debugging or something?

```suggestion
```

tests/unit/app/common/auth/test_authorisation_helper.py
```diff
@@ -0,0 +1,121 @@
...
+        "role",
+        [
+            (RoleEnum.ADMIN),
+            (RoleEnum.MEMBER),
+        ],
+    )
+    def test_is_grant_admin_incorrect_grant(self, factories, role):
+        user = factories.user.build()
+        grant1 = factories.grant.build()
+        grant2 = factories.grant.build()
```
samuelhwilliams, line 55: nit/comment for future reference:

```suggestion
        grants = factories.grant.build_batch(2)
```

tests/unit/app/common/auth/test_authorisation_helper.py
```diff
@@ -0,0 +1,121 @@
...
+        assert AuthorisationHelper.is_grant_admin(user=user, grant_id=grant2.id) is False
+
+    @pytest.mark.parametrize(
+        "role, expected",
+        [
+            (RoleEnum.ADMIN, True),
+            (RoleEnum.MEMBER, False),
+        ],
+    )
+    def test_is_grant_admin_no_grant(self, factories, role, expected):
```
samuelhwilliams, line 66: What does the name of this test mean? 🤔 

```suggestion
    def test_is_grant_admin_for_grant_roles(self, factories, role, expected):
```

tests/unit/app/common/auth/test_authorisation_helper.py
```diff
@@ -0,0 +1,121 @@
...
+        [
+            (RoleEnum.ADMIN, True),
+            (RoleEnum.MEMBER, True),
+        ],
+    )
+    def test_is_grant_member_true(self, factories, role, expected):
+        user = factories.user.build()
+        grant = factories.grant.build()
+        factories.user_role.build(user=user, role=role, grant=grant)
+        assert AuthorisationHelper.is_grant_member(user=user, grant_id=grant.id) is expected
```
samuelhwilliams, line 83: This looks a bit like the test below, so could remove the Trues from the params

```suggestion
    @pytest.mark.parametrize("role", [RoleEnum.MEMBER, RoleEnum.ADMIN])
    def test_is_grant_member_true(self, factories, role):
        user = factories.user.build()
        grant = factories.grant.build()
        factories.user_role.build(user=user, role=role, grant=grant)
        assert AuthorisationHelper.is_grant_member(user=user, grant_id=grant.id) is True
```

tests/unit/app/deliver_grant_funding/test_forms.py
```diff
@@ -52,13 +52,14 @@ def test_grant_ggis_form_fails_when_yes_selected_and_empty(app: Flask):
...
     user = factories.user.build(email="test.user@communities.gov.uk")
+    factories.user_role.build(user=user, role=RoleEnum.MEMBER, grant=grant)
 
     form = GrantAddUserForm(grant=grant)
     form.user_email.data = "test.admin@communities.gov.uk"
 
     with (
         patch("app.deliver_grant_funding.forms.get_user_by_email", return_value=user),
-        patch("app.deliver_grant_funding.forms.get_all_grants_by_user", return_value=[grant]),
+        # patch("app.deliver_grant_funding.forms.get_all_grants_by_user", return_value=[grant]),
```
samuelhwilliams, line 62:
```suggestion
```

tests/unit/test_all_routes_access.py
```diff
@@ -0,0 +1,110 @@
...
+    all_declared_routes_in_test = (
+        routes_with_no_expected_access_restrictions
+        + routes_with_expected_mhclg_login_required_access
+        + routes_with_expected_member_only_access
+        + routes_with_expected_grant_admin_only_access
+        + routes_with_expected_platform_admin_only_access
+    )
+
+    all_routes_in_app = [rule.endpoint for rule in app.url_map.iter_rules()]
+    assert set(all_declared_routes_in_test) - set(all_routes_in_app) == set()
```
samuelhwilliams, line 110: Nice :)

app/deliver_grant_funding/forms.py
```diff
@@ -211,24 +212,22 @@ def validate(self, extra_validators: Any = None) -> bool:
...
+            if not user_to_add:
+                return True
+
+            if AuthorisationHelper.is_platform_admin(user_to_add) or AuthorisationHelper.is_grant_admin(
+                grant_id=self.grant.id, user=user_to_add
+            ):
+                self.user_email.errors = list(self.user_email.errors) + [
+                    "This user already exists as a Funding Service admin user so you cannot add them"
+                ]
+                return False
```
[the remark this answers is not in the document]
nuwan-samarasinghe: Fixed

tests/integration/deliver_grant_funding/test_routes.py
nuwan-samarasinghe: Fixed

tests/unit/app/common/auth/test_authorisation_helper.py
nuwan-samarasinghe: Fixed

tests/unit/app/deliver_grant_funding/test_forms.py
nuwan-samarasinghe: Fixed

PR #334, opened 2025-06-19
Make grant URLs consistent

samuelhwilliams:


app/deliver_grant_funding/routes.py
```diff
@@ -204,7 +204,14 @@ def list_grants() -> Response | str:
     return render_template("deliver_grant_funding/grant_list.html", grants=grants)
 
 
-@deliver_grant_funding_blueprint.route("/grant/users/<uuid:grant_id>", methods=["GET"])
+@deliver_grant_funding_blueprint.route("/grant/<uuid:grant_id>", methods=["GET"])
+@mhclg_login_required
+def view_grant(grant_id: UUID) -> str:
```
sfount, line 209: want to make these consistent too?
```suggestion
def view_grant(grant_id: UUID) -> ResponseReturnValue:
```

app/deliver_grant_funding/routes.py
```diff
@@ -67,7 +67,7 @@ def grant_setup_ggis() -> ResponseReturnValue:
     )
 
 
-@deliver_grant_funding_blueprint.route("/grants/setup/name", methods=["GET", "POST"])
+@deliver_grant_funding_blueprint.route("/grant/setup/name", methods=["GET", "POST"])
```
sfount, line 70: These maybe could reflect that they're not consistent pages with the `/grant` route sand can represent that they're separate with `/grant-setup/name` etc.

I don't think this is important but they don't feel the same from a glance

app/deliver_grant_funding/routes.py
```diff
@@ -218,7 +225,7 @@ def list_users_for_grant(grant_id: UUID) -> str:
     )
 
 
-@deliver_grant_funding_blueprint.route("/grant/share/user/<uuid:grant_id>", methods=["GET", "POST"])
+@deliver_grant_funding_blueprint.route("/grant/<uuid:grant_id>/users/add", methods=["GET", "POST"])
```
sfount, line 228: 🙏

sfount (commented):
Commenting in case you want to make any changes - very happy to approve this as-is

app/deliver_grant_funding/routes.py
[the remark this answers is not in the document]
samuelhwilliams: I'm going to keep the hierarchy the same - the main view is `/users`. I understand what you're saying is that we're only adding one user, but I think there's a small 🤏  difference between adding a user to the grant users, and viewing a single user. I think for viewing a specific user I'd agree that a `/user` feels good, but this is adding to the grant's _users_. So I think plural is still ok here.

srh-sloan (approved):
Please can you add a PR description?

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -931,10 +932,10 @@ def test_grant_check_your_answers_post_creates_grant(authenticated_platform_admi
     assert response.status_code == 302
 
     # Verify redirect is to grant setup confirmation page
-    assert "/setup-confirmation" in response.location
+    assert response.location == AnyStringMatching(r"^/grant-setup/[A-Za-z0-9-]+$")
```
wjrm500, line 935: This would presumably pass on any of the `grant-setup` routes, could we limit the check to confirm we're actually on the confirmation page?

app/deliver_grant_funding/routes.py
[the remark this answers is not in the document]
srh-sloan: I think `users` is a context here so it's not that you are adding multiple users, but you are inside the `users` area, and in an `add` function

PR #348, opened 2025-06-20
FSPT-564 List conditions for question

sfount:
List existing conditions for a given question. commit makes some decisions that we'll likely change based on question validation needs for expressions so will likely wait to be rebased with that.

Screenshot 2025-06-25 at 13 21 29

### DIFF
--- a/app/common/expressions/managed.py
+++ b/app/common/expressions/managed.py
@@ -11,11 +11,12 @@ from app.common.data.types import ManagedExpressionsEnum
 from app.common.expressions import mangle_question_id_for_context
 
 if TYPE_CHECKING:
-    from app.common.data.models import Expression
+    from app.common.data.models import Expression, Question
 
 
 class ManagedExpression(BaseModel):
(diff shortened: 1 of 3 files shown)


app/common/expressions/managed.py
```diff
@@ -26,6 +27,20 @@ def statement(self) -> str:
...
+    @property
+    @abc.abstractmethod
+    def message(self) -> str: ...
+
+    @property
+    def referenced_question(self) -> "Question":
+        # todo: split up the collections interface to let us sensibly reason about whats importing what
+        from app.common.data.interfaces.collections import get_question_by_id
+
+        # todo: this will do a database query per expression on the question - for now we'd anticipate
```
samuelhwilliams, line 39: As long as we're using db.session.get, SQLAlchemy's cache should short-circuit an actual DB query and just fetch the local object from cache. So this might look worse on the surface than it actually is.

samuelhwilliams (approved):
Lovely :)

PR #352, opened 2025-06-20
Seed grants with stable UUIDs

samuelhwilliams:
It can be a tiny bit annoying when the seeded
grant/collection/section/question IDs change when you down/up your local environment, because if you're on one of those pages and reload you can get up with a 404 between local app runs.

This injects some logic to generate stable UUIDs for each of these entities based on its name and hierarchy.


nuwan-samarasinghe (approved):
🥇

samuelhwilliams:
pre-commit.ci run

PR #354, opened 2025-06-23
Typo fixes for WordRange validator

samuelhwilliams:
- "fewer" rather than "less" for countable things
- support validating an exact word count
- update tests to validate the error message shown
- pass `field_display_name` lowercase


MarcUsher (approved):
> "fewer" rather than "less" for countable things
😍

nuwan-samarasinghe (approved):
👍

PR #356, opened 2025-06-23
Streamline dev experience for new managed expressions

samuelhwilliams:
This changeset will be rough to review; the best review is probably
looking wholesale at the actual end-result, which sits across three
files:
- app.common.expressions.forms
- app.common.expressions.managed
- app.common.expressions.registry

---

A bit of an exploration for improving the DX/reducing the number of
touchpoints that need to be edited when adding a new managed expression.

Before this change, adding a new managed expression meant changes across
the following set of places:
- app.common.data.types
  Update `ManagedExpressionsEnum` to define the name of the new
  expression.

- app.common.expressions.forms
  Define a form for the question type that explicitly lists all of the
  managed expression types it supports.

- app.common.expressions.helpers
  Add the above form to a lookup table mapping it against a question
  data type.

- app.common.expressions.managed
  Define the new managed expression as a pydantic model; this describes
  the data needed to define an instance.

- _number_expression.html
  Render the fields for the new expression into a variable and pass them
  into the render for `form.type`. You need to manually care about the
  order you're passing the blobs of HTML into the `type` field
  rendering, because they're implicitly linked.

- generating a migration for the managed expression enum change.

---

And then if you're also adding a new question type altogether, you need
to:

- add_question_condition_select_condition_type.html
  Check the question type and render the equivalent of
  _number_condition.html

- add_question_validation.html
  Check the question type and render the equivalent of
  _number_validation.html

---

This patch reduces the number of touchpoints down to three, in theory:
- app.common.data.types (add an entry to the ManagedExpressionsEnum)
- app.common.expressions.managed (create the new managed expression by
extending the BaseExpression class).
- app.common.data.migrations.versions (new migration for the managed expression
  enum change).

Each managed expression is now a slightly thicker class that declares
both the data needed to define it (the pydantic model equivalent), and
the form fields+metadata needed to collect that successfully from a form
builder user. This mixes purposes slightly, but does co-locate all of
the changes in one place which - theoretically - should make it a
simpler task. At least that's the theory I guess I was exploring here

---

### Known limitations
- Can't pass `params` to field rendering inside Jinja2 templates because the field/subfield rendering is handled automatically
- Because we aren't defining individual forms for each question type, this doesn't support cross-field/form-wide validation for a managed expression, but [that might be solvable separately](https://github.com/communitiesuk/funding-service/pull/395)
- Further extension of this (or some way to break out) would be needed if the way a (form designer) user needs to provide the info for a managed expression should happen across multiple pages. The current implementation exposes the managed expression fields as conditionally-revealed; it could be very easily adapted to always show them on a whole new page. But doing some managed expressions as conditional reveals and some as a new page is not currently supported here.

### DIFF
--- a/app/common/expressions/forms.py
+++ b/app/common/expressions/forms.py
@@ -21,11 +21,12 @@ class _ManagedExpressionForm(FlaskForm):
     _managed_expressions: list[type["ManagedExpression"]]
     type: RadioField
 
-    def get_conditional_field_htmls(self) -> list[dict[str, dict[str, Markup]]]:
-        html = []
+    def get_managed_expression_radio_conditional_items(self) -> list[dict[str, dict[str, Markup]]]:
+        items = []
         for _managed_expression in self._managed_expressions:
-            html.append({"conditional": {"html": _managed_expression.render_conditional_fields(self)}})
-        return html
+            # format the radio items for `govuk-frontend-wtf` macro syntax
+            items.append({"conditional": {"html": _managed_expression.concatenate_all_wtf_fields_html(self)}})
+        return items
 
     def validate(self, extra_validators=None):  # type: ignore[no-untyped-def]
         for _managed_expression in self._managed_expressions:
--- a/app/common/expressions/managed.py
+++ b/app/common/expressions/managed.py
@@ -99,7 +99,7 @@ class ManagedExpression(BaseModel, SafeQidMixin):
         ...
 
     @classmethod
-    def render_conditional_fields(cls, form: "_ManagedExpressionForm") -> Markup:
+    def concatenate_all_wtf_fields_html(cls, form: "_ManagedExpressionForm") -> Markup:
         """
         A hook used by `build_managed_expression_form` to support conditionally-revealed the fields that a user needs
         to complete when they select this managed expression type from the radio list of available managed expressions.
--- a/app/developers/templates/developers/_render_managed_expression_form.html
+++ b/app/developers/templates/developers/_render_managed_expression_form.html
@@ -11,7 +11,7 @@
           "isPageHeading": False
           },
       },
-      "items": form.get_conditional_field_htmls()
+      "items": form.get_managed_expression_radio_conditional_items()
       })
     }}
 


app/common/expressions/managed.py
```diff
@@ -116,15 +302,63 @@ def statement(self) -> str:
...
+        )
+
+    @staticmethod
+    def form_data_from_expression(expression: "Expression") -> dict[str, Any]:
+        return {
+            "between_bottom_of_range": expression.context["minimum_value"],
+            "between_bottom_inclusive": expression.context["minimum_inclusive"],
+            "between_top_of_range": expression.context["maximum_value"],
+            "between_top_inclusive": expression.context["maximum_inclusive"],
+        }
```
sfount, line 356: I wonder if because the form is all already being built dynamically if this mapping could be rolled into `get_form_fields()` which would optionally accept an expression (if it already exists) and set the `default=` property if so

The one advantage to that is it would stop the string version of the form keys i.e "between_bottom_of_range" needing to be re-mapped and line up between where its defined three methods above and here

I'm sure there are some cons of not being able to get this mapping separately so will just note it down as an idea - all of the other methods _I think_ would break at runtime if they were trying to access keys on the form that didn't exist vs. this one might quite silently not work

app/developers/forms.py
```diff
@@ -49,5 +51,13 @@ class ConditionSelectQuestionForm(FlaskForm):
     )
     submit = SubmitField("Continue", widget=GovSubmitInput())
 
-    def add_question_options(self, questions: list["Question"]) -> None:
-        self.question.choices = [(question.id, f"{question.text} ({question.name})") for question in questions]
+    def __init__(self, *args, question: "Question", **kwargs):  # type: ignore[no-untyped-def]
+        super().__init__(*args, **kwargs)
+
+        supported_form_questions = [
```
sfount, line 57: It looks to me like this is what would actually consume the (nicely test covered) `get_supported_form_questions` but it seems to re-implement that logic here. (I may have missed something here!)

app/developers/templates/developers/add_question_condition_select_condition_type.html
```diff
@@ -0,0 +1,41 @@
...
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">Add a condition</h1>
+
+      {% include "developers/_add_question_condition_context.html" %}
+
+      {% if form is not none %}
+        {{ render_managed_expression_form(form, label_html="Only show the question if the answer <span class='govuk-visually-hidden'>to “" ~ depends_on_question.text ~ "”</span> is") }}
```
sfount, line 35: 🤏 preference for HTML to not be inlined and to be set in a separate block so that text editors/ IDEs treat them appropriately 
```suggestion
        {% set label_html %}
          <span>Only show the question if the answer <span class='govuk-visually-hidden'>to “{{ depends_on_question.text }}”</span> is</span>
        {% endset %}
        {{ render_managed_expression_form(form, label_html=label_html) }}
```

app/common/expressions/managed.py
sfount: (I don't think this should be changed right now unless you v. strongly agree for some reason)

app/common/expressions/forms.py
```diff
@@ -1,143 +1,95 @@
...
-                ]
-
-        # fixme: IDE realises this is a FlaskForm and bool but mypy is calling it "Any" on pre-commit
-        return super().validate(extra_validators=extra_validators)  # type: ignore
+class _ManagedExpressionForm(FlaskForm):
+    _question_data_type: QuestionDataType
+    _managed_expressions: list[type["ManagedExpression"]]
+    type: RadioField
+
+    def get_conditional_field_htmls(self) -> list[dict[str, dict[str, Markup]]]:
```
sfount, line 24: I'm not sure if this will be the same for all devs but I think it would be helpful for me to think about these methods in their relation to `govuk-frontend-wtf` rather than some of the more generic language used around rendering (wtforms is still the one responsible for rendering any of the form elements, and `govuk-frontend-wtf` macros are still piecing that all together, I guess this is tying them together).

I don't have a concrete example here but could the names of some of these methods make it clear that this is passing wtform fields into the macro syntax for govuk-frontend-wtf

Something like:

```python
def get_managed_expression_radio_items(self):
    items = []
    for _managed_expression in self._managed_expressions:
        # format the radio items for `govuk-frontend-wtf` macro syntax
        items.append({ "conditional": { "html": _managed_expression.concatenate_all_wtf_fields(self)}})
```

sfount: After trying to write this I don't think it's much if any better but I'm going to leave it here in case any of this thinking resonates - if it doesn't lets just ignore it and move on. Happy to chat it through

sfount (approved):
Bringing together logic split across many templates and modules into one place feels like a huge win for the developer experience that we'll likely be in for the next few months - really impressive balance between making this feel nice to set up but preserving how it works in different contexts 

I think there's one place we might not be consuming a well tested helper that we might want to fixup here but otherwise I think this can go in as-is. I've left a few other comments as thoughts but I don't have strong conviction in any of them

app/common/expressions/managed.py
```diff
@@ -1,20 +1,36 @@
...
     from app.common.data.models import Expression, Question
+    from app.common.expressions.forms import _ManagedExpressionForm
 
 
 class ManagedExpression(BaseModel, SafeQidMixin):
+    # Defining this as a ClassVar allows direct access from the class and excludes it from pydantic instance
+    # Used in the underproved+overcooked demo that lets an expression definition "magically" inject itself into
+    # the "add condition/validation" pages.
+    # This is maybe one of those fun things to show, but might be too "clever" for its own good (ie hard to trace,
+    # understand, extend, vary, etc).
```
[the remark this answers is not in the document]
sfount: Already cleaned up!

app/common/expressions/forms.py
samuelhwilliams: Cool, done 👍

sfount (approved):

PR #359, opened 2025-06-23
FSPT-619: delete validation rules

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-619

Following the deletion pattern laid out for collections, sections, forms, and questions, let's also make it possible to delete validation rules from the "edit validation" page.

We may be able to optimise this (eg remove the warning step) in the future as the decision is very lightweight and should be easy to reverse, but this feels fine for now. Design should be able to tackle this as part of the domain work.

### DIFF
--- a/app/developers/routes.py
+++ b/app/developers/routes.py
@@ -26,6 +26,7 @@ from app.common.data.interfaces.collections import (
     move_question_up,
     move_section_down,
     move_section_up,
+    remove_question_expression,
     update_collection,
     update_form,
     update_question,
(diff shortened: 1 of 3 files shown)


app/developers/routes.py
```diff
@@ -705,8 +706,27 @@ def edit_question_validation(grant_id: UUID, question_id: UUID, expression_id: U
     question = get_question_by_id(question_id)
     db_expression = get_expression(expression_id)
 
+    confirm_deletion_form = ConfirmDeletionForm()
+    if (
+        "delete" in request.args
+        and confirm_deletion_form.validate_on_submit()
+        and confirm_deletion_form.confirm_deletion.data
+    ):
```
srh-sloan, line 714: Feels like this route needs a unit test, or are we still in the realms of developer journey so not writing routes tests, just interface tests?

srh-sloan (approved):

PR #367, opened 2025-06-24
FSPT-411 E2E tests against AWS envs

srh-sloan:
Updates the `authenticated_browser_sso` fixtures to login differently depending on the target environment:
- Local: uses the existing stub SSO server login route
- Dev/Test: Creates a session cookie manually and injects it into the Page context, so the browser redirects you without needing to login.

Also updated the README with how to update the e2e tests to work locally if you have changed your config to use real SSO locally.

Note: The e2e tests currently fail in the deployed environments due to the notify api key we are using being set to 'Team' only. Ongoing discussion on how to fix this.

Latest run against test, showing the `create_grant` test working (this proves the fake cookie login route works): https://github.com/communitiesuk/funding-service/actions/runs/15852249305/job/44688473483

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,5 @@
 SECRET_KEY=pretend_secret_key
+SSO_USER_ID=<Get this from the id column of your local users table, for a user who has the correct permissions for the e2e tests>
 
 DATABASE_HOST=localhost
 DATABASE_PORT=5432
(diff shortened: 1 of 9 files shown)


samuelhwilliams (approved):
Nice, this looks good 👏  needs a rebase and the small typing tweaks would be good to get in so we're not lying to ourselves. Test feels very nice and runs against dev successfully 👍  https://github.com/communitiesuk/funding-service/actions/runs/15871336091

PR #371, opened 2025-06-25
FSPT-564 Combine add conditions and validations form 

sfount:
It's likely that for a given question type we'll be sharing some if not
all managed expressions (simple ways of comparing that data) between
conditions and validations.

This combines the two forms for adding an integer expression which were
put together at different times into one - this should mean devs can
specify one clear interface for managed expressions for each question
type and that will consistently apply to conditions and validations.

A follow up PR will then add the "edit" and "delete" condition functionality to
bring it in line with validations.


samuelhwilliams (approved):
This PR could also delete `_number_condition.html`; but feel free to clean up in another PR since I'm approving to unblock.

Code looks good, nice tidy up, thank you & mild sorry for creating the duplication.

samuelhwilliams (approved):

PR #372, opened 2025-06-25
FSPT-564 Manage and delete conditions

sfount:
Uses the existing "add condition" template and a similar pattern to validations to allow the user to update and delete a condition for a given question.

Some differences to watch out for:
- as the "add condition" flow involves first selecting which question you want to depend on its important that the back link when managing takes you back to the edit question or add condition flow depending on what you were doing
- the condition that you're updating continues to be based on the question type that you depend on not the question that you're updating the condition for - we might want to put in some syntax to make this clear throughout

Out of scoped in this PR:
- this PR doesn't add the proposed unique constraint for having a single managed condition per dependent question, that should follow

Video showing listing conditions, adding a condition, managing a condition and deleting a condition brought together

https://github.com/user-attachments/assets/176ef2f9-3b47-4a52-9dd4-3bef3d8d400a

Note it looks like conditions and validations have been moved outside of the "edit question" form which makes sense as those operations currently independently update the question - this should be carefully considered when looking at combining the "add" and "edit" question pages as it does currently imply that they're separate from a question where when they were in-line its clear that its all the same question you're updating

Depends on https://github.com/communitiesuk/funding-service/pull/371


app/developers/templates/developers/manage_question_condition_select_condition_type.html
```diff
@@ -0,0 +1,72 @@
...
+{% from "developers/_number_expression.html" import render_number_expression %}
+
+{% block pageTitle %}
+  Add a condition - {{ question.name }} - MHCLG Funding Service
+{% endblock pageTitle %}
+
+{% set active_item_identifier = "developers" %}
+
+{% set is_editing = expression %}
+{% set heading = "Manage condition" if is_editing else "Add a condition" %}
```
samuelhwilliams, line 18: The language for validations is `Edit validation` - happy to use Manage but if you can make it consistent one way or the other please 🙏

sfount: Nope I think I just called it this because it was the same as the template/ method names we were using? 

Edit is all good I'll clean that up

app/developers/routes.py
```diff
@@ -654,6 +656,64 @@ def add_question_condition(grant_id: UUID, question_id: UUID, depends_on_questio
...
 
+@developers_blueprint.route(
+    "/grants/<uuid:grant_id>/collections/questions/<uuid:question_id>/condition/<uuid:expression_id>",
+    methods=["GET", "POST"],
+)
+@is_platform_admin
+@auto_commit_after_request
+def edit_question_condition(grant_id: UUID, question_id: UUID, expression_id: UUID) -> ResponseReturnValue:
+    question = get_question_by_id(question_id)
+    expression = next(expression for expression in question.conditions if expression.id == expression_id)
```
samuelhwilliams, line 667: Feels like this might be nice in a small 🤏  - either an interface for get_expression or move this filtering to the Question model and have Question.get_expression()? :think]

sfount: Yeah agree @samuelhwilliams - I'll do whichever we feel is best as a follow up

One thing I'd be keen on getting feeling sensible here is making sure that the expression id thats been passed in reasonably belongs to the question (which is why I didn't get it directly from the DB) - we could always just make most accessor methods have to provide the "context" you're getting them for (i.e the grant, or question in this case) but make it a get_expression method on the question model feels quite nice for now

PR #376, opened 2025-06-25
BAU Remove the empty grant home page

sfount:
Remove the empty grant home page until we have a use for it - it currently makes the build look broken where its actually functional and should encourage people to critisise about it as a functional piece of software.

As soon as we have information to put on a dashboard/ homepage we can put this straight back in, i.e
- overview of your currently open rounds? 
- stats on your open applications/ assessments?
- number of reports on funding progress you have left to sign off?


samuelhwilliams:
Agree with this change, the grant homepage/dashboard does look broken right now. If we can run this past design so the prototype is updated until we fix this properly, that'd be grand 🙇

PR #378, opened 2025-06-25
BAU Tidy up the grants team member page

sfount:
Happy for this to be closed or reverted if its coming in future tickets.

The primary change here is to remove the "Contact us" prompt if the user already has permissions to add a new user - this should only need to be shown to users that don't have permissions to do it themselves (the page currently limits this to users with administrative permissions).

When thats removed the page looks empty and broken without any users - bring in the empty no users disclaimers from the prototype to avoid this.

Misc.
- moves xl headings to l heading as we've agreed we'll use those consistently now
- use empty lists being falsey for what feels like a more readable template to me
- the SQLAlchemy model for Grant will always populate an empty list of users, no need to guard against it being missed for readability

Grant team page if there are no users and you are a platform admin (there should be no reason you're seeing no users without being a platform admin) - you don't get the contact us prompt:

Screenshot 2025-06-25 at 23 00 57

Grant team page if you're a platform admin and there is a user - you don't get the the contact us prompt:

Screenshot 2025-06-25 at 23 00 46

Grant team page if you are not a platform admin - you get the contact us prompt:

Screenshot 2025-06-25 at 23 00 36

### DIFF
--- a/app/deliver_grant_funding/templates/deliver_grant_funding/grant_team/grant_user_add.html
+++ b/app/deliver_grant_funding/templates/deliver_grant_funding/grant_team/grant_user_add.html
@@ -17,8 +17,8 @@
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <h1 class="govuk-heading-xl">
-        <span class="govuk-caption-xl">Add grant team member</span>
+      <h1 class="govuk-heading-l">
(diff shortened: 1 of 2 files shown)


samuelhwilliams (commented):
Yeah this all feels sensible - consideration for some 🤏 test coverage possibly.

Please flag to @etheya so he can update the prototype as needed, and so that he is aware of the need to have different views of the page when building out prototypes in the future.

PR #381, opened 2025-06-26
FSPT-515 Invitations Table

srh-sloan:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-515

## 📝 Description
We currently create users directly when grant team members are added and when users try to authenticate via magic links, but this could lead to a lot of bad data in the user table (users never logging in, incorrect email addresses, etc.)

This adds an 'Invitations' table to the database to act as a kind of provisional user table for Deliver Grant Funding users authenticating with SSO, with the invitation being 'claimed' when a user signs in for the first time and a User is created in the database, and also stop the magic links journey from creating users before a magic link has been accepted.

See "Additional notes" below for some more context.

No vis in this one

## 🧪 Testing
Difficult to test as this is a non-destructive change that mainly adds the new table and a couple of interfaces, but interfaces are tested with integration tests and other related tables should not currently be affected and behaviour should continue as normal. 

### Review Readiness

### Performance and security

### Testing
~- [ ] End-to-end tests have been updated (if applicable)~

## 📚 Additional Notes

This is the first of what we expect will be 3 PRs to cover this change:

1. Create the Invitation table and add relationships 👈  (this PR) 
2. Update the magic link functionality to not create users until they login ([very draft WIP PR](https://github.com/communitiesuk/funding-service/pull/425))
3. Update the SSO functionality to use invitations

Also adds a `last_logged_in_at_utc` column to the User table so that we can start to track when users log in (via sso or magic link), which is set when a user successfully logs in with `login(user)` by listening for the flask_login `user_logged_in` signal) rather than having to rely on whether a User's `name` field is populated in the database to indicate they've signed in with SSO.

One slight change to the ticket detail - the more we worked through the models the less sense it made for Magic Links to share this Invitation mechanism with SSO users, as magic links are already a kind of (very time limited) invitation, and syncing up a 15m expiry magic link with a 7 day expiry invitation would lead to lots of very frustrating and very avoidable edge cases. With magic links also being used for Access Grant Funding, they likely won't use the 'Invitation' functionality at all (at least for now when there's no invitation concept for Access Grant Funding), so instead we'll just wait to create the User until they've claimed the magic link and signed in - this migration preps for this change in flow.

### DIFF
--- a/app/common/data/interfaces/user.py
+++ b/app/common/data/interfaces/user.py
@@ -34,6 +34,11 @@ def get_user_by_azure_ad_subject_id(azure_ad_subject_id: str) -> User | None:
     return db.session.execute(select(User).where(User.azure_ad_subject_id == azure_ad_subject_id)).scalar_one_or_none()
 
 
+def set_user_last_logged_in_at_utc(user: User) -> User:
+    user.last_logged_in_at_utc = func.current_timestamp()
+    return user
+
+
 def upsert_user_by_email(
     email_address: str,
     *,
--- a/app/common/data/interfaces/user.py
+++ b/app/common/data/interfaces/user.py
@@ -187,3 +187,15 @@ def create_invitation(
     db.session.add(invitation)
     db.session.flush()
     return invitation
+
+
+def get_invitation(invitation_id: uuid.UUID) -> Invitation | None:
+    return db.session.get(Invitation, invitation_id)
+
+
+def claim_invitation(invitation: Invitation, user: User) -> Invitation:
+    invitation.claimed_at_utc = func.now()
+    invitation.user = user
+    db.session.add(invitation)
+    db.session.flush()
+    return invitation
--- a/app/common/data/interfaces/user.py
+++ b/app/common/data/interfaces/user.py
@@ -1,14 +1,16 @@
+import datetime
 import uuid
 from typing import cast
 
 from flask_login import current_user
-from sqlalchemy import and_
+from sqlalchemy import and_, func
 from sqlalchemy.dialects.postgresql import insert as postgresql_upsert
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import delete, select
 
 from app.common.data.interfaces.exceptions import InvalidUserRoleError
-from app.common.data.models_user import User, UserRole
+from app.common.data.models import Grant, Organisation
+from app.common.data.models_user import Invitation, User, UserRole
 from app.common.data.types import RoleEnum
 from app.extensions import db
 from app.types import NOT_PROVIDED, TNotProvided
@@ -167,3 +169,21 @@ def remove_all_roles_from_user(user: User) -> None:
     db.session.execute(statement)
     db.session.flush()
     db.session.expire(user)
+
+
+def create_invitation(
+    email: str,
+    organisation: Organisation | None = None,
+    grant: Grant | None = None,
+    role: RoleEnum | None = None,
+) -> Invitation:
+    invitation = Invitation(
+        email=email,
+        organisation_id=organisation.id if organisation else None,
+        grant_id=grant.id if grant else None,
+        role=role,
+        expires_at_utc=func.now() + datetime.timedelta(days=7),
+    )
+    db.session.add(invitation)
+    db.session.flush()
+    return invitation
--- a/app/common/data/models_user.py
+++ b/app/common/data/models_user.py
@@ -1,6 +1,9 @@
+import secrets
 import uuid
+from datetime import datetime
 from typing import TYPE_CHECKING
 
+from pytz import utc
 from sqlalchemy import CheckConstraint, ForeignKey, Index, UniqueConstraint
 from sqlalchemy import Enum as SqlEnum
 from sqlalchemy.orm import Mapped, mapped_column, relationship
@@ -9,7 +12,7 @@ from app.common.data.base import BaseModel, CIStr
 from app.common.data.types import RoleEnum
 
 if TYPE_CHECKING:
-    from app.common.data.models import Grant, MagicLink, Organisation, Submission
+    from app.common.data.models import Grant, Organisation, Submission
 
 
 class User(BaseModel):
@@ -20,8 +23,13 @@ class User(BaseModel):
     azure_ad_subject_id: Mapped[str] = mapped_column(nullable=True, unique=True)
 
     magic_links: Mapped[list["MagicLink"]] = relationship("MagicLink", back_populates="user")
+    invitations: Mapped[list["Invitation"]] = relationship(
+        "Invitation", back_populates="user", cascade="all, delete-orphan"
+    )
     roles: Mapped[list["UserRole"]] = relationship("UserRole", back_populates="user", cascade="all, delete-orphan")
     submissions: Mapped[list["Submission"]] = relationship("Submission", back_populates="created_by")
+
+    last_logged_in_at_utc: Mapped[datetime | None] = mapped_column(nullable=True)
 
     # START: Flask-Login attributes
     # These ideally might be provided by UserMixin, except that breaks our type hinting when using this class in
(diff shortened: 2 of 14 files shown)


app/common/data/interfaces/user.py
```diff
@@ -167,3 +168,33 @@ def remove_all_roles_from_user(user: User) -> None:
...
     db.session.flush()
     db.session.expire(user)
+
+
+def create_invitation(
+    email: str,
+    organisation_id: uuid.UUID | None = None,
+    grant_id: uuid.UUID | None = None,
+    role: RoleEnum | None = None,
+) -> Invitation:
```
samuelhwilliams, line 178: Do we think we'll have the actual organisation/grant objects here? If so, can we pass those in? By using the full objects we have a better chance of typechecking catching us getting the order wrong. We could also switch these to kwarg-only for extra protection.

app/common/data/models_user.py
```diff
@@ -20,9 +23,15 @@ class User(BaseModel):
...
 
     magic_links: Mapped[list["MagicLink"]] = relationship("MagicLink", back_populates="user")
+    invitations: Mapped[list["Invitation"]] = relationship(
+        "Invitation", back_populates="user", cascade="all, delete-orphan"
+    )
     roles: Mapped[list["UserRole"]] = relationship("UserRole", back_populates="user", cascade="all, delete-orphan")
     submissions: Mapped[list["Submission"]] = relationship("Submission", back_populates="created_by")
 
+    # TODO make this nullable=False once all users have a last_login_utc set
+    last_login_utc: Mapped[datetime | None] = mapped_column(nullable=True)
```
samuelhwilliams, line 33:
```suggestion
    last_logged_in_at: Mapped[datetime | None] = mapped_column(nullable=True)
```

optional nit: slight tweak to the column name here just to match the `ed` suffix (login vs logged_in, we have created_at elsewhere)

app/common/data/models_user.py
```diff
@@ -20,9 +23,15 @@ class User(BaseModel):
     azure_ad_subject_id: Mapped[str] = mapped_column(nullable=True, unique=True)
 
     magic_links: Mapped[list["MagicLink"]] = relationship("MagicLink", back_populates="user")
+    invitations: Mapped[list["Invitation"]] = relationship(
+        "Invitation", back_populates="user", cascade="all, delete-orphan"
+    )
     roles: Mapped[list["UserRole"]] = relationship("UserRole", back_populates="user", cascade="all, delete-orphan")
     submissions: Mapped[list["Submission"]] = relationship("Submission", back_populates="created_by")
 
+    # TODO make this nullable=False once all users have a last_login_utc set
```
samuelhwilliams, line 32: Do we just suck this up now and set it to func.now() for existing users as part of the migration? Feel free to ignore this if a future PR is already doing something, but feels like we could just take that rather than leave a bit of cleanup for later.

app/common/data/models_user.py
```diff
@@ -82,3 +91,54 @@ class UserRole(BaseModel):
...
+    expires_at_utc: Mapped[datetime]
+    claimed_at_utc: Mapped[datetime | None]
+
+    user: Mapped[User] = relationship("User", back_populates="magic_links")
+
+    __table_args__ = (Index(None, code, unique=True, postgresql_where="claimed_at_utc IS NOT NULL"),)
+
+    @property
+    def usable(self) -> bool:
+        return self.claimed_at_utc is None and self.expires_at_utc > datetime.now(utc).replace(tzinfo=None)
```
samuelhwilliams, line 113: this is maybe the one that could be a hybrid property 👀

tests/models.py
```diff
@@ -224,3 +224,37 @@ class Meta:
...
+            organisation=factory.SubFactory(_OrganisationFactory),
+        )
+        has_grant = factory.Trait(
+            grant_id=factory.LazyAttribute(lambda o: o.grant.id),
+            grant=factory.SubFactory(_GrantFactory),
+        )
+        is_claimed = factory.Trait(
+            claimed_at_utc=factory.LazyFunction(lambda: datetime.datetime.now()),
+            user=factory.SubFactory(_UserFactory),
+            user_id=factory.LazyAttribute(lambda o: o.user.id if o.user else None),
```
samuelhwilliams, line 259: These traits feel like really nice ways to encapsulate the logic and assist with setup - good idea 🙇

app/common/data/models_user.py
MarcUsher: Good point, I'll tweak to `last_logged_in_at_utc` so we've got the explicit UTC as well

app/common/data/models_user.py
samuelhwilliams: ok yep nice nice

app/common/data/models_user.py
MarcUsher: Just updating the thread for posterity - adding `onupdate` to the `last_logged_in_at_utc` column doesn't actually make sense, you don't want it to update on every random update to the User, and without explicitly passing it to our upserts or setting a `server_default` on the column (which is wrong and defeats the point) then it doesn't work nicely with postgres. Instead am switching to explicitly passing to our user creation upserts so that it will be set each time on login.

There will be a slight period until this whole ticket is complete where we create users with a `last_logged_in_at_utc` value before they've logged in - as it's only us logging in and adding dummy users this should be OK, and this behaviour will be updated very shortly with invitations for SSO users and magic links only creating a user when they claim the magic link.

MarcUsher: Updating for posterity again - the above was nonsense because it meant we were adding `last_logged_in_at_utc` on upserts before a user had technically logged in and with the possibility that they wouldn't be successfully logged in, which was just wrong.

Instead the latest commit switches this to use flask_login's `user_logged_in` signal, which is fired on a successful `login_user(user)`, to set this field for the user logging in.

tests/integration/common/data/interfaces/test_user.py
```diff
@@ -326,3 +335,40 @@ def test_remove_all_roles_from_user(self, db_session, factories) -> None:
...
+
+class TestInvitations:
+    @pytest.mark.freeze_time("2023-10-01 12:00:00")
+    def test_create_invitation(self, db_session, factories):
+        invitation = interfaces.user.create_invitation(email="test@email.com", role=RoleEnum.MEMBER)
+        invite_from_db = db_session.get(Invitation, invitation.id)
+        assert invite_from_db is not None
+        assert invite_from_db.email == "test@email.com"
+        assert invite_from_db.role == RoleEnum.MEMBER
+        assert invite_from_db.expires_at_utc == datetime.strptime("2023-10-08 12:00:00", "%Y-%m-%d %H:%M:%S")
```
samuelhwilliams, line 348:
```suggestion
        assert invite_from_db.expires_at_utc == datetime(2023, 10, 8, 12, 0, 0)
```

nit: not sure the extra parsing from a string is useful here

tests/integration/conftest.py
```diff
@@ -257,14 +257,15 @@ def _track_notification(*args, **kwargs):  # type: ignore
...
-    anonymous_client: FlaskClient, factories: _Factories, request: FixtureRequest
+    anonymous_client: FlaskClient, factories: _Factories, request: FixtureRequest, db_session: Session
 ) -> Generator[FlaskClient, None, None]:
     email_mark = request.node.get_closest_marker("authenticate_as")
     email = email_mark.args[0] if email_mark else "test@communities.gov.uk"
 
     user = factories.user.create(email=email)
 
     login_user(user)
+    db_session.commit()
```
samuelhwilliams, line 268: nit: A small comment on why we need this might be nice

tests/integration/deliver_grant_funding/test_routes.py
```diff
@@ -39,7 +39,7 @@ def test_list_grants_as_admin(
     for expected in expected_headers:
         assert expected in header_texts, f"Header '{expected}' not found in table"
     assert soup.h1.text == "Grants"
-    assert len(queries) == 3  # 1) select grant, 2) rollback, 3) savepoint
+    assert len(queries) == 3  # 1) select user, 2) select user_role, 3) select grants
```
samuelhwilliams, line 42: 🙇  ty

samuelhwilliams (approved):
Looks nice - a few nitty things, but I think the bones are good 👍

tests/models.py
```diff
@@ -224,3 +225,37 @@ class Meta:
...
+    id = factory.LazyFunction(uuid4)
+    email = factory.Faker("email")
+    user_id = None
+    user = None
+    organisation_id = None
+    organisation = None
+    grant_id = None
+    grant = None
+    role = None
+    expires_at_utc = factory.LazyFunction(lambda: datetime.datetime.now() + datetime.timedelta(days=7))
```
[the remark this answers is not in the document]
MarcUsher: Tried adding this and it blows up because `created_at_utc` is a field set on our base model but not a field explicitly created in the factory - so it's either a case of adding it as an explicit field just to base this one off, or we leave it as `datetime.datetime.now() + ...` - for now I've left as the latter as we're not explicitly bringing `created_at_utc` anywhere else in the factories and felt a little unnecessary to do that here unless we explicitly need it in some tests?

MarcUsher:
> Looks nice - a few nitty things, but I think the bones are good 👍

Thanks @samuelhwilliams - all good notes, have just pushed updates to the PR addressing the comments 🙏

PR #382, opened 2025-06-26
Refactor dynamic question form creation and expressions context

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-592

---

The current dynamic question forms that we use to render questions on
the page for a submission only supports a single field, with the
hard-coded name `question`. We know that we will eventually want to
render multiple questions on one page, and that these will need to be on
a single wtform, so the `question` name won't work.

This refactor attaches questions to the form class based on their
question id. By reusing the ("mangled") question ID that feeds the
expression evaluation engine, we can more neatly map data that the user
is submitting in the form to data that the expressions will evaluate
against.

We probably should rename this mangling function and abstract it/provide
a cleaner interface for it soon if we're making this touch a few more
places.

With the form field names matching the question id format in
expressions, we can then pull data straight from the form and pass it to
the evaluation engine without any further mutation.

With each form field scoped to its question ID, the dynamic form should
work automatically when we support adding multiple questions to it, and
those fields' data will flow into the evaluation engine easily and
without any further changes.

We do need helpers on the form to set/grab the field for a specific
question, since we can no longer do something hardcoded like
`form.question`, so this adds getters/setters for form fields based on
question instances.

We also need a helper on the form to render questions in our jinja
templates, enter `render_question(question)`. This can also be passed
`params` for GOV.UK Frontend Jinja macros, as you'd normally pass with
direct macro calls.

---

This tweaks the expression evaluation interface to take a specific
dict-like object instead of an arbitrary dict.

This new 'ExpressionContext' object wraps three levels of dicts:
1. form data
2. submission data
3. expression data

The ExpressionContext is immutable and the wrapped dicts are immutable
too. This means that we can pass them around and use them across
multiple validators without worrying about copying or merging dicts,
which should save a lot of memory churn and - with some familiarity -
make it easier to track/trust that the right data is in the right place.

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -1,16 +1,23 @@
 import uuid
 from datetime import datetime
 from itertools import chain
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Any, Optional, Type
 from uuid import UUID
 
 from pydantic import RootModel, TypeAdapter
+from sqlalchemy.util import immutabledict
 
 from app.common.collections.forms import DynamicQuestionForm
 from app.common.data import interfaces
 from app.common.data.interfaces.collections import get_submission
 from app.common.data.models_user import User
-from app.common.data.types import QuestionDataType, SubmissionEventKey, SubmissionModeEnum, SubmissionStatusEnum
+from app.common.data.types import (
+    QuestionDataType,
+    SubmissionEventKey,
+    SubmissionModeEnum,
+    SubmissionStatusEnum,
+)
+from app.common.expressions import mangle_question_id_for_context
 
 if TYPE_CHECKING:
     from app.common.data.models import Form, Grant, Question, Section, Submission
@@ -103,6 +110,10 @@ class SubmissionHelper:
     @property
     def collection_id(self) -> UUID:
         return self.collection.id
+
+    @property
+    def expression_context(self) -> immutabledict[str, Any]:
+        return immutabledict({mangle_question_id_for_context(uuid.UUID(k)): v for k, v in self.submission.data.items()})
 
     def get_section(self, section_id: uuid.UUID) -> "Section":
         try:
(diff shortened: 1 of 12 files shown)


app/common/helpers/collections.py
```diff
@@ -59,6 +66,14 @@ def name(self) -> str:
     def reference(self) -> str:
         return self.submission.reference
 
+    @property
+    def expression_context(self) -> ExpressionContext:
+        return ExpressionContext(
+            from_submission=immutabledict(
+                {mangle_question_id_for_context(uuid.UUID(k)): v for k, v in self.submission.data.items()}
```
sfount, line 73: currently this fetches straight out of the submission data, do we want to start it off with the contexts being guarded by question types? I'm still thinking through some of the implications (in a comment above) but we could do that tidily for the submission helper without deeping performance too hard

```suggestion
                {mangle_question_id_for_context(uuid.UUID(key)): self.get_answer_for_question(key).root for key in self.submission.data.keys() }
```

app/developers/routes.py
```diff
@@ -898,11 +898,11 @@ def collection_confirmation(submission_id: UUID) -> ResponseReturnValue:
 def ask_a_question(submission_id: UUID, question_id: UUID) -> ResponseReturnValue:
     submission_helper = SubmissionHelper.load(submission_id)
     question = submission_helper.get_question(question_id)
-    answer = submission_helper.get_answer_for_question(question.id)
 
     # this method should work as long as data types are a single field and may
     # need to be revised if we have compound data types
-    form = build_question_form(question)(question=answer.root if answer else None)
+    submission_context = submission_helper.expression_context
+    form = build_question_form(question)(data=submission_context)
```
sfount, line 905: This probably wants a `todo:` about multi-value question types needing to think through this step - I'm guessing the `data=` might need to be done by the form based on the questions that are passed in so it can split up the default values if it needs to 

(although this should work for all of our primitive question types right now so lets not do ti now)

sfount (approved):
This feels like a really good foundation to build on without trying to anticipate too many of the complexities that we don't know enough about - which is a hard balance!

PR #384, opened 2025-06-26
FSPT-566 Visually differentiate conditional questions

sfount:
This commit adds a tag to visually differentiate questions on the manage form page that have conditions and might not be shown to the user.

Adding this isn't to set precedent or establish this as the standard pattern but it gets in the strucutre for us to show this information appropriately given user feedback and any more design passes.

We're currently lazy loading all questions on the manage form page which hasn't been a problem for performance but now that this feature is also inspecting if they do or don't have expressions this is an expensive series of single queries as the number of questions and expressions grows.

Instead of lazy loading add an optional parameter to eagerly fetch all of the questions for a form alongisde all of those questions expressions
- this should be reasonably performant even with a larger set of questions.

Showing conditional questions option 1 (tag before the question): 

Screenshot 2025-06-26 at 19 24 23

Showing conditional questions option 2 (tag after the question - this is what this change request currently has):

Screenshot 2025-06-26 at 19 24 05


app/developers/templates/developers/manage_form.html
```diff
@@ -92,6 +101,7 @@ <h2 class="govuk-heading-m">Questions</h2>
 
       {% if db_form.questions %}
         <p class="govuk-body">This form has {% trans count=db_form.questions | length %}{{ count }} question{% pluralize %}{{ count }} questions{% endtrans %}.</p>
+        {# fixme: the styling for the border top and tag spacing look broken on tablet and mobile #}
```
samuelhwilliams, line 104: this attention to detail is really nice, keep it up (i know you will ofc) 🙏

samuelhwilliams (approved):

PR #385, opened 2025-06-26
FIX Appropriately manage order when deleting entities

sfount:
Note: all of the methods covered by this fix are in a "temporary" interface that doesn't have coverage and is intended to make things work for now. This commit doesn't intend to change that but it should stop us running into errors if removing and then adding again.

Deleting questions, forms or sections and then trying to add another currently raises a `DuplicateValueError` for their ordering constraints (i.e `uq_question_order_form`).

This is especially noticeable when you remove the first entity in the ordered list. When the entity is removed the order of all of the other entities remains the same (in this example if you have 0, 1, 2 and remove 0, the two remaining will stay at 1 and 2).

If you try and add a new entity now it will be assigned the order "2" as there are 2 other entities remaining and the order index starts at 0 - this will then conflict with the existing entity with the order 2.

If we want to lean on the ORM to manage the ordering using an ordered list we'll need to tell it that the entity has been removed which will trigger an appropriate update to the existing elements order and means any new entity then added will have a unique value based on the max of whats there.

Steps to reproduce:
- remove the first question on a form
- add a new question to the form

### DIFF
--- a/app/common/data/interfaces/temporary.py
+++ b/app/common/data/interfaces/temporary.py
@@ -11,11 +11,12 @@ The only place that should import from here is the `app.developers` package.
 
 from uuid import UUID
 
+from sqlalchemy import text
+
 from app.common.data.models import (
     Collection,
     Form,
     Grant,
-    Question,
     Section,
     Submission,
 )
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
is this related to https://github.com/communitiesuk/funding-service/pull/357/commits/6145c681b69735654d51d2f8fda510db6d71e3fb and needing to expire things with relationships?

sfount:
> is this related to [6145c68](https://github.com/communitiesuk/funding-service/commit/6145c681b69735654d51d2f8fda510db6d71e3fb) and needing to expire things with relationships?

Having a little look into it this morning I don't think these are related - being able to expire an ORM object you've loaded into memory will tell SQLAlchemy to refresh its "read" properties/ relationships if its used again later but from my very early understanding I don't think it impacts decisions it will make about updates. The statements previously built and executed an SQL query but I guess not acting on the ORM list object meant it didn't know to update. 

I'll try out a few options to make sure some combination of delete and expire doesn't achieve the same result but I suspect not

Probably should have been written into the commit but a few things of note:
- deferring the ordering unique constraint was needed similarly to moving the ordering around
- the reason for an explicit "delete" of the object from the session before removing it from the parent list is because the default ORM cascade rules are set to not remove orphans! (as that would cause unpredictable behaviour when re-ordering)

app/common/data/interfaces/temporary.py
```diff
@@ -51,19 +52,39 @@ def delete_collection(collection_id: UUID) -> None:
     db.session.flush()
 
 
-def delete_section(section_id: UUID) -> None:
-    section = db.session.query(Section).where(Section.id == section_id).one()
+def delete_section(section_id: UUID, collection_id: UUID) -> None:
```
samuelhwilliams, line 55: These should all just take the DB object `section` itself, that feels like maybe a nicer interface than passing a raw ID? I think we're likely to always have that object loaded already? Then we don't need to do a query to fetch the collection, we can just do `section.collection`

samuelhwilliams: (but they're just temp/dev interfaces you're right)

samuelhwilliams (approved):

samuelhwilliams (approved):
Lots of good sleuthing and documenting in commit messages 👍

PR #388, opened 2025-06-27
Run managed validation on question pages

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-592

## Description
This patch hooks up all of the validation expressions registered against a question to the 'ask a question' page form, so that when the users submits an answer, the validation runs and raises any errors as appropriate.

We only support managed validation at this stage; when we start adding support for custom expressions we'll need to extend the functionality here.

### DIFF
--- a/tests/e2e/developer_pages.py
+++ b/tests/e2e/developer_pages.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
 
+from typing import cast
+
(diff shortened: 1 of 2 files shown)


sfount (approved):
Change looks great! the video and coverage help bring all of this together

sfount (approved):

PR #389, opened 2025-06-27
FSPT-565 Calculate visibility based on question conditions

sfount:
Runs through question conditions when loading the visible questions for a form.

Note that for the time being this runs against all questions whenever the submission helper is used as getting all of the visible questions is how the helper gets the next and previous questions.

For that reason the default behaviour of the visibility check is to allow question conditions to not parse for missing information as it will be checking questions at the end of the form where the answers at the start may not be filled in.

When a question is routed to it should individually sense check that it should be visible - that call should not be allowed to not parse conditions.

When an answer is submitted it should do the same.

We should consider checking the dependency chain when a condition fails to parse for a missing answer to give us some confidence it links back to a question that can be answered or log for misconfiguration (something that would be completely missed by defaulting to hiding the question as it does now).

Note this change out of scopes uniformly serialising answers from submission or forms and directly accesses K:V pairs - this is something that should be tackled on its own.

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -197,7 +197,11 @@ class SubmissionHelper:
         try:
             return all(evaluate(condition, context) for condition in question.conditions)
         except UndefinedVariableInExpression:
-            return False  # fail open for now, checking individual question visibility while routing should error though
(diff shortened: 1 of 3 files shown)


samuelhwilliams (approved):
Really nice and straight forward 🙇  Well described and explained

PR #394, opened 2025-06-27
Delete grants at the end of e2e test runs

samuelhwilliams:
Note: best reviewed with whitespace changes ignored

We want our dev/test environments (especially test) to look+feel fairly legitimate, without a bunch of UUID-ified grants that slowly pile up from our e2e tests.

This patch adds a 'Delete this grant' link following the existing 'delete thing' patterns, and updates the e2e tests which create grants, to delete them at the end as well.

A rudimentary attempt is made to still tidy up if the e2e test fails part way through. There's a small argument for leaving it around in case you want to poke at the grant/collection/etc to see what went wrong, but I think this is more likely to be annoying that actually useful for debugging. The e2e test artifacts (eg screenshots) should be more helpful.

### DIFF
--- a/app/developers/routes.py
+++ b/app/developers/routes.py
@@ -72,11 +72,25 @@ def inject_variables() -> dict[str, Any]:
     return dict(show_watermark=True)
 
 
-@developers_blueprint.route("/grants/<uuid:grant_id>", methods=["GET"])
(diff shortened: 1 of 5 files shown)


MarcUsher (approved):
This is nice, agree it was getting a bit messy in the deployed envs now the e2e tests are running regularly. Approving with just a little question that's for education more than anything.

PR #395, opened 2025-06-28
Refactors to dynamic form generation/validation

samuelhwilliams:
## Move validation chain
Instead of generating the validators when the form class is dynamically
generated and the fields are attached, we now build them only on a call
to `validate`, and we do it within a custom implementation of the
`validate` call. This achieves a couple of things:

- The association between validators and the expression context is no
  longer as spread out across the code; the form context is attached in
  the same place that the validators are built and executed. This should
  make it easier to reason about correctness of the implementation.
- We can run the native WTForm field validation in a separate early
  pass, which makes it easier to reason about all of our custom
  validation rules having data already in the correct 'shape'.
- It should be easier to reason about cross-field/full-form validation
  when it runs in this context, instead of with validators attached
  directly to each field. Note that the implementation will still likely
  need some tweaking to work across multiple fields.

## Refactor base class for dynamic question forms
Move some core methods to the base class so that the form generation
function is cleaner, easier to read, and doesn't create duplicate method
definitions with every call.

### DIFF
--- a/app/common/collections/forms.py
+++ b/app/common/collections/forms.py
@@ -44,7 +44,7 @@ class DynamicQuestionForm(FlaskForm):
 
     def validate(self, extra_validators: Mapping[str, list[Any]] | None = None) -> Any:
         """
-        Run the form's validation chain. This works in two steps, currently intermingled:
+        Run the form's validation chain. This works in two steps:
         - WTForm's built-in field-level validation (eg for IntegerField, that data has been provided, and that it
           can be coerced to an integer value.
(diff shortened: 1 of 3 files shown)


sfount (approved):
Nice! This feels like a great manageable step rather than trying to leap to this stage - great work 

> The association between validators and the expression context is no
longer as spread out across the code; the form context is attached in
the same place that the validators are built and executed. This should
make it easier to reason about correctness of the implementation.

> We can run the native WTForm field validation in a separate early
pass, which makes it easier to reason about all of our custom
validation rules having data already in the correct 'shape'.

I can see how both of these are shaping up and it looks good 👍

PR #396, opened 2025-06-28
Clean up some of the question ID mangling

samuelhwilliams:
There are a few places across the codebase where we need to standardise the question ID into a format that can be used dynamically by Python:

- when attaching question fields to flask forms
- when evaluating expressions

The existing function to do this has spread out into a few corners of the codebase and started to feel a bit leaky. This makes a small attempt to clean that up by moving the transformation logic into a mixin, and adding that to our main two models that need to access the transformed value: Questions and (pydantic) managed expressions.

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -73,11 +73,11 @@ class SubmissionHelper:
     @property
     def expression_context(self) -> ExpressionContext:
         submission_data = {
-            question.safe_qid: answer
+            question.safe_qid: answer.model_dump()
             for section in self.submission.collection.sections
             for form in section.forms
             for question in form.questions
-            if (answer := self.submission.data.get(str(question.id))) is not None
+            if (answer := self.get_answer_for_question(question.id)) is not None
         }
         return ExpressionContext(from_submission=immutabledict(submission_data))
 


app/common/helpers/collections.py
```diff
@@ -73,11 +72,14 @@ def reference(self) -> str:
...
-            from_submission=immutabledict(
-                {mangle_question_id_for_context(uuid.UUID(k)): v for k, v in self.submission.data.items()}
-            )
-        )
+        submission_data = {
+            question.safe_qid: answer.model_dump()
+            for section in self.submission.collection.sections
+            for form in section.forms
+            for question in form.questions
+            if (answer := self.get_answer_for_question(question.id)) is not None
```
sfount, line 80: This little refactor on its own might be worth calling out as a separate change! Very nice, I think this sets us up for consistent context data and we'll just need to work out how its sensible for dynamic forms to be doing this processing too

app/common/helpers/collections.py
```diff
@@ -73,11 +72,14 @@ def reference(self) -> str:
 
     @property
     def expression_context(self) -> ExpressionContext:
-        return ExpressionContext(
-            from_submission=immutabledict(
-                {mangle_question_id_for_context(uuid.UUID(k)): v for k, v in self.submission.data.items()}
-            )
-        )
+        submission_data = {
+            question.safe_qid: answer.model_dump()
```
sfount, line 76: I think its mentioned in comments but this will work nicely well all the question types are primitives (root models) and will probably need to be looked at again if they contain multiple data fields

sfount (approved):
Great cleanup it feels very sensible for a question to be doing this, it makes me wonder if we'll ever need to work backwards from a safe_qid but I think at the moment the proposal is the question ids used by managed expressions are stored alongside its context

sfount (approved):

PR #397, opened 2025-06-28
Sync removed roles from enum

samuelhwilliams:
In https://github.com/communitiesuk/funding-service/pull/330 we removed a few roles from the role enum, but didn't generate a migration for them.

This generates that migration using the `alembic_postgresql_enum` extension, which plugs enum values into alembic for automatic sync detection. This should make it harder to miss in the future.

When dropping values, we need to first delete (or appropriately handle) any rows that use the old enum values.


MarcUsher (approved):
Ah nice catch thanks Sam! V handy to have the little library to catch these when doing migrations, and also a good note that you need to re-add any constraints that rely on that enum after it gets updated in the migration 🙏  I'll add a little update to the enums page on confluence where Sarah documented how we want to do them for this repo

PR #399, opened 2025-06-30
Fix SQLAlchemy query listing for Flask-DebugToolbar

samuelhwilliams:
This patch updates our copy of `record_queries` to store queries on the Flask global object in the same place as Flask-SQLAlchemy (https://github.com/pallets-eco/flask-sqlalchemy/blob/main/src/flask_sqlalchemy/record_queries.py#L90-L91). This implementation takes advantage of ("abuses") a few things in order to make the integration with Flask-DebugToolbar:

- Flask-DebugToolbar (FDT) only supports Flask-SQLAlchemy (FS) for now in theory, not explicitly Flask-SQLAlchemy-Lite (FSL).
- But we have FS installed as a dependency for our app, because Flask-Migrate requires it.
- FDT thinks that FS is installed as an extension because FSL uses the same extension namespace (`app.extensions['sqlalchemy']`).
- So if we configure query recording on, FDT will use FS's `get_recorded_queries` function to lookup executed queries.
- If our copy of the extension writes queries to the same namespace as FS then FDT will read and show our recorded queries.

This is all a little bit fragile and magical, but should work for now (and for a while) because FS and FDT haven't changed much about how this works in a while. Hopefully official support for these query recording can get into FSQLA and FDT though.

### DIFF
--- a/app/extensions/record_sqlalchemy_queries.py
+++ b/app/extensions/record_sqlalchemy_queries.py
@@ -34,6 +34,9 @@ NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 """
 
+# TODO: Remove this module if/when https://github.com/pallets-eco/flask-sqlalchemy-lite/pull/21 is merged and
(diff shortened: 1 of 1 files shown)


MarcUsher (approved):
Ah this is really great! Works nicely for me locally, this will be v v helpful 🙏

PR #400, opened 2025-06-30
Bypass extension for query context

samuelhwilliams:
All query contexts were showing as coming from the `record_sqlalchemy_queries` file, which isn't actually useful.

## Before
image

## After
image


MarcUsher (approved):
Ah that's v useful, I almost commented about this in the other PR but assumed it was a byproduct of how the patch was working that we had to live with 😅

PR #422, opened 2025-07-01
FSPT-596 block questions moving above/ below questions they depend on

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-596

## 📝 Description
For now this implements guards to stop two questions from being moved
above/ below other questions that they depend on.

We'll probably want to think through a method of generally checking the
integrity of the order of all questions in a form and any type of
dependency (condition, validation, personalised text).

For now the only managed option which can be configured that depends on
other questions are conditions so lets just start with those.

By being specific about question conditions its easy to provide which
questions are the problem specifically through the UI without coming up
with brand new patterns.

This proposes encapsulating the dependency failure as a generic
exception which should then be able to be passed through to the
template.

screenshot-jul-1

## 🧪 Testing
As this hooks into the "swap" operation its currently covered at the interface level, that will want to be considered again if doing any higher level integrity checks (the current behaviour relies on _starting_  in a healthy state and should then restrict you to continue in one)

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -36,6 +36,7 @@ from app.extensions import (
 )
 from app.monkeypatch import patch_sqlalchemy_lite_async
 from app.sentry import init_sentry
+from app.types import FlashMessageType
 
 if TYPE_CHECKING:
     from app.common.data.models_user import User
@@ -211,5 +212,6 @@ def create_app() -> Flask:
     # should make an intentional decision for when to be setting this
     app.config["SEND_FILE_MAX_AGE_DEFAULT"] = 3600
     app.add_template_global(AuthorisationHelper, "authorisation_helper")
+    app.add_template_global(FlashMessageType, "flash_message_type")
 
     return app
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -266,7 +266,7 @@ class DependencyOrderException(Exception, FlashableException):
 
 
 # todo: we might want something more generalisable that checks all order dependencies across a form
-#       but this gives us the specific result we want for for the UX for now
+#       but this gives us the specific result we want for the UX for now
 def check_question_order_dependency(question: Question, swap_question: Question) -> None:
     for condition in question.conditions:
         if condition.managed and condition.managed.question_id == swap_question.id:
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -244,13 +244,44 @@ def get_question_by_id(question_id: UUID) -> Question:
     return db.session.get_one(Question, question_id)
 
 
+class DependencyOrderException(Exception):
+    def __init__(self, question: Question, depends_on_question: Question):
+        super().__init__("You cannot move questions above answers they depend on.")
+        self.question = question
+        self.depends_on_question = depends_on_question
+
+    def as_flash_context(self) -> dict[str, str]:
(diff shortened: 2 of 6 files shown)


app/developers/templates/developers/deliver/manage_form.html
```diff
@@ -22,6 +22,29 @@
 {% block content %}
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
+      {% with errors = get_flashed_messages(with_categories="dependency_order_error") %}
```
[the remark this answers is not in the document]
sfount: Yeah agree no reason to not try and put something sensible in here for now, thanks for the prompt

app/common/data/interfaces/collections.py
```diff
@@ -244,13 +244,44 @@ def get_question_by_id(question_id: UUID) -> Question:
     return db.session.get_one(Question, question_id)
 
 
+class DependencyOrderException(Exception):
+    def __init__(self, question: Question, depends_on_question: Question):
+        super().__init__("You cannot move questions above answers they depend on.")
```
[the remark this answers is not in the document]
sfount: I've followed through on most of the suggestion here with a custom message based on what's happening - the only thing I haven't done is moved the template around just to avoid needing to rethink the "delete" warning too. Very happy to tidy that up following after we get a sense for it

They look like this now with custom messages

Screenshot 2025-07-03 at 14 17 54

Screenshot 2025-07-03 at 14 18 02

app/__init__.py
```diff
@@ -211,5 +212,6 @@ def index() -> ResponseReturnValue:
     # should make an intentional decision for when to be setting this
     app.config["SEND_FILE_MAX_AGE_DEFAULT"] = 3600
     app.add_template_global(AuthorisationHelper, "authorisation_helper")
+    app.add_template_global(FlashMessageType, "flash_message_type")
```
samuelhwilliams, line 215: One for us to take away into the future - wrapping enums/types somehow so we're not just bunging them all in a bit randomly.

app/common/data/interfaces/collections.py
```diff
@@ -244,13 +244,54 @@ def get_question_by_id(question_id: UUID) -> Question:
     return db.session.get_one(Question, question_id)
 
 
+class FlashableException(Protocol):
+    def as_flash_context(self) -> dict[str, str]: ...
```
samuelhwilliams, line 248: Nice 👍

app/common/data/interfaces/collections.py
```diff
@@ -244,13 +244,54 @@ def get_question_by_id(question_id: UUID) -> Question:
...
+            "message": self.message,
+            "question_id": str(self.question.id),
+            "question_text": self.question.text,
+            "depends_on_question_id": str(self.depends_on_question.id),
+            "depends_on_question_text": self.depends_on_question.text,
+        }
+
+
+# todo: we might want something more generalisable that checks all order dependencies across a form
+#       but this gives us the specific result we want for for the UX for now
```
samuelhwilliams, line 269:
```suggestion
#       but this gives us the specific result we want for the UX for now
```

samuelhwilliams (approved):
Nice, this looks and feels and works well 👍

samuelhwilliams (approved):

PR #424, opened 2025-07-01
FSPT-596 Block adding a condition if the answer we depend on comes after this question

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-596

## 📝 Description
This blocks managed conditions from being added if they depend on
answer to questions that have come before the current question. This
doesn't get into generic dependency checking for conditions, validations
and personalised text.

Note this implements the block in two places, this is because the end
user can actually do something about it when they're selecting a
question but we'll also ideally check the integrity at the point where
we're adding the expression for the user.

If there's a failure when the expression is being added its not handled
as there's nothing the user can directly do about it and it means prior
checks/ the flow is broken - we should hear about it in Sentry.

screenshot-jul-1-4

## 🧪 Testing
Make sure that you can't add a condition that depends on a question further up the forms question list.

### Review Readiness

### Performance and security

### Testing
~- [ ] End-to-end tests have been updated (if applicable)~

### DIFF
--- a/app/developers/forms.py
+++ b/app/developers/forms.py
@@ -52,9 +52,8 @@ class ConditionSelectQuestionForm(FlaskForm):
         self.question.choices = [
             (question.id, f"{question.text} ({question.name})") for question in get_supported_form_questions(question)
         ]
-        self.question.validators.append(self.question_depends_on_order)  # type: ignore[attr-defined]
 
-    def question_depends_on_order(self, form: "ConditionSelectQuestionForm", field: "Field") -> None:
-        depends_on_question = get_question_by_id(form.question.data)
+    def validate_question(self: "ConditionSelectQuestionForm", field: "Field") -> None:
+        depends_on_question = get_question_by_id(self.question.data)
         if not is_question_dependency_order_valid(self.target_question, depends_on_question):
             raise ValidationError("Select an answer that comes before this question in the form")


app/developers/deliver_routes.py
```diff
@@ -620,14 +621,18 @@ def add_question_condition_select_question(grant_id: UUID, question_id: UUID) ->
...
     if form.validate_on_submit():
-        return redirect(
-            url_for(
-                "developers.deliver.add_question_condition",
-                grant_id=grant_id,
-                question_id=question_id,
-                depends_on_question_id=form.question.data,
+        depends_on_question = get_question_by_id(form.question.data)
+        if not is_question_dependency_order_valid(question, depends_on_question):
+            form.question.errors.append("Select an answer that comes before this question in the form")  # type:ignore[attr-defined]
```
samuelhwilliams, line 626: This is a nice validation check; should we also just remove questions from the dropdown if they aren't valid selections?

Also should this check be part of the form validation itself, rather than on the view?

app/common/data/interfaces/collections.py
```diff
@@ -314,6 +322,13 @@ def clear_submission_events(submission: Submission, key: SubmissionEventKey, for
 
 
 def add_question_condition(question: Question, user: User, managed_expression: "ManagedExpression") -> Question:
+    if not is_question_dependency_order_valid(question, managed_expression.referenced_question):
+        raise DependencyOrderException(
+            "Cannot add managed condition that depends on a later question",
+            question,
+            managed_expression.referenced_question,
+        )
```
samuelhwilliams, line 330: Until we've got this in the DB (if we can 👀), really sensible to add this here as a way of making sure (/ as sure as possible) that we don't miss doing this validation somewhere up the chain.

app/developers/deliver_routes.py
sfount: > This is a nice validation check; should we also just remove questions from the dropdown if they aren't valid selections?

Yeah I did try this out but I think the trade off is that things not showing up for the user require a fairly deep understanding of the mechanisms or a fairly verbose explanation there in the UI explaining that it will only show a certain type of question, and be based on where you currently are in the form. Showing the questions gives you a hook to teach why you wouldn't want to use the answer to that question in this instance I guess but I can see pros and cons to both. 

I think something I'd like to do is to take some real and longer forms and model them up so we can get a sense for what these look like with more realistic data in there.

I'll definitely move the validation onto the form itself thats a good steer I keep forgetting about.

app/developers/forms.py
```diff
@@ -46,6 +47,14 @@ class ConditionSelectQuestionForm(FlaskForm):
...
         super().__init__(*args, **kwargs)
 
+        self.target_question = question
+
         self.question.choices = [
             (question.id, f"{question.text} ({question.name})") for question in get_supported_form_questions(question)
         ]
+        self.question.validators.append(self.question_depends_on_order)  # type: ignore[attr-defined]
+
+    def question_depends_on_order(self, form: "ConditionSelectQuestionForm", field: "Field") -> None:
```
samuelhwilliams, line 57: Could use a field validator: https://wtforms.readthedocs.io/en/3.0.x/validators/#custom-validators

but this is fine

samuelhwilliams (approved):
Another nice validation check 🙇

PR #430, opened 2025-07-01
Add test for unused jinja imports

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Clean up unused imports in jinja templates

### DIFF
--- a/tests/unit/test_jinja_templates.py
+++ b/tests/unit/test_jinja_templates.py
@@ -1,0 +1,32 @@
+from itertools import chain
+from pathlib import Path
+
+import pytest
+from jinja2.nodes import Call, FromImport, Name
+
+
+class UnusedJinjaImports(Exception):
+    def __init__(self, unused_imports, file):
+        separator = "\n  • "
+        formatted_unused_imports = separator.join(unused_imports)
+        super().__init__(f"\n  Found unused Jinja imports in {file}:{separator}{formatted_unused_imports}\n\n")
+
+
+root_dir = Path(__file__).parent.parent.parent.resolve()
+files = sorted(root_dir.glob("app/**/templates/**/*.html"))
+
+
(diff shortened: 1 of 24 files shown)


tests/unit/test_jinja_templates.py
```diff
@@ -0,0 +1,32 @@
...
+    # Tuples used are done so as (<import name>, <alias>)
+    def get_import_names(names):
+        return [name[1] if isinstance(name, tuple) else name for name in names]
+
+    parse_tree = app.jinja_env.parse(file.read_text())
+
+    calls = {node.node.name for node in parse_tree.find_all(Call) if isinstance(node.node, Name)}
+    imports = set(chain.from_iterable(get_import_names(node.names) for node in parse_tree.find_all(FromImport)))
+
+    if unused_imports := imports - calls:
```
sfount, line 31: This is brilliant 👍

tests/unit/test_jinja_templates.py
```diff
@@ -0,0 +1,32 @@
...
+        separator = "\n  • "
+        formatted_unused_imports = separator.join(unused_imports)
+        super().__init__(f"\n  Found unused Jinja imports in {file}:{separator}{formatted_unused_imports}\n\n")
+
+
+root_dir = Path(__file__).parent.parent.parent.resolve()
+files = sorted(root_dir.glob("app/**/templates/**/*.html"))
+
+
+@pytest.mark.parametrize("file", files, ids=[str(path.relative_to(root_dir)) for path in files])
```
sfount, line 19: I believe this will pass (or "skip") if this file gets refactored or moved around an no longer finds any HTML files. Having false confidence in this particular issue probably isn't a huge deal but if it could fail in that instance it would help us catch it (I'm not sure what a nice way to do that is, we might just want to trust the configuration!)

sfount (approved):
Really like this idea - it's so easy to miss after a refactor

sfount (approved):

PR #434, opened 2025-07-03
Add ty check

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Adds [ty](https://docs.astral.sh/ty/) to our app and makes a PR CI check for it. This check won't be enforced/enabled for now as `ty` is still very much in a non-prod-ready state, but I think useful to try it out for a little while and see how it feels.

If it does reach a production state, it will be a lot faster than `mypy`, so would be nice to switch out if it gets there.

```
✦3 ❯ rm -rf .mypy_cache/ && time uv run mypy
Success: no issues found in 136 source files

real    0m8.085s
user    0m7.400s
sys     0m0.612s

✦3 ❯ time uv run mypy
Success: no issues found in 136 source files

real    0m0.230s
user    0m0.153s
sys     0m0.074s

✦3 ❯ time uv run ty check
WARN ty is pre-release software and not ready for production use. Expect to encounter bugs, missing features, and fatal errors.
All checks passed!

real    0m0.099s
user    0m0.473s
sys     0m0.107s
```

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -70,10 +70,20 @@ jobs:
         uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba # v6
         with:
           enable-cache: true
-      - name: install dependencies
-        run: uv sync
       - name: run type checking
         run: uv run mypy
+
+  type_checking_ty:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@bd01e18f51369d5a26f1651c3cb451d3417e3bba # v6
+        with:
+          enable-cache: true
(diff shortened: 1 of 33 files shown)


sfount (approved):
The time difference is huge 🤯 . Being able to sub this out in CI and especially in pre-commit feels like it will be helpful for making those actions feel quick and encourage devs to run them often.

We've tried to run this in pre-commit but it looks like its hijacking/ clearing the terminal and removing previous output when it runs. We can probably come up with some workaround but we may even want to raise this on the ty project

Lets get this in as it will be great to see how it compares to current but everyone should feel like they can tweak or remove this until we make it mandatory!

samuelhwilliams:
> We've tried to run this in pre-commit but it looks like its hijacking/ clearing the terminal and removing previous output when it runs. We can probably come up with some workaround but we may even want to raise this on the ty project

I considered it yesterday and found https://github.com/astral-sh/ty/issues/269, but it basically just says: "we're so alpha that we're not considering precommit right now". I imagine they'll clock + sort this when they start working on it in earnest.

PR #435, opened 2025-07-03
FSPT-602 Form runner wrapper library

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-602

## 📝 Description
Adds a wrapper utility to manage the state routing and integrity checks
for using form pages. It's responsible for:
- consistently managing state between pages
- integrity checking if the user should be in this state
- consistently building the dynamic forms for pages based on the current
  submission

This should allow us to consistently implement a collection tasklist ->
forms -> questions experience in both Access Grant Funding and wherever
its needed in Deliver Grant Funding (preview while form building and
collecting information from grant teams/ stakeholders).

This switches the current deliver grant funding form builder forms to
using the new form runner wrapper.

## 🧪 Testing
- [x] We'll cover the library after we've got a shared idea its about the right shape for this solution.

### Review Readiness

### Performance and security

### Testing
- [x] End-to-end tests have been updated (if applicable)

### DIFF
--- a/app/common/collections/runner.py
+++ b/app/common/collections/runner.py
@@ -56,9 +56,9 @@ class FormRunner:
             self._check_your_answers_form = CheckYourAnswersForm(
                 section_completed=(
                     "yes" if self.submission.get_status_for_form(self.form) == SubmissionStatusEnum.COMPLETED else None
-                )
+                ),
+                all_questions_answered=all_questions_answered,
             )
-            self._check_your_answers_form.set_is_required(all_questions_answered)
 
     @classmethod
     def load(
@@ -104,7 +104,7 @@ class FormRunner:
 
         self.submission.submit_answer_for_question(self.question.id, self.question_form)
 
-    def save_is_form_completed(self, user: "User") -> None:
+    def save_is_form_completed(self, user: "User") -> bool:
         if not self.form:
             raise RuntimeError("Form context not set")
 
@@ -114,18 +114,20 @@ class FormRunner:
                 user=user,
                 is_complete=self.check_your_answers_form.section_completed.data == "yes",
             )
-        except ValueError as e:
+            return True
+        except ValueError:
             self.check_your_answers_form.section_completed.errors.append(  # type:ignore[attr-defined]
                 "You must complete all questions before marking this section as complete"
             )
-            raise ValueError("Failed to save form completed status") from e
+            return False
 
-    def submit(self, user: "User") -> None:
+    def complete_submission(self, user: "User") -> bool:
         try:
             self.submission.submit(user)
-        except ValueError as e:
+            return True
+        except ValueError:
             self._tasklist_form.submit.errors.append("You must complete all forms before submitting the collection")  # type:ignore[attr-defined]
-            raise ValueError("Failed to submit") from e
+            return False
 
     def to_url(
         self,
--- a/app/developers/deliver_routes.py
+++ b/app/developers/deliver_routes.py
@@ -1,4 +1,3 @@
-from contextlib import suppress
 from typing import TYPE_CHECKING
 from uuid import UUID
 
@@ -885,9 +884,7 @@ def submission_tasklist(submission_id: UUID) -> ResponseReturnValue:
     runner = DGFFormRunner.load(submission_id=submission_id, source=FormRunnerState(source) if source else None)
 
     if runner.tasklist_form.validate_on_submit():
-        with suppress(ValueError):
-            runner.submit(interfaces.user.get_current_user())
-
+        if runner.complete_submission(interfaces.user.get_current_user()):
             # todo: for now we'll email and confirm on submission but later DGF form builder will likely
             #       just return to the tasklist editing page on completion
             notification_service.send_collection_submission(runner.submission.submission)
@@ -930,8 +927,7 @@ def check_your_answers(submission_id: UUID, form_id: UUID) -> ResponseReturnValu
     )
 
     if runner.check_your_answers_form.validate_on_submit():
-        with suppress(ValueError):
-            runner.save_is_form_completed(interfaces.user.get_current_user())
+        if runner.save_is_form_completed(interfaces.user.get_current_user()):
             return redirect(runner.next_url)
 
     return render_template("developers/deliver/check_your_answers.html", runner=runner)
--- a/app/common/collections/forms.py
+++ b/app/common/collections/forms.py
@@ -144,8 +144,10 @@ class CheckYourAnswersForm(FlaskForm):
     )
     submit = SubmitField("Save and continue", widget=GovSubmitInput())
 
-    # the form should be validly optional unless all questions in the section have been answered
-    def set_is_required(self, all_questions_answered: bool) -> None:
+    def __init__(self, *args: Any, all_questions_answered: bool, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        # the form should be validly optional unless all questions in the section have been answered
         if all_questions_answered:
             self.section_completed.validators = [DataRequired("Select if you have completed this section")]
         else:


app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+        except ValueError as e:
+            self.check_your_answers_form.section_completed.errors.append(  # type:ignore[attr-defined]
+                "You must complete all questions before marking this section as complete"
+            )
+            raise ValueError("Failed to save form completed status") from e
+
+    def submit(self, user: "User") -> None:
+        try:
+            self.submission.submit(user)
+            notification_service.send_collection_submission(self.submission.submission)
```
samuelhwilliams, line 111: Interesting - maybe this means that we have to take the submit button out of the tasklist macro at some point? If we'd want to send the email in AGF but not in DGF for example.

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+
+                next_question = self.submission.get_next_question(self.question.id)
+                if next_question:
+                    return self.to_url(FormRunnerState.QUESTION, question=next_question)
+
+                return self.to_url(FormRunnerState.CHECK_YOUR_ANSWERS)
+            else:
+                return self.to_url(FormRunnerState.CHECK_YOUR_ANSWERS)
+
+        return self.to_url(FormRunnerState.TASKLIST)
```
samuelhwilliams, line 143: I think the routing logic here is a bit complex - maybe comments on each bit specifying the scenario would be useful?

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def validate(self) -> bool:
```
samuelhwilliams, line 145: nit: I wouldn't mind a slightly more descriptive name here. It's validating that it's ok to show the question page?

I think I get the idea that you just call `.validate()` on every page and based on the context it'll work it out, but I also wouldn't hate a more specific method for validity on each page (eg `validate_can_show_question_page`?) that just deals with its own rules.

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+        elif self.submission.is_completed:
+            self._valid = False
+        else:
+            self._valid = True
+
+        return self._valid
+
+    @property
+    def back_url(self) -> str:
+        # todo: persist the "tasklist" source when going "back" to check your answers
```
samuelhwilliams, line 164: :sweat: backlinks are the worst

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+from app.common.forms import GenericSubmitForm
+from app.common.helpers.collections import SubmissionHelper
+from app.extensions import notification_service
+
+if TYPE_CHECKING:
+    from app.common.collections.forms import DynamicQuestionForm
+    from app.common.data.models import Form, Question
+    from app.common.data.models_user import User
+
+runner_url_map = dict[
```
samuelhwilliams, line 16: maybe this moves to the `types` file as `TRunnerUrlMap`?

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+    FormRunnerState, Callable[["FormRunner", Optional["Question"], Optional["Form"], Optional[FormRunnerState]], str]
+]
+
+
+class FormRunner:
```
samuelhwilliams, line 21: This whole encapsulation feels really clean and the change it makes to what goes on in the views is night and day - very nice idea 🙇

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+    """Responsible for backing form runner pages, consistently takes care of:
+        - managing routing backwards and forwards
+        - integrity checks
+        - setting up forms
+
+    This allows us to implement the form runner in different domain environments consistently."""
+
+    url_map: ClassVar[runner_url_map] = {}
```
samuelhwilliams, line 29: the only iffy bit - but this is fine 👀

app/developers/deliver_routes.py
```diff
@@ -921,86 +877,63 @@ def collection_confirmation(submission_id: UUID) -> ResponseReturnValue:
     )
 
 
-@developers_deliver_blueprint.route("/submissions/<uuid:submission_id>/<uuid:question_id>", methods=["GET", "POST"])
-@is_platform_admin
-@auto_commit_after_request
-def ask_a_question(submission_id: UUID, question_id: UUID) -> ResponseReturnValue:
-    submission_helper = SubmissionHelper.load(submission_id)
-    question = submission_helper.get_question(question_id)
+class DGFFormRunner(FormRunner):
```
samuelhwilliams, line 880: nit: I wouldn't be massively upset with these subclasses just being defined in the `collections/runner` file, so that 'all' of the runners and runner config (url maps) live there?

app/common/templates/common/macros/collections.html
```diff
@@ -0,0 +1,180 @@
+{% from "govuk_frontend_jinja/components/back-link/macro.html" import govukBackLink %}
```
samuelhwilliams, line 1: We might want to split these into separate macro files at some point, but for now I think this is fine 👍

app/common/collections/runner.py
```diff
@@ -0,0 +1,203 @@
...
+    question: Optional["Question"] = None
+
+    def __init__(self, submission: SubmissionHelper):
+        self.submission = submission
```
samuelhwilliams, line 36: 🤏  small wonder if we should make this a semi-private thing and then proxy the important things; could go either way.

```suggestion
        self._submission = submission
```

app/common/collections/runner.py
sfount: Ah very good point 👍 this was originally done at the view layer so maybe simply passing it on to the submission helper and having the view manage the email lifecycle - everything should be wrapped in the same transaction either way 

I'll make that tweak anticipating that I think it helps show where the delineation is

app/common/collections/runner.py
sfount: I think this is a good idea but I've used the submission in a couple of places (views, templates) etc. so will do clean this up as a separate PR

app/common/collections/runner.py
```diff
@@ -0,0 +1,179 @@
...
+                user=user,
+                is_complete=self.check_your_answers_form.section_completed.data == "yes",
+            )
+        except ValueError as e:
+            self.check_your_answers_form.section_completed.errors.append(  # type:ignore[attr-defined]
+                "You must complete all questions before marking this section as complete"
+            )
+            raise ValueError("Failed to save form completed status") from e
+
+    def submit(self, user: "User") -> None:
```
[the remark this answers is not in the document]
sfount: Yeah I actually started like this but it felt weird to have `validate_can_show_question_page` have the same returns True/ False interface as `submit` even though one is just a checker and one mutates state with unpredictable exceptions. This model feels like it encourages the view to handle things it thinks are important in its own way (eventually) so going to leave it as it is for now but yeah had originally started with "just handle everything and return if it was successful or not" so very happy for us to go back to that if it feels more appropriate

app/developers/deliver_routes.py
```diff
@@ -921,86 +877,45 @@ def collection_confirmation(submission_id: UUID) -> ResponseReturnValue:
...
+    runner = DGFFormRunner.load(
+        submission_id=submission_id, question_id=question_id, source=FormRunnerState(source) if source else None
+    )
 
-        next_question = submission_helper.get_next_question(current_question_id=question_id)
-        if next_question:
-            return redirect(
-                url_for("developers.deliver.ask_a_question", submission_id=submission_id, question_id=next_question.id)
-            )
+    if not runner.validate_can_show_question_page():
```
samuelhwilliams, line 911: nit for future thought: should instantiating the runner run this check instead? Could be a bit magical, could be a bit reliable? https://github.com/communitiesuk/funding-service/commit/cd054ba6c2b07b04b71b84bab7d676d4bc13be6b

samuelhwilliams (approved):
This is a really solid start of pulling out the form running logic and making it reusable between DGF and AGF - very nice.

app/developers/deliver_routes.py
```diff
@@ -1009,54 +942,17 @@ def ask_a_question(submission_id: UUID, question_id: UUID) -> ResponseReturnValu
...
-        if back_link_from_context
-        else url_for(
-            "developers.deliver.ask_a_question",
-            submission_id=submission_helper.submission.id,
-            question_id=previous_question.id,
-        )
-    )
+    if runner.check_your_answers_form.validate_on_submit():
+        with suppress(ValueError):
+            runner.save_is_form_completed(interfaces.user.get_current_user())
```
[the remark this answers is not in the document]
sfount: Completely agree I think this happened by making small changes over a long time and not looking back over at the end -- ty will fix this up now so it doesn't propagate

PR #447, opened 2025-07-07
FSPT-515: Add grant team members and SSO flow to now use Invitations table

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-515

## 📝 Description
We currently create users directly when grant team members are added and when users try to authenticate via magic links, but this could lead to a lot of bad data in the user table (users never logging in, incorrect email addresses, etc.). We're now adding an Invitations table to act as a kind of 'pending users' table to account for this behaviour.

This is the third and final PR to cover this change:

1. _[Create the Invitation table and add relationships ](https://github.com/communitiesuk/funding-service/pull/381) (Merged)_
2. _[Update the magic link functionality to not create users until they login](https://github.com/communitiesuk/funding-service/pull/425) (Merged)_
3. Update the 'add grant team member' and SSO sign-in flow to use invitations 👈 (this PR)

Should be reviewable commit by commit - the final commit adds some tidying up to both routes touched by this change in order to have some compound interfaces and make the flow more readable, and adds tests for those new interfaces and the expected and edge-case behaviour of these flows.

Non-viz change

## 🧪 Testing
- Added loadsa integration tests for all the new interfaces and updated route flows and behaviour, including accounting for edge cases
- Updated all the other tests where needed
- Manually tested the changes in various permutations eg. adding a user to multiple grants and claiming all the invitations at once, trying to login with only expired invitations

### Review Readiness

### Performance and security

### Testing
- [X] I need the reviewer(s) to pull and run this change locally (yes please just to sense check 🙏 )

## 📚 Additional Notes
There is one potential edge case of a user logging in for the first time with some valid and some expired invitations, and in the current flow they'd claim the valid invitations but not know that the expired ones had expired (from a user's point of view, they wouldn't have access to all the grants that they might expect to have access to). This seems like it would be relatively rare, and can be solved by simply adding the user to those grants again, but might be something we want to cater for in future if this becomes problematic.

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -104,42 +104,37 @@ def sso_sign_in() -> ResponseReturnValue:
 @auto_commit_after_request
 def sso_get_token() -> ResponseReturnValue:
     result = build_msal_app().acquire_token_by_auth_code_flow(session.get("flow", {}), request.args)
-
     if "error" in result:
         return abort(500, "Azure AD get-token flow failed with: {}".format(result))
 
     sso_user = result["id_token_claims"]
+
+    # Does the user exist already?
     user = interfaces.user.get_user_by_azure_ad_subject_id(azure_ad_subject_id=sso_user["sub"])
 
-    if user is None:
-        # check if they have a valid invite, if not tell them to get a new one
+    # Platform admin route
+    if "FSD_ADMIN" in sso_user.get("roles", []):
+        user = interfaces.user.upsert_user_and_set_platform_admin_role(
+            azure_ad_subject_id=sso_user["sub"], email_address=sso_user["preferred_username"], name=sso_user["name"]
+        )
+
+    # Invitations route
+    elif user is None:
         user_invites = interfaces.user.get_usable_invitations_by_email(email=sso_user["preferred_username"])
+        # TODO: We should remove these 403s - MHCLG people should be able to login but not see anything if no roles
         if not user_invites:
             return render_template(
                 "common/auth/mhclg-user-not-authorised.html",
                 service_desk_url=current_app.config["SERVICE_DESK_URL"],
                 invite_expired=True,
             ), 403
-        # There were valid invite(s) so create a new user and roles
-        user = interfaces.user.upsert_user_by_azure_ad_subject_id(
+        user = interfaces.user.create_user_and_claim_invitations(
             azure_ad_subject_id=sso_user["sub"],
             email_address=sso_user["preferred_username"],
             name=sso_user["name"],
         )
-        for invite in user_invites:
-            # claim all usable invitations
-            interfaces.user.upsert_user_role(user=user, grant_id=invite.grant_id, role=invite.role)
-            interfaces.user.claim_invitation(invitation=invite, user=user)
 
-    redirect_to_path = sanitise_redirect_url(session.pop("next", url_for("index")))
-
-    if "FSD_ADMIN" in sso_user.get("roles", []):
-        user = interfaces.user.upsert_user_by_azure_ad_subject_id(
-            azure_ad_subject_id=sso_user["sub"],
-            email_address=sso_user["preferred_username"],
-            name=sso_user["name"],
-        )
-        interfaces.user.set_platform_admin_role_for_user(user)
+    # Existing User with roles route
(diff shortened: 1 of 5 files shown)


tests/integration/common/auth/test_auth.py
```diff
@@ -358,17 +353,72 @@ def test_platform_admin_remove_all_other_roles(self, anonymous_client, factories
...
             response = anonymous_client.get(url_for("auth.sso_get_token"), follow_redirects=True)
-            updated_user = db_session.scalar(select(User).where(User.azure_ad_subject_id == "someStringValue"))
+            updated_user = db_session.scalar(select(User).where(User.azure_ad_subject_id == "wer234"))
             assert AuthorisationHelper.is_platform_admin(updated_user) is True
             assert db_session.scalar(select(func.count()).select_from(UserRole)) == 1
 
         assert response.status_code == 200
 
+    def test_grant_member_with_valid_invites_first_login(self, anonymous_client, factories, db_session):
+        with patch("app.common.auth.build_msal_app") as mock_build_msal_app:
```
samuelhwilliams, line 368: optional nit: a small pre-condition check that the user is anonymous / has no roles here might be nice?

app/common/auth/__init__.py
```diff
@@ -104,34 +104,37 @@ def sso_sign_in() -> ResponseReturnValue:
 @auto_commit_after_request
 def sso_get_token() -> ResponseReturnValue:
```
samuelhwilliams, line 105: Nice refactor of this endpoint - commenting and separating out the main branches makes it feel easier to reason about, ty 🙇

app/common/auth/authorisation_helper.py
```diff
@@ -15,11 +15,7 @@ def is_logged_in(user: User | AnonymousUserMixin) -> bool:
     def has_logged_in(user: User | AnonymousUserMixin) -> bool:
         if isinstance(user, AnonymousUserMixin):
             return False
-        # FIXME: We should have some actual tracking of whether the user has logged in. This could either be a
-        #        field on the model called `last_logged_in_at` or similar, or we could only create entries in the user
-        #        table when the user actually logs in, rather than at invitation-time. Then we could simply trust that
-        #        if a user entry exists, they have definitely logged in.
-        return bool(user.name)
+        return bool(user.last_logged_in_at_utc)
```
samuelhwilliams, line 18: 🎉

app/common/data/interfaces/user.py
```diff
@@ -169,19 +169,26 @@ def remove_grant_team_role_from_user(user: User, grant_id: uuid.UUID) -> None:
...
     role: RoleEnum | None = None,
 ) -> Invitation:
+    # Expire any existing invitations for the same email, organisation, and grant,
+    # filtering on NULL if org/grant not passed
+    stmt = update(Invitation).where(
+        and_(
+            Invitation.email == email,
+            Invitation.is_usable.is_(True),
+            Invitation.grant_id == grant.id if grant else Invitation.grant_id.is_(None),
+            Invitation.organisation_id == organisation.id if organisation else Invitation.organisation_id.is_(None),
```
samuelhwilliams, line 185:
```suggestion
            (Invitation.grant_id == grant.id) if grant else Invitation.grant_id.is_(None),
            (Invitation.organisation_id == organisation.id) if organisation else Invitation.organisation_id.is_(None),
```

nit: a couple parens here might help make the order of operations a 🤏  cleaner. Genuinely unsure which way this falls right now - `(Invitation.grant_id == grant.id) if grant else Invitation.grant_id.is_(None)` or `Invitation.grant_id == (grant.id if grant else Invitation.grant_id.is_(None))`

MarcUsher: Ha fair point! I think for me your first suggestion is more readable to me personally and having just tried both the second one seems to lead to a bunch of SQLAlchemy and psycopg errors, so I'll go with the first one

MarcUsher: Haha gotcha, I misunderstood what you were saying 😆  yeah it looks like it's doing the first one (which is what we want), I've added the parens in to make it clearer 👍

samuelhwilliams (approved):
Nice!

PR #457, opened 2025-07-09
Fix validators on our form fields

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
In the recent shuffle to validators, we dropped off our DataRequired/InputRequired validation on form fields. This adds them back.

We currently don't have any questions shown to users which can't have an answer provided, so for now this will enforce that.


samuelhwilliams:
Superceded by https://github.com/communitiesuk/funding-service/pull/458

PR #463, opened 2025-07-10
FSPT-577: Add auth method to session and decorator

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-577

## 📝 Description
As a final layer of security, we want to make sure that the users use the right authentication method to access the right service (ie. Deliver Grant Funding should only be accessed via SSO). The previous PR added checks to stop this happening but, in case it somehow happens, we now inject a session variable `auth` when a user authenticates via magic link or SSO to indicate which auth method they used, and adds some additional checks in our Deliver Grant Funding route decorators which enforce auth to make sure only users authenticating via SSO can access these routes.

With the changes in this ticket, the `login_required` decorator also started to make less sense now that we have a clearer concept of Access Grant Funding routes and that Deliver Grant Funding users should always be authenticating with SSO, as the default `login_required` redirect was for SSO. Now adds a new decorator for Access Grant Funding routes to redirect to the magic link sign in page for unauthenticated users.

One of two PRs on this work:

- ~[Clean up SSO/Magic Link flows to include redirects and blocks](https://github.com/communitiesuk/funding-service/pull/453)~ ✅  **Merged**
- Add session variable to track if a user has authenticated via magic link or SSO, and add additional route security that magic link users should 403 if they visit any DGF routes 👈 (this PR)

No visible changes for screenshots!

## 🧪 Testing
- Auth via magic link and try to hit some Deliver Grant Funding routes (though this should have been protected by the previous PR)
- Mess with the session in tests to make sure you get redirected/500s/403s as appropriate

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -10,6 +10,7 @@ from app.common.auth.decorators import redirect_if_authenticated
 from app.common.auth.forms import SignInForm
 from app.common.auth.sso import build_auth_code_flow, build_msal_app
 from app.common.data import interfaces
+from app.common.data.types import AuthMethodEnum
 from app.common.forms import GenericSubmitForm
 from app.common.security.utils import sanitise_redirect_url
 from app.extensions import auto_commit_after_request, notification_service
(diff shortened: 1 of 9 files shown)


samuelhwilliams (approved):
Nice test coverage - looks good as another layer of security for the high-value DGF side of things 🙇

PR #472, opened 2025-07-14
FSPT-674: AnyOf managed expression for radios

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-674

## 📝 Description
Having implemented the [new radio question type](https://mhclgdigital.atlassian.net/browse/FSPT-672), we need to consider what managed expressions we should provide form builders.

For branching, form designers will want some questions to show up based on specific radio choices being selected. There is nothing else we could let a form designer branch on for radios.

There is no sensible managed validation we can provide for radios. Validation inherently ensures that the user picks one of the radio choices.

## 🧪 Testing
- Create a radios question
- Create another question
- Add a condition to question 2 based on question 1
- Do a test submission and check that question 2 shows up/does not show up correctly

### Review Readiness

### Testing

### DIFF
--- a/app/common/data/migrations/.current-alembic-head
+++ b/app/common/data/migrations/.current-alembic-head
@@ -1,1 +1,1 @@
-014_sync_data_type_enum
+015_any_of_radios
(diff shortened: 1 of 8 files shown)


samuelhwilliams:
Yeah I think the main interface for that is `build_managed_expression_form`, which was already taking the question. The form definition itself is mostly implementation detail, so didn't feel like it needed as much explicit testing itself.

MarcUsher (approved):
Nice! And we discussed on pair review why the tests didn't go mad with that refactor, makes sense 👍

PR #482, opened 2025-07-15
FSPT-686: Add "Yes/No" question

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-686

## 📝 Description
We’ve added support for radio questions to the service, allowing users to define any list of options. But we expect that, very often, form designers will want to add a simple “yes” / “no” question type. This is such a common use-case that we think it’s worth building a specific question type just for this specific type of radio question.

This implementation does not use the 'data source' implementation of other radio question. It felt like the options were either:

1. Have a single platform-level data source for yes/no that all questions of this type use.
2. Create a new yes/no data source for each question.

Option 1 hits on some functionality we don't have yet, so would have increased scope quite a bit. Option 2 would definitely be fine but also feels a tiny bit wasteful (absolutely negligible though).

Neither of the options feel fundamentally better than just having a hard-coded yes/no question that processes the information in a distinct way (treating it as a boolean rather than a string yes/no).

Also updates our seeded grant to include a yes/no question with a dependent question.

### Review Readiness

### Testing

### DIFF
--- a/app/common/collections/forms.py
+++ b/app/common/collections/forms.py
@@ -131,6 +131,15 @@ def build_question_form(question: Question, expression_context: ExpressionContex
                 widget=GovTextInput(),
                 validators=[InputRequired(f"Enter the {question.name}")],
             )
+        case QuestionDataType.YES_NO:
+            field = RadioField(
+                label=question.text,
+                description=question.hint or "",
+                widget=GovRadioInput(),
+                choices=[(1, "Yes"), (0, "No")],
+                validators=[InputRequired("Select yes or no")],
+                coerce=lambda val: bool(int(val)),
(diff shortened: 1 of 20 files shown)


sfount (approved):
Code looks good to me 👍 feeling nice to get that much functionality from not too big of a change set

Setting this up to just persist true/ false and keep it separate for now feels like the right move

Intuitively I do like the idea of this being a convenient way to set up a select with a platform supported data source, it would mean a few less managed expressions and one less way-that-things work. Maybe we can write this up as something to consider in the future "Data sources" epic, it should be a very small change to move over if needed

PR #489, opened 2025-07-15
WIP Allow the email question type to check for a certain domain

sfount:
## 🎫 Ticket
WIP

## 📝 Description
Currently just a thought experiment trying out how much needs to be added to hook up a new expression for a new question type.

This works both as a validator to ensure the question is being answered with an email address from a certain domain and as a method of branching - only showing certain questions if a previous answer is from a given domain.

If people think this is interesting then we can make it merge-able - just one call to make about if a generic "includes" is sufficient or if we'd want to prove whats needed to set a custom simpleeval function to ensure the domain part of the email address specifically is correct.

funding communities gov localhost_8080_developers_deliver_submissions_d9fdde55-fc7d-4ef7-8347-46e5f36dc646_b51759ce-0690-49d1-9a0f-d6f34eda28f0_source=check-your-answers

funding communities gov localhost_8080_developers_deliver_submissions_d9fdde55-fc7d-4ef7-8347-46e5f36dc646_b51759ce-0690-49d1-9a0f-d6f34eda28f0_source=check-your-answers (1)

funding communities gov localhost_8080_developers_deliver_submissions_d9fdde55-fc7d-4ef7-8347-46e5f36dc646_b51759ce-0690-49d1-9a0f-d6f34eda28f0_source=check-your-answers (2)

### DIFF
--- a/app/common/expressions/managed.py
+++ b/app/common/expressions/managed.py
@@ -10,7 +10,7 @@ from flask_wtf import FlaskForm
 from govuk_frontend_wtf.wtforms_widgets import GovCheckboxesInput, GovCheckboxInput, GovTextInput
 from markupsafe import Markup
 from pydantic import BaseModel, TypeAdapter
-from wtforms import BooleanField, IntegerField, SelectMultipleField
+from wtforms import BooleanField, IntegerField, SelectMultipleField, StringField
 from wtforms.fields.core import Field
 from wtforms.validators import DataRequired, InputRequired, Optional, ValidationError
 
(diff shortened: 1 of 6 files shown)


app/common/expressions/managed.py
```diff
@@ -496,6 +496,57 @@ def build_from_form(form: "_ManagedExpressionForm", question: "Question") -> "Is
...
+
+    @property
+    def message(self) -> str:
+        return f"The answer must be from “{self.domain}”"
+
+    @property
+    def statement(self) -> str:
+        # fixme: this could register a function with simpleeval to explicitly check the domain part of
+        #       a given email input
+        return f"'{self.domain}' in {self.safe_qid}"
```
samuelhwilliams, line 523: Yeah this sounds clever - would be interested in seeing an example of it. I've felt the lack on another expression I was writing; feels like you've got a thought and it'd be cool to see that explored/proved 🙇

PR #498, opened 2025-07-16
Stop 500s on test submission when task has no questions

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
The existing code would throw a 500 trying to generate a URL for the first question in a task on the tasklist page, if the task has no questions in it at all.

This patch fixes that by detecting "no questions" as part of the tasklist macro and rendering the text without an href. It also changes the status to "No questions" in red rather than "Not started" in blue, because it can't be started. 

This is intended to make it easier for a form designer to understand why the link isn't there, but does change the pattern slightly. We'd never want to have a live collection in this state, but I think we can rely on that being handled by human processes. And I think even if it did slip through, showing an end user this is better than them seeing an arbitrary 500.

### Before
image

### After
image

### Review Readiness

### Testing


app/common/data/types.py
```diff
@@ -68,6 +68,15 @@ class SubmissionStatusEnum(enum.StrEnum):
     COMPLETED = "Completed"
 
 
+class TasklistTaskStatusEnum(enum.StrEnum):
+    NOT_STARTED = SubmissionStatusEnum.NOT_STARTED.value
+    IN_PROGRESS = SubmissionStatusEnum.IN_PROGRESS.value
+    COMPLETED = SubmissionStatusEnum.COMPLETED.value
+    assert len(SubmissionStatusEnum) == 3
```
sfount, line 75: is this hanging over from something or left in to intentionally make you change this when adding more? maybe a comment if so

app/common/templates/common/macros/status.html
```diff
@@ -3,9 +3,10 @@
 {% macro status(status) %}
   {%
     set classMap = {
-        enum.submission_status.NOT_STARTED: "govuk-tag--grey",
-        enum.submission_status.IN_PROGRESS: "govuk-tag--blue",
-        enum.submission_status.COMPLETED: "govuk-tag--green",
+        enum.tasklist_task_status.NOT_STARTED: "govuk-tag--grey",
+        enum.tasklist_task_status.IN_PROGRESS: "govuk-tag--blue",
+        enum.tasklist_task_status.COMPLETED: "govuk-tag--green",
+        enum.tasklist_task_status.NO_QUESTIONS: "govuk-tag--red",
```
sfount, line 9: Is there a file missing that would have this actually calculated and this enum returned instead of a submission status enum? I might just be missing something big!

sfount (commented):
Nice! I've been confused about why I'm coming across this myself so a lot more confusing for actual users

A few small comments, things I might just have misunderstood

Copied over from other comms: 
> feels like rounding out a very sharp edge while building the form nice one - makes my think a little bit about better formalising the previous discussions on whats persisted about a status and whats calculated (i.e this one very likely a calculated only one)

app/common/templates/common/macros/status.html
samuelhwilliams: i've added a get_tasklist_status_for_form to the submission helper

srh-sloan (approved):
Works locally, and nice to tell the form designer that they haven't added any questions

PR #515, opened 2025-07-18
Answer type refactors

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
- Adds new answer types specifically for `Email` and `Url`.
- Updates web rendering of `Email` and `Url` answers to be links. This feels quite right for URLs, maybe for emails it doesn't feel quite right? Open to turning that off again.
- Refactors to all of the Answer classes:
  - Add a standard `Answer` suffix
  - Make all root and base pydantic model answer classes subclass from a specific custom base.
  - Add a test that fails if a new answer is added, to prompt writing tests.


app/common/helpers/collections.py
```diff
@@ -410,42 +408,42 @@ def generate_csv_content_for_all_submissions(self) -> str:
...
+def _form_data_to_question_type(question: "Question", form: DynamicQuestionForm) -> AllAnswerTypes:
     _QuestionModel: type[PydanticBaseModel]
 
     answer = form.get_answer_to_question(question)
 
     match question.data_type:
         case QuestionDataType.TEXT_SINGLE_LINE | QuestionDataType.EMAIL | QuestionDataType.URL:
-            return TextSingleLine(answer)
+            return TextSingleLineAnswer(answer)  # ty: ignore[missing-argument]
         case QuestionDataType.TEXT_MULTI_LINE:
```
srh-sloan, line 419: What are these missing-argument bits about?

srh-sloan (approved):
Nice, like the `Answer` suffix, makes it clearer what they are doing 👍

app/common/helpers/collections.py
samuelhwilliams: Ty doesn't seem to like us subclassing `RootModel` - to be honest I haven't dug very deeply, but it's spitting out this kind of error:

```
error[missing-argument]: No argument provided for required parameter `root`
   --> app/common/helpers/collections.py:418:20
    |
416 |     match question.data_type:
417 |         case QuestionDataType.TEXT_SINGLE_LINE | QuestionDataType.EMAIL | QuestionDataType.URL:
418 |             return TextSingleLineAnswer(answer)
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
419 |         case QuestionDataType.TEXT_MULTI_LINE:
420 |             return TextMultiLineAnswer(answer)  # ty: ignore[missing-argument]
    |
info: rule `missing-argument` is enabled by default

Found 1 diagnostic
```

But it works totally fine. Mypy doesn't complain. I think it might be a case where `ty` is still a bit immature, but I can't explain _exactly_ why it's telling us it's wrong.

If/when `ty` starts to recognise this as OK, it will complain about the `ty ignore` comment being unused and we'll be forced to clean it up, so it shouldn't be left dangling if it becomes incorrect.

PR #516, opened 2025-07-18
Rename variable

samuelhwilliams:
`errors` is a bad name for this - it's just standard flash messaging, not specifically errors.


srh-sloan (approved):
Much better! Thank you!

PR #521, opened 2025-07-18
Improving performance on the generate CSV function

srh-sloan:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-693

## 📝 Description
This ticket was to see if we had any potential performance issues with the generate CSV function. On invesigation, we were creating one query per submission when generating the csv output. With the changes in this ticket it is down to one query to generate the csv content for any number of submissions.

- README contains information on a new seeding command that creates collections with 100 submissions - this number can be changed if we need more for something
- Using any kind of performance testing plugin or instrumentation was out of scope for this work
- Adds 2 new unit tests that are now skipped - I used these to generate initial timings before making any changes, but they are no longer useful in their current state because the factory has cached all the data, so they don't generate any queries.

The conclusions that we have reduced the number of queries is based on the approach mentioned in the README file - we hacked the export csv endpoint to render a template, and then used the information in the flask debug toolbar.

## 🧪 Testing
This can be tested using the approach detailed in the README and the flask debug toolbar. Also tested before and after the change on a huddle with @MarcUsher and @samuelhwilliams as witnesses of the improvement 🙂 

### Review Readiness
- [ ] PR title and description provide sufficient context for reviewers

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -2,7 +2,7 @@ import uuid
 from typing import TYPE_CHECKING, Any, Never, Protocol
 from uuid import UUID
 
-from sqlalchemy import select, text
+from sqlalchemy import ScalarResult, select, text
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload, selectinload
 
@@ -84,6 +84,31 @@ def update_submission_data(submission: Submission, question: Question, data: All
     submission.data[str(question.id)] = data.get_value_for_submission()
     db.session.flush()
     return submission
+
+
+def get_all_submissions_with_mode_for_collection_with_full_schema(
+    collection_id: UUID, submission_mode: SubmissionModeEnum
+) -> ScalarResult[Submission]:
+    """
+    Use this function to get all submission data for a collection - it
+    loads all the question/expression/user data at once to optimise
+    performance and reduce the number of queries compared to looping
+    through them all individually.
+    """
+    return db.session.scalars(
+        select(Submission)
+        .where(Submission.collection_id == collection_id)
+        .where(Submission.mode == submission_mode)
+        .options(
+            joinedload(Submission.collection)
+            .selectinload(Collection.sections)
+            .selectinload(Section.forms)
+            .selectinload(Form.questions)
+            .selectinload(Question.expressions),
+            joinedload(Submission.events),
+            joinedload(Submission.created_by),
+        )
+    ).unique()
 
 
 def get_submission(submission_id: UUID, with_full_schema: bool = False) -> Submission:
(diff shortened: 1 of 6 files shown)


app/common/data/interfaces/collections.py
```diff
@@ -86,6 +86,31 @@ def update_submission_data(submission: Submission, question: Question, data: All
...
+        select(Submission)
+        .where(Submission.collection_id == collection_id)
+        .where(Submission.mode == submission_mode)
+        .options(
+            joinedload(Submission.collection)
+            .selectinload(Collection.sections)
+            .selectinload(Section.forms)
+            .selectinload(Form.questions)
+            .selectinload(Question.expressions),
+            joinedload(Submission.events),
```
samuelhwilliams, line 108: Ah I missed the commas earlier. We can switch this to a selectinload as long as we remove the comma above. The fact these are chaining together is not very obvious with those sneaky commas sporadically dotted in.

app/developers/commands.py
```diff
@@ -179,3 +179,43 @@ def seed_grants() -> None:
 
     db.session.commit()
     click.echo(f"Loaded/synced {len(export_data['grants'])} grant(s) into the database.")
+
+
+@developers_blueprint.cli.command(
+    "seed-grants-many-submissions", help="Load grants with 100 random submissions into the database"
+)
+def seed_grants_many_submissions() -> None:
```
samuelhwilliams, line 187: nice - this feels like a really handy thing to have locally so we can start caring about, and validating, some of the performance characteristics 🙇

app/common/data/interfaces/collections.py
srh-sloan: Nice! Fixed

tests/models.py
```diff
@@ -204,6 +204,87 @@ def _create_submission(mode: SubmissionModeEnum, complete_question_2: bool = Fal
             _create_submission(SubmissionModeEnum.LIVE, complete_question_2=True)
             _create_submission(SubmissionModeEnum.LIVE, complete_question_2=False)
 
+    @factory.post_generation  # type: ignore
+    def create_completed_submissions_conditional_question_random(  # type: ignore
```
samuelhwilliams, line 208: I think at some point maybe these just become helper functions rather than part of the definition of the factory itself - they feel super specific. But easy to lift out in the future.

samuelhwilliams (approved):
Nice - good to see the number of queries drop to a O(1) rather than O(n) 🙇 I think this sort of care+thinking will help ensure platform stability for the long-term.

There will be a lot more of this to do 😆

tests/models.py
srh-sloan: Yeah, I agree

PR #524, opened 2025-07-21
Make "Not answered" more like NOT_ASKED

samuelhwilliams:
This is meant to be clearly indicative of something other than user input, which "Not answered" could theoretically be mistaken for.


MarcUsher (approved):
Nice, good catch. Was gonna say maybe we could check this in a test but all the tests are checking completed submissions which won't have anything that's not answered  one for us to come back to when we come to make some questions not mandatory!

PR #541, opened 2025-07-23
FSPT-685: Pull data source item choices for expressions from DB

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-685

## 📝 Description
When setting up a managed expression for a question with a data source, we copy the information about the referenced options into the context of the expression. We also then copy+hardcode the option keys in the expression statement.

This change makes us dynamically reference the data source items through the DB relationships instead, preventing some copying and keeping things in sync better.

No viz changes

## 🧪 Testing
- Add some conditions that rely on a radio questions
- Save the conditions
- Check the Expression in the DB to see it's referring to `data_source` and that the data source items are not referenced in the context

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -30,14 +30,13 @@ from app.common.data.types import (
     SubmissionModeEnum,
     SubmissionStatusEnum,
 )
-from app.common.expressions.managed import BaseDataSourceManagedExpression
 from app.common.utils import slugify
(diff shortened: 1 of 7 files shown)


samuelhwilliams:
we've tried this out but aren't really happy with the developer/code compromises we've had to made, so abandoning the piece of work

PR #542, opened 2025-07-23
Fix "None of the above" default for radios

samuelhwilliams:
## Before
image

## After
image


MarcUsher (approved):
Pulled locally and the default is appearing again for me 🙏

PR #548, opened 2025-07-24
BAU: Fix data source item dependency exception

MarcUsher:
## 🎫 Ticket
BAU - lil bugfix

## 📝 Description
When you try to delete data source items on which another question depended on, we raised an error. However, with interfaces using other interfaces using other interfaces, the way in which this error was raised was wrong as it would preserve the original data source items but not rollback any other changes made to the form, as when the exception was raised it wasn't in a try/except block or using what would eventually be a base IntegrityError (but rather raising pre-emptively so as to have the context needed to build useful error messages).

We might want to take a look at where certain interfaces/which interfaces try to flush changes and raise exceptions so that this can be consistent and we don't end up with issues like this happening again - happy to shelve this PR if that's something we want to do now for this instance, or if we want to do a fuller sweep.

## 🧪 Testing
- Add a radios question with a 'None of the above' type option, and a question which depends on that 'None of the above'
- Edit the original question and try to untick the final answer option
- Get the correct error message and see that the rest of the question details (including the ticked 'final answer option' and original final answer wording) are preserved

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -479,6 +479,7 @@ def raise_if_data_source_item_reference_dependency(
             data_source_item_dependency_map[dependent_question].add(data_source_item)
 
     if data_source_item_dependency_map:
+        db.session.rollback()
         raise DataSourceItemReferenceDependencyException(
             "You cannot delete or change an option that other questions depend on.",
             question_being_edited=question,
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):

samuelhwilliams:
Nice, let's take this fix for now and follow up with a bit more thinking around how to prevent anything like this happening again

PR #549, opened 2025-07-24
BAU: Fix auth 500 error

MarcUsher:
## 🎫 Ticket
None - bugfix

## 📝 Description
Before the auth work around magic link redirects and the set up of some initial Access Grant Funding pages, we'd been using an abort(500) for authenticated users trying to return to the sign-in pages if they user didn't have an email address with an internal domain. Now we have user flows that required signing in with magic link (and therefore non-internal emails) this was outdated and incorrect behaviour. Instead we now redirect those users to the same place we do as the fallback for magic link login.
    
This behaviour also highlighted that a) we didn't have a default 500 template and b) Sentry wasn't picking up 500 errors by default. Adds in a template and consistent 500 error handler plus a call to `current_app.logger.error()` as part of our 500 error handler should mean we get notified of all 500 errors in Sentry going forward (I hope, if I've understood that right).

500 template:
image

## 🧪 Testing
- Comment out the `session["auth"]` additions in the sign-in flow for magic link or sso
- Sign out and delete any session cookie that exists
- Log in (now without the `session["auth"]` addition which prompts a 500 still, as we _do_ want to be notified if someone has somehow managed to login through either flow without this additional session variable) and a) get redirected to the 500 template and b) see an error in the logs.

### Review Readiness

### Performance and security

### Testing

## 📚 Additional Notes
The logger call might be slightly wrong as we don't have any explicit error calls in the new platform yet - happy to rejig this if necessary!


app/common/auth/decorators.py
```diff
@@ -68,8 +68,10 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> ResponseReturnValue:
             internal_domains = current_app.config["INTERNAL_DOMAINS"]
             if user.email.endswith(internal_domains):
                 return redirect(url_for("deliver_grant_funding.list_grants"))
-
-            return abort(500)
+            # There's no default 'landing page' yet for Access Grant Funding - on magic link sign-in we fallback to a
+            # redirect to this Access Grant Funding grants_list page in lieu of anything else so doing the same here
+            # (the issue is that this page is platform admins only, but we need to redirect people _somewhere_)
+            return redirect(url_for("developers.access.grants_list"))
```
samuelhwilliams, line 74: What if we just don't redirect them? 🤔

samuelhwilliams (approved):
I think happy for this to go in - handling 500s properly and stopping that 500 we were hitting both handled here 👍 

A bit of a question around how we make sure that people logging in dont land on access denied pages, but maybe we can tolerate that a tiny bit longer.

app/common/auth/decorators.py
MarcUsher: Yeah I did think of this - it felt like an ok temp solution as it's what we do if someone tries to login via magic link directly from the magic link sign-in page, but agree it's a bit gross that we're redirecting people to something we know 403s (like you say below it's maybe something we can tolerate a little while longer and deal with when there's more foundation to Access Grant Funding)

MarcUsher:
Thanks @samuelhwilliams 🙏  as discussed on the quick call I'll merge to get this in but also do testing to make sure the error logging added here works nicely with the 500 message we pass as part of the SSO get token flow 
```
result = build_msal_app().acquire_token_by_auth_code_flow(session.get("flow", {}), request.args)
    if "error" in result:
        return abort(500, "Azure AD get-token flow failed with: {}".format(result))
```
and if not will tweak the logger flow in a separate PR

PR #562, opened 2025-07-28
Allow platform admins to downgrade to grant team members

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-587

This is a very quick and easy change that will allow platform admins (ie us) to quickly and temporarily downgrade our accounts to that of a 'grant team member'. This should make it easy for eg product managers and designers to view the service through a "read only"/reduced permissions role.

It doesn't give us everything we might want for permission permutations (ie someone who's a member of multiple grants) - but it gets us quite a lot for very little effort.

### DIFF
--- a/app/developers/deliver_routes.py
+++ b/app/developers/deliver_routes.py
@@ -48,6 +48,7 @@ from app.common.data.types import (
     FormRunnerState,
     QuestionDataType,
     QuestionPresentationOptions,
+    RoleEnum,
     SubmissionModeEnum,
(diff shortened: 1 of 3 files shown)


MarcUsher (approved):
Ah nice! I think there was a stub ticket somewhere for this that we can probably mark as "done" for now (prob on the unrefined future work)? This is v tidy though, who knew having nice compound interfaces like that would come in so handy 😆

PR #564, opened 2025-07-28
Start adding the 'Reports' tab

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-719

## 📝 Description
Adds the first part of the 'Reports' tab, which is where deliver grant funding users will see and manage monitoring reports for a grant. This is intended to serve two types of users currently: form designers (internal to Funding Service), and 'form reviewers' (grant team members). Form designers will use this tab to set up and build out monitoring reports, including configuring the data collection and setting dates for when grant recipients can provide data. Grant team members will use this for a read-only view of the report that has been set up for them, in order to validate that they're happy with it. In the long term we hope that grant team members will be able to build the reports themselves, but we're not aiming at that for MVP.

Currently, none of the links on the page are functional. These will be hooked up by future tickets that add those pages.

A followup PR will add a new column to the `Collection` table to track the 'type' of collection - which for now will just be MONITORING, but could be eg APPLICATION or PROSPECTUS as well.

### Grant members - read only
image

### Grant admins - read/write
image

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/auth/authorisation_helper.py
+++ b/app/common/auth/authorisation_helper.py
@@ -30,6 +30,8 @@ class AuthorisationHelper:
     def is_grant_admin(grant_id: UUID, user: User | AnonymousUserMixin) -> bool:
         if isinstance(user, AnonymousUserMixin):
             return False
+        if AuthorisationHelper.is_platform_admin(user=user):
+            return True
         return any(
             role.role == RoleEnum.ADMIN and role.organisation_id is None and role.grant_id == grant_id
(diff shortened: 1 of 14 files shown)


sfount (approved):
Code looks good to me, question on if it's reasonable to hide eagerly loading collections behind a flag but I'm not 100% sure on other usage so doesn't feel like a hard requirement at this stage

PR #565, opened 2025-07-28
Add types to split collections

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-719

## 📝 Description
We expect to have multiple kinds of collection focused on specific funding domains, such as applications, monitoring reports and maybe prospectuses. This will let us filter appropriately to get just the collections we're interested in.

### DIFF
--- a/app/common/data/models.py
+++ b/app/common/data/models.py
@@ -11,6 +11,7 @@ from sqlalchemy_json import mutable_json_type
 from app.common.data.base import BaseModel, CIStr
 from app.common.data.models_user import Invitation, User
 from app.common.data.types import (
+    CollectionType,
     ExpressionType,
     ManagedExpressionsEnum,
     QuestionDataType,
@@ -54,6 +55,10 @@ class Grant(BaseModel):
         viewonly=True,
     )
 
+    @property
+    def reports(self) -> list["Collection"]:
+        return [collection for collection in self.collections if collection.type == CollectionType.MONITORING_REPORT]
+
 
 class Organisation(BaseModel):
     __tablename__ = "organisation"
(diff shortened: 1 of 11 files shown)


app/common/data/models.py
```diff
@@ -71,6 +76,8 @@ class Collection(BaseModel):
     #       identifier for linked collection/versioning.
     version: Mapped[int] = mapped_column(default=1, primary_key=True)
 
+    type: Mapped[CollectionType]
```
MarcUsher, line 79: Do we need any of the extra `mapped_column()` info here that we have on other columns that use enums?

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_reports.html
```diff
@@ -20,8 +20,7 @@ <h1 class="govuk-heading-l">
       {% if not grant.collections %}
```
MarcUsher, line 20: Can maybe update this to `if not grant.reports` now as well?

MarcUsher (commented):
One lil comment and one lil question 🤔

app/common/data/models.py
samuelhwilliams: I have updated 👀  good catch, it _maybe_ matters? 🤔

MarcUsher: Ha yeah I'm not 100% sure it matters 🙃 but I guess good to be consistent with the other enum columns. Merci !

app/developers/templates/developers/deliver/grant_developers.html
```diff
@@ -38,7 +38,7 @@ <h1 class="govuk-heading-l">
     <div class="govuk-grid-column-full">
       <h2 class="govuk-heading-m govuk-!-margin-top-4">Reports</h2>
 
-      {% if not grant.collections %}
+      {% if not grant.reports %}
         <p class="govuk-body">This grant has no monitoring reports.</p>
       {% else %}
         {% for collection in grant.collections %}
```
MarcUsher, line 44: This is v picky but now you've update the above to `grant.reports` can maybe update this line too, or could lead to weird behaviour once we have more collection types?

samuelhwilliams: hopefully the developer pages disappear before we have more collection types 👍

MarcUsher: Yes true that's a v good point!

MarcUsher (approved):

PR #570, opened 2025-07-29
FSPT-721: manage collections

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-721

## 📝 Description
Adds a page to the Reports tab for 'managing' a collection: specifically renaming and deleting it. We aren't super confident in the terminology here, but will go with it and adjust based on feedback if we need to.

We haven't yet dealt with collection versioning, which we need to, but we try to anticipate some production-based problems by making it impossible to delete a collection (version) that has associated live submissions.

It's possible that we should actually block all changes to a collection that's gone live, but we don't have the concept of collection versioning built in yet. So there's bits to revisit here.

### Default state
image

### Clicked 'delete report'
image

### Has live submissions
image

## 🧪 Testing
TBD

### Review Readiness

### Performance and security

### Testing


samuelhwilliams:
@Kateharries have added some words+and a state that blocks deleting the report - can you have a review/think about content please?

Kateharries (commented):
I think it should say 'Important' rather than 'Warning'. 

"It's possible that we should actually block all changes to a collection that's gone live, but we don't have the concept of collection versioning built in yet. So there's bits to revisit here." - yes, will need to come back to this and also possibly adding in a page to ask the user if they are sure they want to delete, rather than using a banner.

I'm wondering if it's not possible to delete (as there are live submissions) we don't show the content and link at all?

samuelhwilliams:
Changed to Important @Kateharries  👍

app/common/data/interfaces/collections.py
```diff
@@ -57,7 +57,13 @@ def create_collection(*, name: str, user: User, grant: Grant, version: int = 1,
...
+def get_collection(
+    collection_id: UUID,
+    grant_id: UUID | None = None,
+    type_: CollectionType | None = None,
+    version: int | None = None,
+    with_full_schema: bool = False,
+) -> Collection:
     """Get a collection by ID and optionally version.
 
     If you do not pass a version, it will retrieve the latest version (ie highest version number).
```
MarcUsher, line 69: Small thing but is it worth us adding a line here in the docstring to say the `grant_id` and `type_`filters are only applied if you don't pass a version?

app/deliver_grant_funding/templates/deliver_grant_funding/reports/manage.html
```diff
@@ -0,0 +1,46 @@
+{% extends "deliver_grant_funding/grant_base.html" %}
+{% from "govuk_frontend_jinja/components/notification-banner/macro.html" import govukNotificationBanner %}
+
+{% set page_title = "Reports" %}
+{% set active_item_identifier = "reports" %}
+
```
MarcUsher, line 6: It could be nice to have a backlink on this page to take us back to the main reports tab? `Cancel` does the same job but some users might be looking for a backlink I suppose.

MarcUsher (commented):
Just a couple of small comments, looks good though! Loooots of tests 👍😆😭

app/common/data/interfaces/collections.py
samuelhwilliams: oh yeah no that's gross and probably bad logic, you should just call me out directly 😂

app/deliver_grant_funding/templates/deliver_grant_funding/reports/manage.html
samuelhwilliams: ah screw it, I'll just add it to both places 👀

MarcUsher (approved):
Noice one 🥳

PR #575, opened 2025-07-30
FSPT-727: Add checkbox question type

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-727

## 📝 Description
Adding the Checkbox question type to our form builder and form runner, making use of a lot of the heavy lifting already done for radios.

- Implements the ability to add a final answer for users if none of the options are appropriate.
- Caps the max checkbox options at 10 (+1 if including a final 'Other' option)
- Makes use of GOV.UK checkbox component's 'exclusive' option for the final 'other' option which will deselect any other checkboxes if selected, and vice versa. Also adds a fallback validator to the form in case JS is disabled
- Export to CSV lists choices as line-separated strings

Still to do in subsequent PRs:
- Update e2e tests
- Add a possible Condition for Checkbox questions

## 🧪 Testing
- Add Checkbox question types with and without the 'other' fallback option
- Edit the questions and see that the options and data is updated correctly
- Test the task/report and check the database and CSV export that the choices are correctly recorded/stored/serialised etc.

### Review Readiness
- [X] PR title and description provide sufficient context for reviewers
- [X] Commits are logical, self-contained, and have clear descriptions

### Performance and security
- ~[ ] No N+1 query problems introduced~
- ~[ ] Any new DB queries include appropriate where clauses based on the user's permissions~

### Testing
- [X] I have tested this change and it meets the acceptance criteria for the ticket
- [X] I need the reviewer(s) to pull and run this change locally
- [X] New (non-developer) functionality has appropriate unit and integration tests
- ~[ ] End-to-end tests have been updated (if applicable)~ _E2E test flow will be updated in a subsequent PR_ 
- [X] Edge cases and error conditions are tested

PR #584, opened 2025-08-04
FSPT-727: Add 'Specifically' managed expression for Checkboxes questions

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-727

## 📝 Description

👉  [First PR adding checkboxes](https://github.com/communitiesuk/funding-service/pull/575)

Adds support for a new managed expression type, `Specifically`, for use with Checkbox questions. 

This can be used for conditional routing to allow the form designer to select exactly one option from a list the list of data source items from a checkbox question type - if the user selects that option (singly or amongst other selected options) then the expression evaluates to True and the conditional question is shown.

_NB. Yes, 'Specifically' is not a nice name for it, I agree, but having spent more time than I should have thinking about it I couldn't come up with an alternative that worked as well - `OneOf` suggests that it could be one or more of the subset of options, `Exactly` suggests mathematical equality._

## 🧪 Testing
- Add a checkbox question type
- Add a subsequent question (any type)
- Add a condition to the subsequent question targeting the checkbox question type and select one of the data list options for the condition
- Test the form and check that the routing works as expected (should display the conditional question if the user selects only the specified option or selects it as part of a number of checkbox choices)
- Try to edit the original question and delete the option on which the condition relies and see that you get an error
- Try to delete the original question and see that you get an error that other questions depend on it

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/collections/types.py
+++ b/app/common/collections/types.py
@@ -127,8 +127,8 @@ class MultipleChoiceFromListAnswer(SubmissionAnswerBaseModel):
     def get_value_for_form(self) -> list[str]:
         return [choice["key"] for choice in self.choices]
 
-    def get_value_for_expression(self) -> list[str]:
-        return [choice["key"] for choice in self.choices]
+    def get_value_for_expression(self) -> set[str]:
+        return {choice["key"] for choice in self.choices}
 
     def get_value_for_text_export(self) -> str:
         return "\n".join(choice["label"] for choice in self.choices)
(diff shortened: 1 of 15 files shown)


samuelhwilliams (approved):
Lovely stuff - nice work 🙇 (lots of pair review+commenting out of band, sorry

PR #598, opened 2025-08-07
FSPT-605 Add group component type

sfount:
Add a group component type that extends and shares properties with the
core component table.

The group sub component will be able to access child components that it
contains (currently both questions and other groups).

Adds basic group component interfaces to ensure that all of the
schema and migrations connect. These interfaces will be used and
extended by the developer pages to let us add and manage groups on the
form management pages.


samuelhwilliams (dismissed):

samuelhwilliams:
🤏  reminder to squash merge/fixup commits also 🙇

PR #599, opened 2025-08-07
FSPT-605 Support nested components 

sfount:
Add loaders to the form/ collection/ submission interfaces to
appropriately load all of the possible nested components for a given
schema. This uses a fairly unoptimised/ naive approach for now but it
should be consistent given any number of components (within some upper
bound) and heavily leans into letting the ORM manage nested component
state. We can optimise the queries and nested parsing in the future but
we may have to workaround the ORM configuration to make sure developers
know which changes to the model should expect database changes and which
are read-only representations of the schema.


app/common/data/interfaces/collections.py
```diff
@@ -251,8 +299,20 @@ def get_form_by_id(form_id: UUID, grant_id: UUID | None = None, with_all_questio
         )
 
     if with_all_questions:
-        # todo: this will need refining again when we have different levels of grouped questions
-        query = query.options(selectinload(Form.components).joinedload(Component.expressions))
+        # todo: this needs to be rationalised with the grant_id behaviour above, having multiple places to
+        #       specify joins and options feels risky for them to collide or produce unexpected behaviour
```
samuelhwilliams, line 303: yeah good to call out 👍

app/common/data/models.py
```diff
@@ -226,20 +226,42 @@ class Form(BaseModel):
...
     )
 
     @property
     def questions(self) -> list["Question"]:
         """Consistently returns all questions in the form, respecting order and any level of nesting."""
-        return [component for component in self.components if isinstance(component, Question)]
+        return get_ordered_nested_questions_for_components(self.components)
+
+
+# todo: unit test reasonably extensively
```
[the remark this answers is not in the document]
sfount: I wouldn't say, extensively 🫣 https://github.com/communitiesuk/funding-service/pull/599/files#diff-dcf73ac94158211638e5a6e3c13d8165a7231afa16504518e7bdacb370eb6c0fR66-R80

app/common/data/interfaces/collections.py
```diff
@@ -119,16 +135,35 @@ def get_all_submissions_with_mode_for_collection_with_full_schema(
...
+            .selectinload(Form._all_components)
+            .joinedload(Component.expressions),
+            # get any nested components in one go
             joinedload(Submission.collection)
-            .selectinload(Collection.sections)
-            .selectinload(Section.forms)
+            .joinedload(Collection.sections)
+            .joinedload(Section.forms)
+            .selectinload(Form._all_components)
+            .selectinload(Component.components.and_(Component.type == ComponentType.GROUP))
```
samuelhwilliams, line 158: I think same question here? Maybe I've just not thought about this enough 😬

app/common/data/interfaces/temporary.py
```diff
@@ -100,9 +100,9 @@ def delete_question(question: Question) -> None:
     # todo: when/if this becomes a non-temporary interface, TEST THOROUGHLY. The OrderingList we're using for this
     # definitely has a few quirks.
     db.session.delete(question)
-    if question in question.form.questions:
-        question.form.components.remove(question)
-    question.form.components.reorder()
+    if question in question.contained_by.components:
```
samuelhwilliams, line 103: _Might_ want to call it `container` instead of `contained_by`, which makes me think of created_by/updated_by which are users. But that feels very personal preference so - no big feeling

app/common/data/models.py
```diff
@@ -226,20 +226,42 @@ class Form(BaseModel):
         UniqueConstraint("slug", "section_id", name="uq_form_slug_section"),
     )
 
+    # support fetching all of a forms components so that the selectin loading strategy can make one
+    # round trip to the database to optimise this further only load components flat like this and
+    # manage nesting through properties rather than subsequent declarative queries
```
samuelhwilliams, line 231: 👍  ta for the comment, useful

app/common/helpers/collections.py
```diff
@@ -236,9 +236,21 @@ def get_ordered_visible_forms_for_section(self, section: "Section") -> list["For
...
+        # we can optimise this to exit early and do this in a sensible order if we switch
+        # to going through questions in a nested way rather than flat
+        def get_all_conditions(question: "Component") -> list["Expression"]:
+            conditions = []
+
+            # start outside and move in from top level conditions to innermost
+            if question.parent:
+                conditions.extend(get_all_conditions(question.parent))
+            conditions.extend(question.conditions)
+            return conditions
```
samuelhwilliams, line 250: possibly a bit of extra testing for this logic?

samuelhwilliams (approved):

PR #600, opened 2025-08-07
Make data source item keys case insensitive

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
It is very unlikely (and probably a mistake) if someone enters two data source items with just case differences.

### DIFF
--- a/app/common/data/migrations/versions/025_ds_item_key_case_insensitive.py
+++ b/app/common/data/migrations/versions/025_ds_item_key_case_insensitive.py
@@ -1,0 +1,26 @@
+"""make data source item keys case insensitive
+
+Revision ID: 025_ds_item_key_case_insensitive
+Revises: 024_add_specifically_mgd_exp
+Create Date: 2025-08-07 15:05:03.406920
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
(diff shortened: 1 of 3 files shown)


app/common/data/migrations/versions/025_ds_item_key_case_insensitive.py
```diff
@@ -0,0 +1,26 @@
...
+
+revision = "025_ds_item_key_case_insensitive"
+down_revision = "024_add_specifically_mgd_exp"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("data_source_item", schema=None) as batch_op:
+        batch_op.alter_column("key", existing_type=sa.VARCHAR(), type_=postgresql.CITEXT(), existing_nullable=False)
```
sfount, line 21: Would this cause existing environment migrations to fail the unique constraint? As the commit suggests I think this is unlikely so hopefully no current instances to resolve those as part of the migration

sfount (approved):

app/common/data/migrations/versions/025_ds_item_key_case_insensitive.py
samuelhwilliams: Yep, it would

PR #601, opened 2025-08-07
Export as CSV from the reports tab

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Adds an initial pass of exporting a CSV of all submissions to the reports tab. This endpoint is not really optimised at all - we'll have to keep an eye on it as the number of submissions/etc grows.

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -388,6 +388,7 @@ class CollectionHelper:
         ]
 
     def generate_csv_content_for_all_submissions(self) -> str:
+        # TODO: make me generate rows (yield) rather than hold everything in memory
         metadata_headers = ["Submission reference", "Created by", "Created at"]
         question_headers = {
             question.id: f"[{question.form.title}] {question.name}"
--- a/app/deliver_grant_funding/routes/reports.py
+++ b/app/deliver_grant_funding/routes/reports.py
@@ -1,7 +1,8 @@
+import io
 import uuid
 from uuid import UUID
 
-from flask import abort, current_app, flash, redirect, render_template, request, url_for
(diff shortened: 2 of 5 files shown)


app/deliver_grant_funding/routes/reports.py
```diff
@@ -325,9 +326,41 @@ def list_submissions(grant_id: UUID, report_id: UUID, submission_mode: Submissio
         grant=report.grant,
         report=report,
         helper=helper,
+        submission_mode=submission_mode,
     )
 
 
+@deliver_grant_funding_blueprint.route(
+    "/grant/<uuid:grant_id>/report/<uuid:report_id>/submissions/<submission_mode:submission_mode>/export/<export_format>",
```
sfount, line 334: Plenty in the commit messages about this not yet being finalised or optimised but just a thought for the future when we know more about the need and have the full requirements, this `export_format` feels a bit like its trying to recreate Accept/ Content-Type headers which hopefully it wouldn't be too much effort to just hook in here, the client could then specify if it wants `text/csv`, `application/json`, `application/vnd.ms-excel` and these methods could route based on which was provided

app/common/helpers/collections.py
```diff
@@ -388,6 +388,7 @@ def get_all_possible_questions_for_collection(self) -> list["Question"]:
         ]
 
     def generate_csv_content_for_all_submissions(self) -> str:
+        # TODO: make me generate rows (yield) rather than hold everything in memory
```
sfount, line 391: 👍

sfount (approved):
I think fine to keep an eye on this when we've got more of a feel for the volume/ variety of data, maybe on a list of a final pass of checks before prod

PR #606, opened 2025-08-08
FSPT-605 Add group pages and macros

sfount:
Updates the question table to include appropriately rendering groups as
summary cards.

This commit also adds the "add group" handler and tweaks the interface
actions for moving components to make sure they work generically with
both questions and groups.

https://github.com/user-attachments/assets/56946b97-28cf-4911-b842-1bbc79392c7e

### DIFF
--- a/app/common/templates/common/macros/move-up-down-table.html
+++ b/app/common/templates/common/macros/move-up-down-table.html
@@ -1,6 +1,6 @@
 {# todo: I think all of this can be accomplished by overriding some design system classes #}
 {#       which would allow us to use the standard summary list macro with a few style changes #}
-{% macro moveUpDownTable(table_rows) %}
+{% macro table(table_rows) %}
   <dl class="govuk-summary-list app-move-up-down-table">
     {% for row in table_rows %}
       <div class="govuk-summary-list__row app-move-up-down-table__row">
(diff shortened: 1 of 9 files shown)


app/common/templates/common/macros/move-up-down-table.html
```diff
@@ -40,3 +40,53 @@
...
+              })
+            }}
+          {% endif %}
+        </h2>
+        <ul class="govuk-summary-card__actions">
+          <li class="govuk-summary-card__action">
+            <a class="app-link-button" href="#">Manage group</a>
+          </li>
+          <li class="govuk-summary-card__action">
+            {# todo: this is based on the template context again, there's a way of generically passing down context #}
```
samuelhwilliams, line 65: Please can we not - can we just pass in the things we need to the macro? 🙏

sfount: yeah sorry comment phrasing bad - I meant in this instance theres a way of passing down the context we need into this macro, vs. using `with context` (which the current macro does)

samuelhwilliams:
Also just so it's written down (though I suspect this is in your head already), one of these PRs should also:

- update the seeded grant(s) to have some question groups
- update the end to end tests to add/edit/etc question groups

app/common/templates/common/macros/move-up-down-table.html
sfount: https://github.com/communitiesuk/funding-service/pull/606/commits/b76f0d5f7f91c4c5cbcbc878867c01b5762b2eed

samuelhwilliams (approved):

PR #624, opened 2025-08-12
FSPT-606 Edit/ manage group page

sfount:
This adds an edit group page with updates to make journeys work for:
- listing and adding conditions for a group
- listing and adding questions that belong to a group

This commit needs to be extended with the "same page" presentation option and to be included in the developer journey end to end test runner.

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -521,7 +521,7 @@ def get_component_by_id(component_id: UUID) -> Component:
 
 
 class FlashableException(Protocol):
-    def as_flash_context(self) -> dict[str, str]: ...
+    def as_flash_context(self) -> dict[str, str | bool]: ...
 
 
 class DependencyOrderException(Exception, FlashableException):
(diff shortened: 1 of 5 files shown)


app/developers/deliver_routes.py
```diff
@@ -716,6 +761,60 @@ def edit_question(
...
+) -> ResponseReturnValue:
+    group = get_component_by_id(component_id=group_id)
+
+    confirm_deletion_form = ConfirmDeletionForm()
+    if "delete" in request.args:
+        try:
+            raise_if_question_has_any_dependencies(group)
+
+            if confirm_deletion_form.validate_on_submit() and confirm_deletion_form.confirm_deletion.data:
+                delete_component(group)
```
samuelhwilliams, line 785: Deleting a group that contains a question referenced elsewhere is throwing a 500 rather than showing a nice error message

sfount: thanks very helpful, fixed this up + moving questions that depend on questions inside the group in a commit just pushing (I think that covers all combinations but we'll do another run through)

samuelhwilliams:
Are we at the point where we can just build this straight in the reports tab?

app/common/data/interfaces/collections.py
```diff
@@ -582,40 +586,77 @@ def as_flash_contexts(self) -> list[dict[str, str]]:
...
+    )
+    compare_target_swap_components = (
+        [target_swap_component] + target_swap_component.all_components
+        if isinstance(target_swap_component, Group)
+        else [target_swap_component]
+    )
+    for c in (
+        [target_component] + target_component.all_components
+        if isinstance(target_component, Group)
+        else [target_component]
```
samuelhwilliams, line 606: Can we reuse `compare_target_components`? This looks the same?

PR #634, opened 2025-08-13
Phase 2 content changes

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Phase 2 content tweaks from https://mhclg-my.sharepoint.com/:w:/g/personal/kate_harries_communities_gov_uk/ET41SSBkUTFPspvmgu6fB48BtOAkmcCBxxG8pKE4VjkztQ?e=FDvRbI

### DIFF
--- a/app/deliver_grant_funding/templates/deliver_grant_funding/reports/edit_question.html
+++ b/app/deliver_grant_funding/templates/deliver_grant_funding/reports/edit_question.html
@@ -4,6 +4,7 @@
 {% from "govuk_frontend_jinja/components/notification-banner/macro.html" import govukNotificationBanner %}
 {% from "deliver_grant_funding/macros/dependency_banner.html" import dependency_banner %}
 {% from "govuk_frontend_jinja/components/summary-list/macro.html" import govukSummaryList %}
+{% from "govuk_frontend_jinja/components/table/macro.html" import govukTable %}
 
 {% set page_title = "Edit question - " ~ db_form.title %}
 {% set active_item_identifier = "reports" %}
@@ -28,7 +29,7 @@
         {% if flashes %}
           {% set question_created_content %}
             <a class="govuk-notification-banner__link" href="{{ url_for('deliver_grant_funding.list_task_questions', grant_id=grant.id, form_id=db_form.id) }}">Return to the task</a>
-            to add more questions, or edit this question to add conditions and validations.
+            to add more questions. Continue editing this question to add conditions and validations.
           {% endset %}
           {{
             govukNotificationBanner({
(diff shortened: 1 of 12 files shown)


app/deliver_grant_funding/templates/deliver_grant_funding/reports/edit_question.html
```diff
@@ -217,9 +218,21 @@ <h2 class="govuk-heading-m">Validations</h2>
       </form>
 
       <h2 class="govuk-heading-m govuk-!-margin-top-5">Manage question</h2>
-      <p class="govuk-body">
-        <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('deliver_grant_funding.edit_question', grant_id=grant.id, question_id=question.id, delete='') }}">Delete this question</a>
-      </p>
+      <div class="govuk-body">
+        {% if deleted_errors %}
+          <p class="govuk-body">This question cannot be deleted because the user's answer decides what other questions will be asked.</p>
+          <p class="govuk-body">To delete this question, you'll need to remove conditions from the other questions or delete them:</p>
```
MarcUsher, line 224: In the word doc Kate's put in a comment if we can make "questions" only plural here if there are more than one questions that rely on this one, otherwise just the singular?

app/deliver_grant_funding/templates/deliver_grant_funding/reports/edit_question.html
```diff
@@ -217,9 +218,21 @@ <h2 class="govuk-heading-m">Validations</h2>
...
+
+          <ul class="govuk-list govuk-list--bullet">
+            {% for error in deleted_errors %}
+              {% set question_link = url_for("deliver_grant_funding.edit_question", grant_id=error.grant_id, question_id=error.question_id) if not error.question_is_group else url_for("deliver_grant_funding.list_group_questions", grant_id=error.grant_id, group_id=error.question_id) %}
+              <li><a class="govuk-link govuk-link--no-visited-state" href="{{ question_link }}">{{ error.question_text }}</a></li>
+            {% endfor %}
+          </ul>
+        {% else %}
+          <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('deliver_grant_funding.edit_question', grant_id=grant.id, question_id=question.id, delete='') }}">Delete this question</a>
+        {% endif %}
```
MarcUsher, line 234: - For me this currently only lists & links off to one question that has conditions, but wording suggest it should show all of them?
eg.
image

All three of these conditions rely on the first question
image

MarcUsher (commented):
This is looking good, thank you both for working through all of these! 

Just a couple of things I think are missed from the content doc/huddle, which wasn't in the things we wanted to split out into separate PRs:

- Success flash when adding a condition or validation (see bottom of p5 of the word doc)
- Small content tweaks to the `Include an 'other' option` and `'Other' option text` are missing (should be `Include ‘Other’ option` and `Enter custom text or leave as ‘Other’`)
- Error message when user tries to delete list options that are relied on by conditions - we talked about consolidating these into one using Chanel's suggested copy and just using a normal error summary rather than a flash (so the erroring input can be highlighted)

PR #671, opened 2025-08-21
FSPT-751 Extend integrity checks for deleting a group

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-751

## 📝 Description
Assert that at any level of depth a question can be removed and preserve
an appropriate ordering. Removing a group will also cascade and remove
questions in the group.

The existing integrity checks for deleting components directly checks if
that component is relied on by any other component in the schema. This
change updates the logic to include checking if that component has any
child components and making sure they're also not relied on.

Assert that the appropriate error message is shown to the user if this
happens.

funding communities gov localhost_8080_grant_caf0ce3f-5175-f69c-66a9-41e2c2245845_group_b9a4f920-7704-4799-92b9-c4457e1c8705_questions (2)

### DIFF
--- a/app/common/data/models.py
+++ b/app/common/data/models.py
@@ -251,7 +251,6 @@ class Form(BaseModel):
         return get_ordered_nested_components(self.components)
 
 
-# todo: unit test reasonably extensively
 def get_ordered_nested_components(components: list["Component"]) -> list["Component"]:
     """Recursively collects all components from a list of components, including nested components."""
     flat_components = []
@@ -438,6 +437,10 @@ class Group(Component):
     @property
     def questions(self) -> list["Question"]:
         return [q for q in get_ordered_nested_components(self.components) if isinstance(q, Question)]
+
+    @property
+    def all_components(self) -> list["Component"]:
+        return get_ordered_nested_components(self.components)
 
 
 class SubmissionEvent(BaseModel):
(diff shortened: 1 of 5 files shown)


app/common/data/models.py
```diff
@@ -244,20 +244,23 @@ class Form(BaseModel):
...
         """Consistently returns all questions in the form, respecting order and any level of nesting."""
-        return get_ordered_nested_questions_for_components(self.components)
+        return [q for q in get_ordered_nested_components(self.components) if isinstance(q, Question)]
+
+    @property
+    def all_components(self) -> list["Component"]:
+        return get_ordered_nested_components(self.components)
 
 
 # todo: unit test reasonably extensively
```
samuelhwilliams, line 254: I feel like you've paid this down now, feel free to reap the rewards and drop the todo 🙇

app/deliver_grant_funding/routes/reports.py
```diff
@@ -396,12 +396,17 @@ def list_group_questions(grant_id: UUID, group_id: UUID) -> ResponseReturnValue:
         if not AuthorisationHelper.has_grant_role(grant_id, RoleEnum.ADMIN, user=get_current_user()):
             return redirect(url_for("deliver_grant_funding.list_group_questions", grant_id=grant_id, group_id=group_id))
 
-        if delete_wtform.validate_on_submit():
-            delete_question(group)
+        try:
+            raise_if_question_has_any_dependencies(group)
```
samuelhwilliams, line 400: Do we think this should be part of `delete_question`?

samuelhwilliams (approved):

samuelhwilliams (approved):

PR #684, opened 2025-08-27
BAU: Strip empty characters from email & url inputs

MarcUsher:
If the user input a valid email or URL but with spaces at the start/end while completing a form, the form would raise a validation error that these were in the incorrect format. We should strip empty spaces from the start and end of these question inputs so that the validation is checking the format of the actual string inputs and not caring about extraneous spaces.


samuelhwilliams (approved):
nit: maybe a tiny test for this

samuelhwilliams (approved):

PR #685, opened 2025-08-27
Add JSON export format and functionality

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-665

## 📝 Description
Adds a basic JSON export for submissions

### Review Readiness

### Testing


app/common/collections/types.py
```diff
@@ -110,6 +118,16 @@ def get_value_for_expression(self) -> int:
...
+    def get_value_for_json_export(self) -> dict[str, Any]:
+        data: dict[str, str | int] = {"value": self.value}
+
+        if self.prefix:
+            data["prefix"] = self.prefix
+        if self.suffix:
+            data["suffix"] = self.suffix
+
+        return data
+
```
MarcUsher, line 130: Strangely if you put `£` as a prefix what gets exported to the json is the unicode character string instead of `£` (at least for me in testing briefly just now)
image

app/common/helpers/collections.py
```diff
@@ -424,19 +438,50 @@ def generate_csv_content_for_all_submissions(self) -> str:
...
-                    submission_csv_data[header_string] = answer.get_value_for_text_export() if answer else NOT_ANSWERED
+                    submission_csv_data[header_string] = (
+                        answer.get_value_for_text_export() if answer is not None else NOT_ANSWERED
+                    )
 
             csv_writer.writerow(submission_csv_data)
 
         return csv_output.getvalue()
 
+    def generate_json_content_for_all_submissions(self) -> str:
```
MarcUsher, line 458: The TestCollectionHelper has a few CSV download test that check what's exported in terms of conditionals, skipped answers etc. - in some ways it feels like we can rely on a lot of that for the json download as it's following pretty much all the same steps but is it worth chucking one in just for sense checking?

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_submissions.html
```diff
@@ -41,6 +41,14 @@ <h1 class="govuk-heading-l">
             "disabled": true
           })
         }}
+        {{
+          govukButton({
+            "text": "Export as JSON",
+            "href": url_for("deliver_grant_funding.export_report_submissions", grant_id=grant.id, report_id=report.id, submission_mode=submission_mode, export_format='json'),
+            "classes": "govuk-button--secondary govuk-!-margin-bottom-2",
+            "disabled": true
```
MarcUsher, line 49: 2 questions on this:
1. do we want this to be disabled?
2. This doesn't actually seem to be doing anything (the button isn't disabled, and seemingly adding it to other buttons on other templates also doesn't disable them, so maybe something funky going on with govuk version/govuk jinja?)

Caveat - I might have misunderstood what this is meant to do

samuelhwilliams: nah I think you're on the money, removing these, good spot 👍

app/common/collections/types.py
MarcUsher: Doesn't do it for other special characters though it seems (though this is suffix, but that shouldn't make any difference)
image

app/common/helpers/collections.py
samuelhwilliams: I've added a single test on the JSON export stuff - the other CSV tests, for JSON equivalents, felt like it was reproducing a _lot_ of the core logic of the endpoint we were testing. Which to me doesn't feel like a super good use of time...

PR #765, opened 2025-09-12
FSPT-590: use accessible autocomplete for question conditions

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-590

## 📝 Description
Use the accessible autocomplete component instead of a standard select. This might be nicer when there are a lot of questions.


sfount (approved):
Approving this from a code point of view, I this this will be a nice step towards choosing the right question and I imagine the right question/ answer is often known to the form designer so being able to start typing it feels intuitive

A good thing to get feedback on, it would be good to get written down any other ideas discussed that would need more design work (likely not needed unless we get form designer feedback) like tweaking the existing question table to show the questions in relation to each other (i.e order and group relationships)

samuelhwilliams:
I do also wonder whether we should just show radios if there are <x options (similar to how we do it on form runner, but probably switching to autocomplete at a lower threshold).

I also think we probably should filter out the questions you can't select; might follow up to do that.

PR #774, opened 2025-09-16
Refactor ExpressionContext

samuelhwilliams:
## 🎫 Ticket

## 📝 Description
This was effectively a custom implementation of python's native ChainMap, which takes a set of dictionaries and essentially layers them on top of each other.

To reduce the amount of custom code we have, this strips it down to prefer using ChainMap.

When passing data to an expression, there are currently two things we care about:

- the latest view of data for the submission (answers that a user has provided to the questions in a report)
- the data attached to an expression itself

The latest view of data for a submission has two layers:

- the data submitted by a user in the current POST request 
- all of the data we have on the Submission model in the DB

Any data submitted in the current post request (generally just one question) should override an older answer on the Submission in the DB, so we check form data and then submission data.

The expression data is entirely separate, but because we're using ChainMap it needs to be considered either 'more important' or 'less important' than submission data. I've fairly arbitrarily chosen that the current order/preference is:

- submission data (answers to questions)
- expression data

These are not currently namespaced. We might want to namespace these in the future so that there is no chance of overlap at all.

### Review Readiness

### Testing

## 📚 Additional Notes
Are tests running a bit slower with this change? Maybe 🫠

### DIFF
--- a/app/common/collections/forms.py
+++ b/app/common/collections/forms.py
@@ -12,7 +12,6 @@ from govuk_frontend_wtf.wtforms_widgets import (
     GovTextArea,
     GovTextInput,
 )
-from immutabledict import immutabledict
 from wtforms import DateField, Field, Form, RadioField
 from wtforms.fields.choices import SelectField, SelectMultipleField
 from wtforms.fields.numeric import IntegerField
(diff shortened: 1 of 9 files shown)


srh-sloan (approved):
Nice! Once I'd got my head round it I agree it's simpler. Pulled and run locally, expressions working still 🎉

PR #805, opened 2025-09-24
Prevent breaking referential integrity between components

samuelhwilliams:
## 🎫 Ticket

## 📝 Description
Now that components (questions and groups) can include answers from other questions in their text, we need to make it impossible to invalidate that reference after it's been set up. One of the ways this is possible is by setting up a reference from q2 to q1, and then making q2 be shown to a user before q1.

This logic already exists in the function `_check_component_order_dependency`, which is called when moving components up and down.

By changing the logic for this function to scan our new ComponentReference table instead of checking just the conditions on each component, we can achieve this check in a more reliable way. ComponentReferences are created whenever a question refers to another one, either in its text or through its expressions (conditions and validation).

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/tests/models.py
+++ b/tests/models.py
@@ -612,9 +612,7 @@ class _QuestionFactory(SQLAlchemyModelFactory):
 
         # Wipe the cache of questions on a form - because we're likely to be creating more forms/questions
         del self.form.cached_questions
-
-        if create:
-            db.session.commit()
+        db.session.commit()
 
 
 class _GroupFactory(SQLAlchemyModelFactory):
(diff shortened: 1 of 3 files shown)


tests/models.py
```diff
@@ -668,6 +668,24 @@ def expressions(self, create: bool, extracted: list[Any], **kwargs: Any) -> None
...
+            ExpressionContext.build_expression_context(
+                collection=self.form.collection, fallback_question_names=True, mode="interpolation"
+            ),
+        )
+
+        # Wipe the cache of questions on a form - because we're likely to be creating more forms/questions
+        del self.form.cached_questions
+
+        if create:
+            db.session.commit()
```
sfount, line 687: tiniest nit ever on a fixture but this should always be true as this looks like a no-op at the top when its just building

srh-sloan (dismissed):
Nice simplifcation

sfount (approved):
Nice this table feels like its going to clean up a lot of this code 🏅

PR #806, opened 2025-09-24
Prevent using questions from later in the task

samuelhwilliams:
## 🎫 Ticket

## 📝 Description
When inserting a reference to some contextual data, we currently let you add a reference to a question that is displayed after the one being edited.

Referentially this doesn't make sense, so let's not list those questions in the select dropdown.

Once the reference is set up, other integrity checks will prevent it from being broken (eg by re-ordering questions)

### Edge case

If you are trying to insert a reference into the first question in the task, the dropdown list is just empty. This is a broken case. Open for thoughts on how to address this, we could:

1. Instead of showing the dropdown, show some text saying something like "There are no earlier questions in this task to reference."
2. Have the radio option on the "select data source" page greyed out? Or throw an error on submission saying there are no questions available?

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/models.py
+++ b/app/common/data/models.py
@@ -196,6 +196,9 @@ class Form(BaseModel):
         """Consistently returns all questions in the form, respecting order and any level of nesting."""
         return [q for q in get_ordered_nested_components(self.components) if isinstance(q, Question)]
 
+    def global_question_index(self, question: "Question") -> int:
+        return self.cached_questions.index(question)
+
     @cached_property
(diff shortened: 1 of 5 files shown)


srh-sloan (changes requested):
Just pulled locally and found a small bug:
- If I've got the first question in a form and try to add data, select 'repvious question in this task', then I get the error.
- If I do the same thing above whilst adding the first question (so before it really exists) then I can select 'previous question in this task' and get to the drop down, which is then empty.

PR #811, opened 2025-09-24
Catch manual out of order errors

samuelhwilliams:
## 🎫 Ticket

## 📝 Description
We've stopped form designers from inserting references to questions that are later in the form through the user interface, but they could also type them in manually. This patch will detect and validate any references that are manually included, preventing them from being saved and erroring appropriately.

## To test/reproduce

1. Open the edit question page for the first question of the 'Contact information' task for the Cheeseboard in parks seeded grant
2. Add `((q_b51759ce069049d19a0fd6f34eda28f0))` to the question text. This is a reference to the second question in that task.
3. Try to save it. Should be able to save this on main, and not save it on this branch.

### DIFF
--- a/app/deliver_grant_funding/forms.py
+++ b/app/deliver_grant_funding/forms.py
@@ -426,8 +426,10 @@ class AddContextSelectSourceForm(FlaskForm):
             available_questions = [
                 q
                 for q in self.form.cached_questions
-                if self.current_question is None
-                or self.form.global_question_index(q) < self.form.global_question_index(self.current_question)
+                if (
+                    self.current_question is None
+                    or self.form.global_component_index(q) < self.form.global_component_index(self.current_question)
+                )
             ]
             if not available_questions:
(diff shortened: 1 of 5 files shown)


app/deliver_grant_funding/forms.py
```diff
@@ -458,8 +460,10 @@ def __init__(
         self.question.choices = [("", "")] + [
             (str(question.id), interpolate(question.text))
             for question in form.cached_questions
-            if current_question is None
-            or form.global_question_index(question) < form.global_question_index(current_question)
+            if (
```
MarcUsher, line 463: These fixes are really nice qol for form designers - thinking we can maybe do the same in the question select dropdown for conditions as well to avoid people clicking on stuff that comes later (not saying you should do this as part of this PR, more commenting to remind myself "that's a thing I can do in a little 5m gap")

MarcUsher (approved):

PR #814, opened 2025-09-25
Redirect to group when editing a question in a group

samuelhwilliams:
At the moment if you save edits to a question which is inside a group, you get redirect to the top of the task/form list of questions. This navigation feels wrong.


MarcUsher (approved):
Yesss thank you, this is a v nice quality of life change!

One 🤏 q is should the save take you back to where you initiated the edit from (ie. if you clicked the group question from the task page, should it take you back to the task page?) But that could also start to get really gnarly with nested groups, so it always taking you back to the question group it's in makes sense I think, and overall is better behaviour than always going back to the task.

PR #815, opened 2025-09-25
Change check-your-answers page to full width

samuelhwilliams:
## 🎫 Ticket
N/A

## 📝 Description
The check your answers summary list looks overly cramped and dense when in a two-thirds column. Given that it it limits the width of each column to less than a standard two-thirds, we should be able to increase the width of this component to full so that the question and answers have a little bit more room to spread out, which should generally increase readability.

### Before
image

### After
image


MarcUsher (approved):
Yeah this is a good change I think, especially when multi-line text answers come in and start being pretty long.

PR #818, opened 2025-09-25
Updating env var to fix seg fault

srh-sloan:
Setting PGGSSENCMODE to disable when running pytest as it fixes a segmentation fault we saw after upgrading to MacOS Tahoe. The fix in this post seems to help: https://github.com/ged/ruby-pg/issues/538#issuecomment-1591629049

See also: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-GSSENCMODE


MarcUsher (approved):
🙏

PR #820, opened 2025-09-25
BAU: Don't allow questions to reference themselves

MarcUsher:
## 🎫 Ticket
BAU

## 📝 Description
When inserting data from other questions into a question's text, hint or guidance we don't surface the current question in the flow, but a user could manually add a reference to itself we weren't catching it.

This change just raises an error if a user tries to manually reference the current question, as this wouldn't make sense in the context of the form and would look broken for grant recipients.

## 🧪 Testing
On main when editing a question, manually reference the same question as context in the question text/hint/guidance and save the question. It should let you save.

Pull the branch and try to do the same again; it should now raise an error, stopping you from referencing the question in itself.

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -893,7 +893,7 @@ def _validate_and_sync_component_references(component: Component, expression_con
                     )
 
                 # Prevent manually injecting a reference to a question that appears later in the same form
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
Thank you - really good catch 🙇

PR #821, opened 2025-09-25
Remove param from `build_expression_context`

samuelhwilliams:
We always want to fall back on question names when interpolating expressions, and we can't fallback when evaluating them. So let's remove this redundant parameter to clean up the calls a little bit.


MarcUsher (approved):
🧹 💪

PR #824, opened 2025-09-25
Remove code use of data source item references

samuelhwilliams:
## 🎫 Ticket
[](https://mhclgdigital.atlassian.net/browse/FSPT-821)

## 📝 Description
This stops all of our code from using the DataSourceItemReference table in preparation for deleting it altogether. All code now inspects the ComponentReference table instead, looking at depends_on_data_source_item_id as needed.

When we merge and deploy this, we'll be able to follow up with a PR that drops the table altogether. It's possible that between this one and that one going out, we won't be able to fully manage some existing questions because SQLAlchemy won't know to delete DataSourceItemReferences, so we won't be able to delete data source items.

We can manage this fine for the moment because there aren't any users. In a production use case we'd have to do this differently.

### Review Readiness

### Performance and security

### Testing


MarcUsher (approved):
Nice this works as expected for me locally - the seed command seems to have correctly synced all the references, and new references from conditions (on data source items), validations and adding contextual data all create the expected rows in ComponentRef and nothing in data source item ref 👍 

Grant seeding fails for me with this PR but we know this and expect it should be fine with both PRs combined 🤝

PR #828, opened 2025-09-26
Adding E2E tests for nested groups

srh-sloan:
## 🎫 Ticket
[FSPT-759 Nested groups: E2E tests](https://mhclgdigital.atlassian.net/browse/FSPT-759)

## 📝 Description
E2E tests for the above ticket:
- Adds a new nested group to the existing 'one per page' group
- Updates functions for completing and viewing answers to recurse through the list of questions to account for nested groups


tests/e2e/test_create_preview_collection.py
```diff
@@ -558,12 +592,14 @@ def task_check_your_answers(
...
-        else:
-            assert_check_your_answers(check_your_answers_page, question_to_test)
+    def _assert_check_your_answers(questions_to_check: list[TQuestionToTest]):
+        for question_to_check in questions_to_check:
+            if question_to_check["type"] == "group":
+                _assert_check_your_answers(question_to_check["questions"])
+            else:
+                assert_check_your_answers(check_your_answers_page, question_to_check)
+
+    _assert_check_your_answers(list(questions_to_test.values()))
```
samuelhwilliams, line 602: Can we move this outside of the function somewhere? And give it a better name? Having a function called `assert_view_report_answers` and `_assert_view_report_answers` feels a bit confusing. I think we probably want to take a step back and refactor to make sure things are feeling simple and clean and clear.

samuelhwilliams (approved):
Looks good and nice additional coverage for groups - small suggestion on refactoring 👍

tests/e2e/test_create_preview_collection.py
srh-sloan: Yeah, fair point. I think add another will further complicate so will look to refactor then?

PR #832, opened 2025-09-29
FSPT-759 - nested groups integrity tests

srh-sloan:
Writing integration tests for nested groups to check integrity rules:
- on moving a question above/below a nested group with dependencies
- on deleting a question with dependencies in a nested group

### DIFF
--- a/tests/integration/common/data/interfaces/test_collections.py
+++ b/tests/integration/common/data/interfaces/test_collections.py
@@ -39,6 +39,7 @@ from app.common.data.interfaces.collections import (
     move_form_up,
     raise_if_data_source_item_reference_dependency,
     raise_if_group_questions_depend_on_each_other,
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
Nice additional tests 💪 

Small optional one - I find it a nicer developer experience if we can wrap logically-related tests in a class so that it's easier to run all of the tests about some larger unit together; can we wrap some of these inside a `class TestNestedGroups` - or `TestDeleteQuestion` or `TestMoveComponent` or something? Whichever one feels appropriately - I think generally I name the class on the function or view being tested, so probably TestMoveComponent, TestDeleteQuestion, TestRaiseIfNestedGroupCreationNotValidHere feel right?

samuelhwilliams (approved):

PR #834, opened 2025-09-29
FSPT-852 Add Another - DB Migration and model changes

srh-sloan:
## 🎫 Ticket
[Add Another - DB migration](https://mhclgdigital.atlassian.net/browse/FSPT-852)

## 📝 Description
First small part of the [add another ticket ](https://mhclgdigital.atlassian.net/browse/FSPT-849). This PR adds the column `add_another` to the Component table, with a property `add_another_container` to determine the containing component of this question if it exists in an add another context.

- If a single question is an add another question, the container is itself
- If a question or group is contained inside an add another group, the container is that add another group.

We are implementing a restriction to stop an add another group containing any further add anothers (groups or questions) so we also add a function to validate whether an add another component can be created.

All of the above is tested. There is no visible change to users or the journey, and nothing uses the new code or column yet.

## 🧪 Testing
- Unit and integration tests added, and factories updated

### Review Readiness
- [ x] PR title and description provide sufficient context for reviewers
- [x ] Commits are logical, self-contained, and have clear descriptions

### Performance and security
- [ x] No N+1 query problems introduced
- [x ] Any new DB queries include appropriate where clauses based on the user's permissions

### DIFF
--- a/tests/integration/common/data/interfaces/test_collections.py
+++ b/tests/integration/common/data/interfaces/test_collections.py
@@ -6,6 +6,7 @@ from sqlalchemy.exc import IntegrityError, NoResultFound
 from app.common.collections.types import TextSingleLineAnswer
 from app.common.data.interfaces import collections
 from app.common.data.interfaces.collections import (
+    AddAnotherNotValidException,
     DataSourceItemReferenceDependencyException,
     DependencyOrderException,
     NestedGroupDisplayTypeSamePageException,
@@ -37,6 +38,7 @@ from app.common.data.interfaces.collections import (
     move_component_up,
     move_form_down,
     move_form_up,
+    raise_if_add_another_not_valid_here,
     raise_if_data_source_item_reference_dependency,
     raise_if_group_questions_depend_on_each_other,
     raise_if_nested_group_creation_not_valid_here,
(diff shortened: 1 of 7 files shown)


tests/integration/common/data/interfaces/test_collections.py
```diff
@@ -2501,3 +2503,109 @@ def test_works_with_groups(self, db_session, factories):
...
+
+    def test_raise_if_add_another_not_valid_does_not_raise_for_non_add_another_nested_group_question(
+        self, db_session, factories
+    ):
+        form = factories.form.create()
+        group = factories.group.create(form=form, add_another=False)
+        group4 = factories.group.create(form=form, parent=group, add_another=False)
+        question3 = factories.question.build(form=form, parent=group4, add_another=False)
+        raise_if_add_another_not_valid_here(question3)
+        group5 = factories.group.build(form=form, parent=group4, add_another=False)
```
MarcUsher, line 2578: It's the smallest nit so feel free to ignore but the fact these group/question numbers run on from the numbering in the previous test and don't reset for this test was initially confusing to read (especially starting with just `group`) but it's also a v short test and doesn't take long to clock 😆

MarcUsher (approved):
Praise be the ancestor parents 😆 this all seems sensible and logical. Tests look great too, made it v easy to see the scope and boundaries of the change and where the exception expects to be raised 🙏

tests/integration/common/data/interfaces/test_collections.py
srh-sloan: Thanks, that's because I think I had this as one test then split it out. Fixed tho as this is complicated enough without making it harder to read! :)

MarcUsher (approved):

PR #835, opened 2025-09-30
FSPT-853 Add Another index for update_submission_data

srh-sloan:
## 🎫 Ticket
[FSPT-853](https://mhclgdigital.atlassian.net/browse/FSPT-853)

## 📝 Description
This builds on PR #834 and enables the `update_submission_data` function to work with add another questions.

`update_submission_data` now takes in an optional `add_another_index` to determine which item in a group of add another responses to add/update.

For example, if there are 2 questions in an add another group as follows:

- Contact details group (add another = True), ID: `111`
    - What is the contact's name? ID: `222`
    - What is the contact's email? ID: `333`

With 2 existing responses, so the data looks like this:
```
{
    111: [
        {
            222: "name 1",
            333: "email1@domain.com"
        },
        {
            222: "name 2",
            333: "email2@domain.com"
        }
    ]
}
```

Then to update the name (question id `222`) for 'name 2', you would pass `add_another_index=1`.

So `update_submission_data` has been updated to handle the `add_another_index`, with appropriate validation, and this PR adds unit tests for that function.

🔍  Nothing to see here - still just backend stuff that's not used from the UI

## 🧪 Testing
Unit tests, including and update the Collection factory to create a submission that uses a nested add another group. Although this is only used once at present in the tests, it feels like something that will be needed more as this work on add another continues.

### Review Readiness

### Performance and security

## 📚 Additional Notes
- Currently if validation of the `add_another_index` fails, a generic `ValueError` is raised. This may be changed into a custom exception in future PRs if we think we will be surfacing this to users. Still tbd by the further work to come.

[FSPT-853]: https://mhclgdigital.atlassian.net/browse/FSPT-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/common/data/interfaces/test_collections.py
+++ b/tests/integration/common/data/interfaces/test_collections.py
@@ -2827,12 +2827,12 @@ class TestAddAnother:
         self, db_session, factories
     ):
         form = factories.form.create()
-        group = factories.group.create(form=form, add_another=False)
-        group4 = factories.group.create(form=form, parent=group, add_another=False)
-        question3 = factories.question.build(form=form, parent=group4, add_another=False)
-        raise_if_add_another_not_valid_here(question3)
-        group5 = factories.group.build(form=form, parent=group4, add_another=False)
-        raise_if_add_another_not_valid_here(group5)
+        group1 = factories.group.create(form=form, add_another=False)
+        group2 = factories.group.create(form=form, parent=group1, add_another=False)
+        question1 = factories.question.build(form=form, parent=group2, add_another=False)
+        raise_if_add_another_not_valid_here(question1)
+        group3 = factories.group.build(form=form, parent=group2, add_another=False)
+        raise_if_add_another_not_valid_here(group3)
 
     def test_raise_if_add_another_not_valid_does_not_raise_for_valid_question(self, db_session, factories):
         form = factories.form.create()


tests/integration/common/data/interfaces/test_collections.py
```diff
@@ -1862,6 +1862,44 @@ def test_update_submission_data_add_another_fail_if_index_greater_than_available
...
+        )
+        update_submission_data(submission, add_another_name_question, TextSingleLineAnswer("Added name 6"), 5)
+        updated_submission = update_submission_data(
+            submission, add_another_email_question, TextSingleLineAnswer("Added_email_6@email.com"), 5
+        )
+
+        updated_data = updated_submission.data[str(add_another_group.id)]
+
+        # check the name and email of entries we didn't change are still the same
+        for i in [0, 1, 3, 4]:
```
MarcUsher, line 1888: If someone ever changes the `create_completed_submissions_add_another_nested_group` method I could see us relying on specific indexes here could lead to tests failing, but also that's probably quite unlikely as that method is very specifically for this use case and if we need to extend it/have add another groups more regularly across the board then we can create another post_generation method for a fuller submission.

MarcUsher (approved):
Really good test coverage again 🙏  makes it v easy to understand how this is expected to be used and the edge cases you're protecting against

Agree that maybe a more specific error might be helpful later down the line, especially once we're surfacing it in the UI (if something goes wrong with the form runner and add another then the end user won't understand much about the indexes) but agree it's good for now!

tests/integration/common/data/interfaces/test_collections.py
srh-sloan: Yeah that's a good shout, I might change it as I get more into this and if we need to extend the factory method.

PR #843, opened 2025-10-02
Wait for this autocomplete to initialise

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
e2e tests are currently failing on this page as they're skipping filling in the autocomplete. I can see from playwright traces that it's looking for the `data-accessible-autocomplete` attribute and then just immediately moving on, so it's not finding anything with that attribute. We don't need to look for it here - it's guaranteed to be there. So we can just wait for the JS to have run directly.

image


MarcUsher (approved):
Let's goooo

PR #861, opened 2025-10-07
Adding formatting of scss files to pre-commit

srh-sloan:
- Adds 2 new make targets:
    - format-css: runs prettier on our scss files to update them with standard formatting rules
    - format-frontend: runs both the new format-css target and the existing format-html to format all our frontend files

Also updates our scss files to use this new formatting so future PRs won't introduce whitespace changes, making it harder to review

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ci:
   autofix_prs: false
-  skip: ["mypy", "format-html"]
+  skip: [ "mypy", "format-frontend" ]
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
@@ -14,7 +14,7 @@ repos:
     rev: v0.13.3
     hooks:
       - id: ruff
-        args: [--fix]
+        args: [ --fix ]
       - id: ruff-format
   - repo: https://github.com/Yelp/detect-secrets
     rev: v1.5.0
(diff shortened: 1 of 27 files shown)


MarcUsher (commented):
Nice! Thanks for doing this Sarah 🙏 

Might be worth splitting the formatting changes into a separate commit and adding that commit hash to the `git-blame-ignore-revs`?

.pre-commit-config.yaml
```diff
@@ -40,12 +40,12 @@ repos:
...
         entry: uv run mypy
         pass_filenames: false
-      - id: format-html
-        name: Format Jinja HTML templates with Prettier
-        entry: make format-html
+      - id: format-frontend
+        name: Format Jinja HTML templates and CSS files with Prettier
+        entry: make format-frontend
         language: node
         files: \.html$
```
samuelhwilliams, line 50: Do we need to change this to include scss/sass/css/js as well?

Makefile
```diff
@@ -28,6 +28,14 @@ check-html:
 format-html:
 	npx prettier --plugin=prettier-plugin-jinja-template --parser=jinja-template --tab-width=2 --html-whitespace-sensitivity css --bracket-same-line=true --print-width=240 --write **/*.html
 
+.PHONY: format-css
+format-css:
+	npx prettier --tab-width=4 --write **/*.scss
+
+.PHONY: format-frontend
+format-frontend: format-css format-html
```
samuelhwilliams, line 36: Should we be formatting js files as well? We only have one but it's currently stored with 2 spaces, so immediately feels a little out of sync. Would be good to know that js will just be taken care of if/when we start writing more.

samuelhwilliams (approved):
lovely 👏

samuelhwilliams (approved):

PR #865, opened 2025-10-07
Bugfix: Update Guidance template for inserting context

MarcUsher:
## 📝 Description
As part of the refactor in https://github.com/communitiesuk/funding-service/pull/858 we updated how inserting contextual data for question text/hint and guidance worked to consistently extract the entire form data and rebuild it when returning from the insert data flow.

We missed some small updates to initiating this flow from the manage_guidance pages, which meant clicking the `Insert data` button didn't actually kick off the flow as expected.


samuelhwilliams (approved):
goooood catches!

PR #891, opened 2025-10-15
FSPT-907 Add another context for next/ previous helpers

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-907

## 📝 Description
The final helpers that might be specific to your current position in the form are get next/ previous question. As this is a "flat" representation of visible questions it should be updated to set an add another context if provided by the consumer - this means questions in an add another group will be able to conditionally reference other questions in that group and continue to work consistently for routing.

The existing expression set up will continue to be evaluated to check for visibility even when theres no given add another context, for that reason we'll need to namespace the full list of add another question answers in the expression context - that will mean if theres an add another context the question ID will have relevant answers and otherwise expressions will need to use the namespace if using aggregate method (a future anticipated feature!)

Allows updating the context for next/ previous if you're in a specific entry for an add another question.

### Review Readiness

### Performance and security

### Testing
- [x] New (non-developer) functionality has appropriate unit and integration tests (in progress)

### DIFF
--- a/tests/unit/common/helpers/test_collections.py
+++ b/tests/unit/common/helpers/test_collections.py
@@ -249,19 +249,22 @@ class TestSubmissionHelper:
             assert question.id == question_three.id
 
         def test_next_question_with_add_another_index(self, factories):
-            group = factories.group.build(add_another=True, order=0)
+            form = factories.form.build()
+            q0 = factories.question.build(form=form, order=0)
+            group = factories.group.build(form=form, add_another=True, order=1)
             q1 = factories.question.build(parent=group, data_type=QuestionDataType.INTEGER, order=0)
             q2 = factories.question.build(parent=group, order=1)
             q3 = factories.question.build(parent=group, order=2)
-            q4 = factories.question.build(form=group.form, order=1)
+            q4 = factories.question.build(form=form, order=2)
 
-            submission = factories.submission.build(collection=group.form.collection)
+            submission = factories.submission.build(collection=form.collection)
             factories.expression.build(question=q2, type_=ExpressionType.CONDITION, statement=f"{q1.safe_qid} > 50")
 
             submission.data = {str(group.id): [{str(q1.id): {"value": 55}}, {str(q1.id): {"value": 20}}]}
 
             helper = SubmissionHelper(submission)
 
+            assert helper.get_next_question(q0.id).id == q1.id
             assert helper.get_next_question(q3.id).id == q4.id
 
             # without any information to go with, the default behaviour is for the condition to fail closed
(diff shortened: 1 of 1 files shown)


app/common/expressions/__init__.py
```diff
@@ -251,6 +251,11 @@ def is_valid_reference(self, reference: str) -> bool:
 
         return True
 
+    def __hash__(self) -> int:
+        # separate immutable instances are used when a context is extended so the instance id
+        # should be sufficient for the lru_cache
+        return hash(id(self))
```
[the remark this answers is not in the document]
sfount: The implication for using the instance (memory) ID is that if a developer in the future mutates the existing context object in-place the cached response would still point to the previous answer (based on old context) - we should always be using the ".with_x" method of returning an instance of a new context (i.e with_add_another_context). This could be a footgun for future devs - we might want to revisit.

We briefly discussed that in a future refactor we probably want to structure the "visible" questions so that whenever you're dealing with the state of the schema you can interrogate which questions are visible at which entry (currently it "flattens" this to give it to you at a specific index - which works with all of the other interfaces/ methods we currently have).

app/common/helpers/collections.py
```diff
@@ -405,13 +405,26 @@ def toggle_form_completed(self, form: "Form", user: "User", is_complete: bool) -
                 self.submission, SubmissionEventKey.FORM_RUNNER_FORM_COMPLETED, form
             )
 
-    def get_next_question(self, current_question_id: UUID) -> Optional["Question"]:
+    # todo: decide if the add another index should be available submission helper wide where it just checks self
+    #       does having it on lots of methods increase the cognitive load/ complexity
```
srh-sloan: I actually think having it here in the signature reduces complexity as it's very clearly telling this method which one to use

app/common/helpers/collections.py
```diff
@@ -420,12 +433,25 @@ def get_next_question(self, current_question_id: UUID) -> Optional["Question"]:
 
         raise ValueError(f"Could not find a question with id={current_question_id} in collection={self.collection}")
 
-    def get_previous_question(self, current_question_id: UUID) -> Optional["Question"]:
+    def get_previous_question(
```
srh-sloan: This and `get_next_question` look almost identical, apart from the order of the iterator. Could we combine them somehow?

srh-sloan (commented):
Walked through this in a huddle, all makes sense. Couple of comments around duplicated logic and adding one more edge case test

app/common/helpers/collections.py
sfount: Yeah I considered having something to consistently manage making the right context, I don't think they need to ever be extended (i.e a different iterator used if we made that a variable) so I think I landed on it being OK for now and something we could refactor if we end up with a more consistent way of dealing with add another contexts

sfount: We think we'll streamline this with a decorator that manages the context overrides for helper methods that need it. Plan to do this as a refactor once we've got all the functionality/ consumers of these methods in!

PR #892, opened 2025-10-16
Make grants 'owned by' an org (MHCLG)

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-913

## 📝 Description
We're about to split our current "platform admin" user management into two separate roles:

- true platform admins (ie developers)
- people who manage grants, eg creating new grants and setting up reports (ie form designers)

In order to support this split, we need to tweak what permissions define each of these people and what they can do.

True platform admins will retain the permission: org=None, grant=None, role=ADMIN.

Form designers will now get the permission org=<something>, grant=None, role=ADMIN. In order to distinguish between grant recipient org admins and grant-owning org admins, we add a new property to the Grant table: `can_manage_grants`. If your permission is linked to an org that can manage grants, you can access Deliver grant funding. If your permission is linked to an org that cannot manage grants, you can use Access grant funding.

Theoretically we should also support org=<grant managing org>, grant=<something>
- which would (in lieu of other perms) let you edit and manage reports for just that specific grant, but not create new grants.

This patch prepares us for the above, in that it will let us create organisations, set them as 'grant managing orgs', and associate existing grants with those orgs.

For now, we only expect to have a single grant-managing org: MHCLG. And we enforce that with a DB constraint, which lets us avoid a number of design and technical considerations around Deliver grant funding needing to understand multiple grant-owning orgs. We expect this product request to come soon, and so this technically prepares us for that future, but it pushes back any handling of the edge cases until we approach that feature holistically as a team.

---

I've also added two bulk actions to our platform admin pages to help transition existing grants (to make them owned by MHCLG) and users (to give them form designer permissions).

image

image

image

## 🧪 Testing
Manual

### Review Readiness

### Testing

### DIFF
--- a/scripts/run-ad-hoc-task.py
+++ b/scripts/run-ad-hoc-task.py
@@ -55,7 +55,15 @@ run_task_response = ecs_client.run_task(
     networkConfiguration={
         "awsvpcConfiguration": {"subnets": subnets, "securityGroups": security_groups, "assignPublicIp": "DISABLED"}
     },
-    overrides={"containerOverrides": [{"command": command, "name": "task-runner-container"}]},
+    overrides={
+        "containerOverrides": [
+            {
+                "command": command,
+                "name": "task-runner-container",
+                "environment": [{"name": "SEED_SYSTEM_DATA", "value": "false"}],
+            }
+        ]
+    },
 )
 
 task_arn = run_task_response["tasks"][0]["taskArn"]


MarcUsher (approved):
This is really comprehensive and clear - the guards/warnings around seeding more things in this way make a lot of sense, and this also enforcing the difference in using create & build in integration & unit tests seems like a good thing to have.

Approving but it's not a small boy so maybe also worth getting Sarah/Steven to take a look in case I've missed anything? 

Pulled the branch and it works for me locally though, and I can run the actions in the admin pages.

app/common/data/migrations/versions/007_add_grant_to_org_fk.py
```diff
@@ -0,0 +1,44 @@
...
+
+revision = "007_add_grant_to_org_fk"
+down_revision = "006_add_another"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("grant", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("organisation_id", sa.Uuid(), nullable=True))
```
sfount: I'm guessing this will just be for a future tickets A/C (along with the dev and design work you mention in commits for how this is set up) to make this required in the future

.github/workflows/build_and_test.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         DATABASE_PORT: 5432
         DATABASE_NAME: postgres
         DATABASE_SECRET: '{"username":"postgres","password":"password"}'  # pragma: allowlist secret
+        SEED_SYSTEM_DATA: false
```
sfount: I suspect this will also need to be set in the jobs that run migrations in environments which probably assume a truthy value if they don't set an explicit test env (which I hope they don't!)

PR #895, opened 2025-10-16
FSPT-850 Generate a single line summary for a given add another question group

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-850

## 📝 Description
When there are multiple questions this helper can be used to generate a single line summary of all of their answers - as this is just stored in presentation options for now it might contain references to old questions so the helper is set up to default to returning a summary of everything if there is either no configuration or no valid configuration.

This will be used for a consistent header/ summary row of add another entries on the add another summary and check your answers pages (maybe data exports but not confirmed).

This proposes configuring this in the containing groups presentation options which would then be configurable by the form designer - this trades of the database integrity of adding this as a separate table + relationship with having this configuration loaded whenever we load the group being used.

Any questions that have been removed from the group whenever this is applied will just be ignored and no relevant questions in the configuration will result in all the answers used for the summary (which is the default option if the form designer has selected no questions).

When the presentation options are being configured, all of the groups questions will be shown and selected if there are no current presentation options or if the existing presentation options aren't relevant. If relevant question ids are stored in the presentation options only those will be selected.

### Review Readiness

### Testing

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -6,7 +6,7 @@ from datetime import datetime
 from functools import cached_property, lru_cache, partial
 from io import StringIO
 from itertools import chain
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union, cast
 from uuid import UUID
 
 from pydantic import BaseModel as PydanticBaseModel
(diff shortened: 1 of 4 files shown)


app/common/helpers/collections.py
```diff
@@ -435,6 +435,30 @@ def get_previous_question(self, current_question_id: UUID) -> Optional["Question
...
+                        pass
+            questions = questions or cast("Group", component.add_another_container).cached_questions
+        else:
+            questions = [cast("Question", component)]
+
+        answers = []
+        for question in questions:
+            answer = self.cached_get_answer_for_question(question.id, add_another_index=add_another_index)
+            if answer:
+                answers.append(answer.get_value_for_text_export())
```
samuelhwilliams: 🤏  tiny wonder if this should be rendering the answer template instead, so that it shows up consistent with how these answers display otherwise in eg check your answers or submission pages?

Not totally sure it's compatible - but would be worth exploring that, I think, if you haven't already?

PR #900, opened 2025-10-17
Allow creating invitations for Deliver org admins

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-914

## 📝 Description
This will let platform admins invite users to become org admins for MHCLG, which is the permission we'll use for (internal-to-funding-service) form designers for now.

Form designers receive a new email that has been 🤏  adapted from the existing MHCLG grant team member invitation email: https://www.notifications.service.gov.uk/services/239747da-5aa1-4fe0-85ab-39d0272ca5c8/templates/fd143e8b-c735-4e12-9eb5-1655724216d5

### Email
image

### Review Readiness

### Testing


sfount (approved):
Code and configuration looks good to me 👍 I've pulled locally and verified it all works as expected

A few quick thoughts on what might be refinements to the process, I can't speak to how challenging they'd be to configure though and very happy for them to be optional or join on them:
- either making the invite page specific to FS org admins (and removing the ability to associate a grant) or splitting it out into different types of invites that either require or don't need a grant I think would help
- question on if we plan to be using this or the "Make user org admin" button on the user list page (or both)
- it took me a while to realise the way to "Create" a user was to go to invite and create an invite that wasn't associated with a user - I think you would learn that once and then be sorted but maybe theres some hint text or action links between pages that can help that association

PR #902, opened 2025-10-18
Switch to using FS_PLATFORM_ADMINS group

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-910

## 📝 Description
We now have a new Azure AD group specifically for platform admins on the new Funding Service platform. Previously we had an FSD_ADMINS group, but this was for more than just platform admins - it's also for form designers or any generic
  'funding service' staff/users.

We want to very tightly lock down access to the platform admin panel of the integrated service, so this new group will represent access to that.

All other users will be set up within the service itself, using the service's authorisation models (user/userrole).

After this is merged, anyone with the 'platform admin' permission in the system will have it stripped away if they login and aren't a member of FS_PLATFORM_ADMINS. This will be quite a few of our users; we can manually migrate them using the platform admin user role page.

### Todo

 - [x] After merging/releasing this, we need to run the 'make mhclg form designer' action in flask-admin for all of the relevant people.
 - [ ] And then raise a PR removing the two temporary actions from flask-admin.

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -125,7 +125,7 @@ def sso_get_token() -> ResponseReturnValue:
     user = interfaces.user.get_user_by_azure_ad_subject_id(azure_ad_subject_id=sso_user["sub"])
 
     # Platform admin route
-    if "FSD_ADMIN" in sso_user.get("roles", []):
+    if "FS_PLATFORM_ADMIN" in sso_user.get("roles", []):
         user = interfaces.user.upsert_user_and_set_platform_admin_role(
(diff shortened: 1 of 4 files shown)


MarcUsher (approved):
Noice! Yeah good shout on proactively making other FS users org admins to avoid a flurry of "why am I getting access denied now?".

I think test will have a few people who've done UR still in the DB as 'users' but with no roles, but I guess we can just leave them there as is.

PR #908, opened 2025-10-20
FSPT-918: use google registry mirror

gidsg:
## 🎫 Ticket
N/A BAU

## 📝 Description
Docker Hub outage has drawn attention that the build will fail currently if docker hub is down.

This adds a mirror from google to use as an alternative before falling back to the docker hub default

See example: https://github.com/communitiesuk/funding-service/actions/runs/18647612236/job/53158173566#step:7:22

## 🧪 Testing
Have tested as per above example

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -63,6 +63,7 @@ jobs:
           ECR_IMAGE_URI="$AWS_ACCOUNT_ID.dkr.ecr.$AWS_REGION.amazonaws.com/$REPO_NAME:$IMAGE_TAG"
           echo "::add-mask::$ECR_IMAGE_URI"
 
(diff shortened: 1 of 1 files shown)


app/assets/css/context-aware-editor/_index.scss
```diff
@@ -163,8 +163,8 @@
 
 // When toolbar is disabled, ensure proper spacing
 .app-context-aware-editor--wrapper:not(
-    .app-context-aware-editor--with-toolbar
-) {
+        .app-context-aware-editor--with-toolbar
+    ) {
```
sfount: Not sure how these snuck in here but it doesn't look relevant to this change and isn't what the formatter wants things to do - could you check this file out of main?

PR #912, opened 2025-10-21
FSPT-923: Set a group to be add another

srh-sloan:
## 🎫 Ticket
[FSPT-923](https://mhclgdigital.atlassian.net/browse/FSPT-923)

## 📝 Description
First part of the form builder work to allow a group to be answered more than once (add another).
- Adds displaying the current 'add another' status to the group list questions page
- Allows the user to change this after the group is created
- Prevents the user from editing this setting if the group already contains add another components (and validates this in the backend incase they bypass the UI somehow)
- Prevents the user from changing add another if later questions depend on anything in this group

To come in a later PR:
- add the add another option to the create flow
- e2e tests

### After
Display add another status:
Screenshot 2025-10-21 at 14 13 42

Change add another status:
Screenshot 2025-10-21 at 14 13 48

Prevent changing:
Screenshot 2025-10-21 at 14 13 29

### Review Readiness

### Testing

## 📚 Additional Notes
The form runner is not yet setup to work with add another groups, so if you set a group to be add another and then preview that task, you will get an error

[FSPT-923]: https://mhclgdigital.atlassian.net/browse/FSPT-923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/common/data/models.py
+++ b/app/common/data/models.py
@@ -486,15 +486,17 @@ class Group(Component):
         based on the maximum number of levels of nested groups"""
         return bool(self.nested_group_levels < current_app.config["MAX_NESTED_GROUP_LEVELS"])
 
-    @cached_property
+    @property
     def contains_add_another_components(self) -> bool:
+        """Whether or not this group contains any components that have add_another set to True"""
         for component in self.cached_all_components:
             if component.add_another:
                 return True
         return False
 
-    @cached_property
+    @property
     def contains_questions_depended_on_elsewhere(self) -> bool:
+        """Whether or not any questions in this group (or nested groups) are depended on elsewhere"""
         for component in self.cached_all_components:
             if len(component.depended_on_by) > 0:
                 return True
--- a/app/deliver_grant_funding/forms.py
+++ b/app/deliver_grant_funding/forms.py
@@ -237,7 +237,9 @@ class GroupAddAnotherOptionsForm(FlaskForm):
         ],
         default="no",
         validators=[
-            DataRequired("Select whether people be able to answer all questions in this question group more than once")
+            DataRequired(
+                "Select whether people should be able to answer all questions in this question group more than once"
+            )
         ],
         widget=GovRadioInput(),
     )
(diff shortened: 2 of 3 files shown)


app/common/data/models.py
```diff
@@ -486,6 +486,21 @@ def can_have_child_group(self) -> bool:
...
+        for component in self.cached_all_components:
+            if component.add_another:
+                return True
+        return False
+
+    @cached_property
+    def contains_questions_depended_on_elsewhere(self) -> bool:
+        for component in self.cached_all_components:
+            if len(component.depended_on_by) > 0:
+                return True
```
samuelhwilliams: Do these need to be cached? I think we should only cache things that are _needed_, because more caching means more having to remember to cache bust things and potential sources of hard-to-track-down bugs. These are using cached calls so I think, unless these are getting called 100s+ times a request, we should not cache them.

We also have prefixed cached things with `cached_` to help make that clear, so we should do that if we _need_ these to be cached.

app/common/data/interfaces/collections.py
```diff
@@ -599,6 +622,32 @@ def as_flash_contexts(self) -> list[dict[str, str | bool]]:
...
+        contexts = self.as_flash_contexts()
+        return contexts[0] if contexts else {}
+
+    def as_flash_contexts(self) -> list[dict[str, str | bool]]:
+        flash_contexts = []
+        flash_context: dict[str, str | bool] = {
+            "message": self.message,
+            "group_name": self.group.name,
+            "group_id": str(self.group.id),
+            "grant_id": str(self.group.form.collection.grant_id),
```
samuelhwilliams:
```suggestion
            "grant_id": str(self.form.collection.grant_id),
```

I think we can always go directly to the form?

samuelhwilliams (approved):
Nice :) (other than unit test failures 🫣

samuelhwilliams (approved):

PR #918, opened 2025-10-22
Remove old admin actions; revoke permissions+invitations

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
At the moment we can only revoke invitations by deleting them entirely. It'd be nice to retain the record that an invitation was created, but still make it unusable. We can do this by moving the `expires_at_utc` date forward.

Also makes it possible to remove all permissions from a user. This is a catch-all and we have more fine-grained ability to do this from the User Role tab.

### Review Readiness

### Testing

### DIFF
--- a/app/deliver_grant_funding/admin/entities.py
+++ b/app/deliver_grant_funding/admin/entities.py
@@ -1,18 +1,18 @@
 import datetime
 from abc import abstractmethod
 
-from flask import flash
+from flask import current_app, flash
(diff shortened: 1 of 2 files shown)


MarcUsher (commented):
Nice! This all works well for me locally, invitations cancelled and user permissions removed as expected 🙏  some lil nits below

MarcUsher (approved):

PR #944, opened 2025-10-27
Set up organisations through platform admin

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-937

## 📝 Description
We will be loading our list of organisations by syncing with Delta's. Initially we need to do this through a manual bulk upload as Delta aren't regularly exporting or sharing their list with us yet and we need to meet deadlines. When they start exporting we'll have a nightly scheduled task.

This PR adds columns to the organisation table that we'll need and updates the platform admin CRUD views.

It's unclear what we need to do when an organisation is 'retired', so we log an error for Sentry to catch that will flag when this happens. Then we can investigate as a team and take the appropriate action.

### Review Readiness

### Testing

### DIFF
--- a/app/common/data/migrations/versions/010_add_org_columns.py
+++ b/app/common/data/migrations/versions/010_add_org_columns.py
@@ -1,0 +1,70 @@
+"""update organisations
+
+Revision ID: 010_add_org_columns
+Revises: 009_add_grant_status
+Create Date: 2025-10-27 10:38:16.378678
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+revision = "010_add_org_columns"
+down_revision = "009_add_grant_status"
+branch_labels = None
+depends_on = None
+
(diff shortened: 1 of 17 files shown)


app/common/data/migrations/versions/010_add_org_columns.py
```diff
@@ -0,0 +1,102 @@
...
+                    "METROPOLITAN_DISTRICT",
+                    "LONDON_BOROUGH",
+                    "SHIRE_COUNTY",
+                    "COMBINED_AUTHORITY",
+                    "NORTHERN_IRELAND_AUTHORITY",
+                    "SCOTTISH_UNITARY_AUTHORITY",
+                    "WELSH_UNITARY_AUTHORITY",
+                    name="organisationtype",
+                    create_type=False,
+                ),
```
MarcUsher: If you created the enums already above, do you need to relist out all the enum options here or is it enough to just pass the name and create_type?

PR #955, opened 2025-10-30
FSPT-908 Add toggle add another to question group create flow

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-908

## 📝 Description
Lets the form designer select if a question group should be asked multiple times while creating it.

The one complication here is that they shouldn't be able to create a question group set to be add another inside an existin add another question group - for now we'll just avoid asking the add another question in that case.


samuelhwilliams (approved):
Trio reviewed on Slack, Steven has left our comments, otherwise looks good to get in to unlock UR today

PR #959, opened 2025-10-30
Prevent users inserting references to same-page questions

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
When a user is creating a question within a same-page group, we were showing them all of the questions in that same-page group in the 'reference data from a previous question in this section' dropdown.

These references are invalid: you cannot reference data from a question that shows on the same page, because that question hasn't been answered yet.

We should exclude these questions from the dropdown.

### Review Readiness

### Testing

### DIFF
--- a/app/common/forms/helpers.py
+++ b/app/common/forms/helpers.py
@@ -1,10 +1,10 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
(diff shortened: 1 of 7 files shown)


MarcUsher (commented):
A couple of Qs/sense-checks from me, but otherwise this is really tidy 🙏

PR #967, opened 2025-11-03
Set collection dates

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net.mcas.ms/browse/FSPT-940

## 📝 Description
I am consciously/intentionally accepting a small amount of 'tech debt' here - happy to discuss.

This patch allows a platform admin to set reporting and submission dates for a
monitoring report, with some simple validation that these are chronologically
ordered.

A collection must have these dates set in order to go live/be scheduled to go
live.

I've taken a few shortcuts here and put the dates directly on the Collection
model. Theoretically we may want to have some kind of separate `Report` DB model
to hold these changes, and we'd probably want to also move the `name` attribute
from the collection to that model as well. However, I'm choosing not to do that
for a few overlapping reasons:

- We're short on time and this is much simpler; adding that interstitial model
  would require a chunky refactor to do it properly.

- We don't have anything other than reports which use collections right now;
  when we have concrete reasons to support applications, prospectus info, grant
  setup information etc - some other kind of collection basically - we will be
  able to reason through a sensible model more thoroughly.

image

image

### Review Readiness

### Testing

### DIFF
--- a/app/deliver_grant_funding/admin/views.py
+++ b/app/deliver_grant_funding/admin/views.py
@@ -4,7 +4,7 @@ from uuid import UUID
 from flask import current_app, flash, redirect, url_for
 from flask_admin import AdminIndexView, BaseView, expose
 
-from app.common.data.interfaces.collections import get_collection
+from app.common.data.interfaces.collections import get_collection, update_collection
 from app.common.data.interfaces.exceptions import NotEnoughGrantTeamUsersError
 from app.common.data.interfaces.grant_recipients import (
     create_grant_recipients,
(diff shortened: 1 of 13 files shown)


app/deliver_grant_funding/admin/views.py
```diff
@@ -36,65 +39,91 @@ class PlatformAdminReportingLifecycleView(PlatformAdminBaseView):
     def index(self) -> Any:
         form = PlatformAdminSelectGrantForReportingLifecycleForm(grants=get_all_grants())
         if form.validate_on_submit():
-            grant = get_grant(form.grant_id.data)
-            return redirect(url_for("reporting_lifecycle.tasklist", grant_id=grant.id))
+            grant = get_grant(form.grant_id.data, with_all_collections=True)
+            if len(grant.reports) == 1:
```
sfount: Nice 👌

PR #978, opened 2025-11-04
Fixing service desk url now we have a specific deliver grant funding option

srh-sloan:
Updates the service desk URL to take the user directly to the Deliver Grant Funding option

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -255,7 +255,9 @@ class _SharedConfig(_BaseConfig):
     INTERNAL_DOMAINS: tuple[str, ...] = ("@communities.gov.uk", "@test.communities.gov.uk")
 
     # Service Desk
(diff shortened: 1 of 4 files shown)


app/config.py
```diff
@@ -256,6 +256,9 @@ def AZURE_AD_AUTHORITY(self) -> str:
 
     # Service Desk
     SERVICE_DESK_URL: str = "https://mhclgdigital.atlassian.net/servicedesk/customer/portal/5"
+    # TODO update this when we create an access funding option in the service desk
```
samuelhwilliams: 1344 does look like an Access specific service desk URL already? 🤔

srh-sloan: D'oh, forgot to remove the comment!

samuelhwilliams (approved):

PR #979, opened 2025-11-04
Allow non-platform-admins to see all grants properly

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Allows the 'See all grants' link, and the 'List grants' view, to be visible and work if you have any way of accessing more than one grant in Deliver grant funding. Specifically:

- If you're an MHCLG org admin (form designer) or member (theoretical)
- If you're a grant team member in multiple grants.

We do this through a new `user.deliver_grants` (and matching `user.access_grants`) property, which lists all of the grants a user has (non-platform-admin) permissions for in the system.

### DIFF
--- a/app/deliver_grant_funding/templates/deliver_grant_funding/grant_base.html
+++ b/app/deliver_grant_funding/templates/deliver_grant_funding/grant_base.html
@@ -27,7 +27,7 @@
 
 {% set right_nav_items = [] %}
 
-{% if authorisation_helper.is_platform_admin(current_user) %}
+{% if authorisation_helper.is_deliver_org_member(current_user) or current_user.deliver_grants | length > 1 %}
   {%
     do right_nav_items.append({
         "text": "See all grants",
--- a/app/common/data/interfaces/grants.py
+++ b/app/common/data/interfaces/grants.py
@@ -1,7 +1,7 @@
 from typing import Sequence
 from uuid import UUID
 
-from sqlalchemy import and_, or_, select
+from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import joinedload, selectinload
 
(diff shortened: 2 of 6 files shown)


app/deliver_grant_funding/templates/deliver_grant_funding/grant_base.html
```diff
@@ -27,7 +27,7 @@
 
 {% set right_nav_items = [] %}
 
-{% if authorisation_helper.is_platform_admin(current_user) %}
+{% if authorisation_helper.is_deliver_org_member(current_user) or current_user.deliver_grants | length > 1 %}
```
sfount: Nice this feels like a clean abstraction 👌

app/common/data/models_user.py
```diff
@@ -30,6 +30,57 @@ class User(BaseModel):
...
+    )
+
+    # For access_grants, we need to join through GrantRecipient for organization-level access
+    # The logic is:
+    # - Direct: Grant.id == UserRole.grant_id, OR
+    # - Org-level: exists(GrantRecipient where GrantRecipient.grant_id == Grant.id
+    #              AND GrantRecipient.organisation_id == UserRole.organisation_id)
+    # We also filter by Organisation.can_manage_grants == False
+    # NOTE: This does not account for platform admin privileges; this is specifically for normal users
+    access_grants: Mapped[list["Grant"]] = relationship(
```
sfount: I wonder if we'll have to do some thinking about a consistent way of keeping properties like this useful in unit test factories (in general not this change because I think on the ORM model like this is a way of keeping database business logic neatly in one place) - there's one place on components we had to encapsulate what the join _should_ have been doing with an underlying database. 

I think in most cases this won't matter as the unit test itself can take care of it if it actually matters to that test but some of these concepts might find themselves solving the same issue in a few places (what we found on components)

PR #988, opened 2025-11-05
Error handling for 'OAuth2 Authorization code was already redeemed'

srh-sloan:
We are seeing the above error occasionally in prod and test environments:
https://funding-service-design-team-dl.sentry.io/issues/6799047399/events/76500e74e05a47449e599a928771d5a9/events/

## 📝 Description
The error occurs when we try and reclaim the same token more than once from AD, which isn't something we should do because those tokens are only ever valid once.

We think the scenario users are hitting to generate this error is as follows:
- They login to deliver grant funding and authenticate successfully with AD
- When they come back into our system, we don't have any roles or valid invitations for them, and they are not a platform admin
- We then return a `403` status code and display an error page
- If the user then refreshes that page (maybe trying to fix the error) we reload the same route, which means they are still authenticated with AD so we try and reclaim the same existing token. This then results in error 54005`.

We have taken 2 approaches to improve this:

1. If users generate this error, we redirect them to the sign out route. There is no way to recover from this as we need a new token, so signing them out will force them to sign in again and generate a new token.
2. We can reduce the likelihood of them getting this error, but not just returning a `403` error if the user has no roles. Instead we now redirect them to a new endpoint (`auth.signed_in_but_no_permissions`). This displays the same error and still returns a `403`, but if they refresh the page they are just refreshing our error page, we aren't trying to claim anything from AD.

### Before
Users could generate a 500 error page if they refresh after trying to login with no permissions. 

### After
They will now see the permissions error page, and will see it again if they refresh. 

## 🧪 Testing
- Point your local machine at real SSO (as per instructions in the README file)
- Remove yourself from the group `TEST-SG-APP-FUNDING-SERVICE-PLATFORM-DEVELOPERS`
- Remove any reference to your test AD account from your local database (`user`, `user_role`, `invitation`)
- Login locally (in a fresh session, eg. Incognito)
- You should see the 'You do not have permission to log into the Funding Service.' page, and the url should end in `/sso/permissions-error`.
- If you refresh this page, you should just see the permissions error again, not a generic 500 error page.

### Review Readiness

### Testing

### DIFF
--- a/app/common/auth/__init__.py
+++ b/app/common/auth/__init__.py
@@ -147,11 +147,10 @@ def sso_get_token() -> ResponseReturnValue:
 
     # Invitations route
     elif user is None:
-        user_invites = interfaces.user.get_usable_invitations_by_email(email=sso_user["preferred_username"])
-        if not user_invites:
-            all_invites = interfaces.user.get_invitations_by_email(email=sso_user["preferred_username"])
+        user_invites = interfaces.user.get_invitations_by_email(email=sso_user["preferred_username"])
+        if not any(invite.is_usable for invite in user_invites):
             return redirect(
-                url_for("auth.signed_in_but_no_permissions", invite_expired=len(all_invites) > 0),
+                url_for("auth.signed_in_but_no_permissions", invite_expired=len(user_invites) > 0),
             )
         user = interfaces.user.create_user_and_claim_invitations(
             azure_ad_subject_id=sso_user["sub"],
--- a/app/common/data/interfaces/user.py
+++ b/app/common/data/interfaces/user.py
@@ -305,14 +305,11 @@ def get_invitation(invitation_id: uuid.UUID) -> Invitation | None:
     return db.session.get(Invitation, invitation_id)
 
 
-def get_usable_invitations_by_email(email: str) -> Sequence[Invitation]:
-    return db.session.scalars(
-        select(Invitation).where(and_(Invitation.email == email, Invitation.is_usable.is_(True)))
-    ).all()
-
-
-def get_invitations_by_email(email: str) -> Sequence[Invitation]:
-    return db.session.scalars(select(Invitation).where(Invitation.email == email)).all()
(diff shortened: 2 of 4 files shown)


app/common/data/interfaces/user.py
```diff
@@ -234,6 +234,10 @@ def get_usable_invitations_by_email(email: str) -> Sequence[Invitation]:
     ).all()
 
 
+def get_invitations_by_email(email: str) -> Sequence[Invitation]:
+    return db.session.scalars(select(Invitation).where(Invitation.email == email)).all()
```
sfount: As a separate interface method this might want a very light touch integration test, or for less work it might be a named property on the method above 
```python
def get_invitations_by_email(email: str, *, usable: bool = False) -> Sequence[Invitation]:
```

I can see that there is a scenario in the page integration tests that checks the query param is set if there are remaining invitations which I think does indirectly cover this

app/common/auth/__init__.py
```diff
@@ -118,6 +127,8 @@ def sso_sign_in() -> ResponseReturnValue:
 def sso_get_token() -> ResponseReturnValue:
     result = build_msal_app().acquire_token_by_auth_code_flow(session.get("flow", {}), request.args)
     if "error" in result:
+        if MSAL_ERROR_AUTHORIZATION_CODE_WAS_ALREADY_REDEEMED in result.get("error_codes", []):
+            return redirect(url_for("auth.sign_out"))
```
sfount: It might be nice to log warnings when this is happening so that we'd be able to dig in to if we're still getting this failure mode and users might be getting confused - as this will go from a very vocal 500 to something only experienced by users indirectly

sfount (approved):
Looks like a good solution to avoid the 500 errors and nice coverage on the different route level scenarios

app/common/data/interfaces/user.py
srh-sloan: Refactored and test added

PR #991, opened 2025-11-05
BAU Align the grant tables on desktop

sfount:
This will probably get redesigned when more thought is put into it or things like pagination are needed but this keeps the columns in line on desktop for the time being.

We don't fix the widths on mobile because the default behaviour of picking as much space as possible for longer columns should win instead of keeping a fixed width for the two RHS columns.

### Before

funding communities gov localhost_8080_deliver_grants (5)

### After

funding communities gov localhost_8080_deliver_grants (4)


samuelhwilliams (approved):
Yeah really nice, ty

PR #992, opened 2025-11-05
Trial run github copilot auto-review

samuelhwilliams:
Mostly picked from the exemplar docs here: https://docs.github.com/en/copilot/tutorials/customization-library/custom-instructions/code-reviewer

Just curious to see if this ever feels useful, no real harm/cost in trying it


sfount (approved):
Feels sensible to me to try this out and we can always wind it back if its over zealous 👍

PR #999, opened 2025-11-07
Prevent changes to non draft reports

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-974

## 📝 Description
When a report has been scheduled to go live (ie moved out of the DRAFT status), we want to prevent anyone from making changes to it as that could undermine the sign-off that has happened and risks introducing new bugs or breaking the report.

This patch adds that layer of protection around all of the endpoints that allow mutating a report (or any of its children entities, such as groups, questions, expressions, etc).

Non-DRAFT reports will now show to everyone as read-only (the same as grant team members currently always see).

NOTE: There's an optional followup to consider where we might add some lower-level restrictions on editing non-DRAFT collections, eg through DB constraints/triggers or in our interfaces.

| Page | Before | After |
|---|---|---|
| List reports |image | image |
| List report sections | image | image |
| List section questions | image | image |

### Review Readiness

### Testing

### DIFF
--- a/app/common/auth/decorators.py
+++ b/app/common/auth/decorators.py
@@ -253,7 +253,10 @@ def collection_is_editable[**P]() -> Callable[[Callable[P, ResponseReturnValue]]
             user = interfaces.user.get_current_user()
             if not AuthorisationHelper.can_edit_collection(user=user, collection_id=collection.id):
                 # TODO: FSPT-549 - Reliably show flash messages everywhere, currently these just ... won't show up
-                flash("You cannot edit that collection as it is no longer in the drafting status.")
+                flash(
+                    f"You cannot edit the “{collection.name}” {collection.type} as it is {collection.status}",
+                    "error",
+                )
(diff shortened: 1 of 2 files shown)


app/deliver_grant_funding/routes/reports.py
```diff
@@ -1777,6 +1802,7 @@ def add_question_validation(grant_id: UUID, question_id: UUID) -> ResponseReturn
     methods=["GET", "POST"],
 )
 @has_deliver_grant_role(RoleEnum.ADMIN)
+@collection_is_editable()
```
sfount: Currently what you can see in "read only" is very very limited (i.e you can't click around in nested groups or see any info on conditions or validations)

I wonder if this decorator might be able to receive an optional param to only restrict on mutate (POST) requests so that it could be used for pages that should still load and present a limited version of themselves but want confident it won't run if somehow the user sneaks in a edit request

PR #1015, opened 2025-11-10
Re-raise integrity errors

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Instead of masking these with our own ValueError, just raise the original IntegrityError from sqlalchemy - this has better information on it for debugging and investigation.


srh-sloan (approved):
Nice 🚀

PR #1022, opened 2025-11-11
Set up certifiers

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-938

## 📝 Description
Adds pages to the platform admin reporting lifecycle tab to allow us to set up (and revoke) certifiers for Access grant funding.

This flow is designed to work with the organisations+certifiers exported spreadsheet from Delta, until such a time as they are exporting this data nightly to a private s3 bucket for us to ingest automatically. A copy of the latest export is in sharepoint; new copies can be requested through Delta's service desk.

TODO: future PR will add a grant-level task for overriding certifiers for this specific grant

### Review Readiness

### Testing

### DIFF
--- a/app/deliver_grant_funding/admin/forms.py
+++ b/app/deliver_grant_funding/admin/forms.py
@@ -4,12 +4,12 @@ from typing import TYPE_CHECKING, Any, Mapping, Sequence
 
 from flask import current_app
 from flask_wtf import FlaskForm
-from govuk_frontend_wtf.wtforms_widgets import GovDateInput, GovSubmitInput, GovTextArea
+from govuk_frontend_wtf.wtforms_widgets import GovDateInput, GovSubmitInput, GovTextArea, GovTextInput
 from markupsafe import Markup, escape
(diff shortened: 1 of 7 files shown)


app/common/data/interfaces/user.py
```diff
@@ -159,6 +193,42 @@ def remove_platform_admin_role_from_user(user: User) -> None:
...
+) -> UserRole | None:
+    user_role = get_user_role(user, organisation_id, grant_id)
+    if not user_role:
+        return None
+
+    existing_permissions = user_role.permissions
+    combined_permissions = list(set(existing_permissions) - set(permissions))
+
+    if not combined_permissions:
+        db.session.delete(user_role)
```
sfount: I don't know if we ever made a decision about this but reminder to myself there was a suggestion about removing the "member" role from both access and deliver permissions set ups which might change what the implication of having no permissions means (you still have access but no elevated perms). Unless we do that separately this feels good for keeping the user role table tidy

PR #1028, opened 2025-11-11
FSPT-958 Access grant funding list grants

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-958

## 📝 Description
Builds out the select grant page including adding in the fixture and folder structure for access grant funding.

funding communities gov localhost_8080_access_organisation_3ca29557-e6b9-4e6d-8290-5fb1c9535d1e_grants

### Review Readiness
- [x] PR title and description provide sufficient context for reviewers
- [x] Commits are logical, self-contained, and have clear descriptions

### Testing
- [x] I have tested this change and it meets the acceptance criteria for the ticket
- [x] I need the reviewer(s) to pull and run this change locally
- [x] New (non-developer) functionality has appropriate unit and integration tests

PR #1030, opened 2025-11-12
FSPT-960: App index redirects to Access index

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-960

## 📝 Description
Deliver grant funding and Access grant funding have two distinct and separate user groups, and most users will go directly to the relevant service. However, there's still a chance a user will try to access the root of the service, in which case we need to make a decision about where we redirect them or whether we show options for going through to Access or Deliver to the user.

In the short term, we decided to redirect to the Access grant funding index. The reasons for this are:
- Magic link login (which is what the Access grant funding index will in turn redirect to) is set up to deal with internal users who might have accidentally made their way there, and will send them on to the SSO login route
- We want to minimise friction for external grant recipient users
- These users are also less likely to be aware of the difference between Access and Deliver grant funding, so we should default to the service they need

This does have implications for internal (mainly Funding Service) users who might have bookmarked funding.communities.gov.uk and will then be redirected to Access rather than Deliver as expected, so we should discuss this.

Originally part of https://github.com/communitiesuk/funding-service/pull/1028 but pulled out into a separate PR so it can be considered in full.

### Review Readiness

### Testing

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -306,7 +306,7 @@ def create_app() -> Flask:  # noqa: C901
 
     @app.route("/", methods=["GET"])
     def index() -> ResponseReturnValue:
-        return redirect(url_for("deliver_grant_funding.list_grants"))
+        return redirect(url_for("access_grant_funding.index"))
 
     # when developing we want the toolbar assets to not cause the page to flicker
     # otherwise we don't want the server to continually 304 on assets the browser has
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
do we still want this? or shall we close it off?

PR #1034, opened 2025-11-13
Add 'send emails' task to reporting lifecycle

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-942

## 📝 Description
When a report has been opened and is now live on Access grant funding, we need to send emails to all grant recipient users to let them know they can now start filling in the report. Until we have some background worker-style task manager, this will need to be done off-service.

We make this as easy as possible by providing a CSV of all user data that can be uploaded to Notify to send the email batch.

### Review Readiness

### Testing


app/common/data/interfaces/user.py
```diff
@@ -41,9 +41,12 @@ def get_users_with_permission(
     organisation_id: uuid.UUID | None | TNotProvided = NOT_PROVIDED,
     grant_id: uuid.UUID | None | TNotProvided = NOT_PROVIDED,
 ) -> Sequence[User]:
-    if organisation_id is None and grant_id is not None and grant_id is not NOT_PROVIDED:
-        raise ValueError("If specifying grant_id, must also specify organisation_id")
+    """Finds all users in the system with the given permission and matching organisation and/or grant.
 
+    NOTE: If you pass a grant but don't pass an organisation, this will only return users with the permission
+          *specifically* for that grant. It will not return users who have that permission in the grant through an
+          org-level permission (ie UserRole.grant_id IS NULL)
```
sfount: As we're already set up for TNotProvided can we do 

```python
if grant_id is not NOT_PROVIDED and organisation_id is NOT_PROVIDED:
  organisation_id = grant.organistion_id
```

This feels a bit more intuitive to me, could remove this note and the nice thing about how you've set it up is you could still provide an explicit `organisation_id=None` to get only users with direct access, if you ever did need that

app/deliver_grant_funding/admin/views.py
```diff
@@ -466,3 +475,76 @@ def make_report_live(self, grant_id: UUID, collection_id: UUID) -> Any:
             grant=grant,
             collection=collection,
         )
+
+    @expose("/<uuid:grant_id>/<uuid:collection_id>/send-emails-to-data-providers", methods=["GET"])  # type: ignore[misc]
+    @auto_commit_after_request
```
sfount:
```suggestion
```

app/common/data/interfaces/user.py
samuelhwilliams: Not totally sure I agree it's more intuitive, but I think one we can keep an eye on.

app/deliver_grant_funding/admin/views.py
```diff
@@ -466,3 +475,76 @@ def make_report_live(self, grant_id: UUID, collection_id: UUID) -> Any:
...
+        csv_writer = csv.DictWriter(
+            csv_output,
+            fieldnames=["email_address", "grant_name", "reporting_period", "report_deadline", "grant_report_url"],
+        )
+        csv_writer.writeheader()
+
+        grant_recipients = get_grant_recipients(grant=grant, with_data_providers=True)
+        grant_recipient_data_providers = [
+            data_provider for grant_recipient in grant_recipients for data_provider in grant_recipient.data_providers
+        ]
```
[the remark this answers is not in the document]
samuelhwilliams: I don't believe this would return the same user multiple times 👀

PR #1057, opened 2025-11-17
Set up Cache-Control headers

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
We're not currently doing a good job of caching our assets, for two reasons:

1) We set a default cache duration of 31536000 (1 year) seconds for any 'sent files' (assets) 2) We haven't enabled caching on CloudFront.

This PR deals with the first part of that, setting up much longer cache durations for all of our static assets (1 year), and explicitly disabling caching on everything else. I'll follow up by enabling caching on CloudFront.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -193,6 +193,20 @@ def create_app() -> Flask:  # noqa: C901
     record_sqlalchemy_queries.init_app(app, db)
     govuk_markdown.init_app(app)
 
+    @app.after_request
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
Let's get this in as it makes caching work locally, but https://mhclgdigital.atlassian.net/browse/FSPT-856 exists to make it work in cloudfront

PR #1077, opened 2025-11-21
FPST-962 Link to reports page when emailing data providers

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-962

## 📝 Description
Updates the call to action to sign in and fill in a new report thats available to link to that grant and organisation specifically.

This chooses to link to the reports page so the data provider lands on a page that gives them a context of whats been done so far and what they now have to do (rather than the report page specifically which will take them straight into filling in the tasklist).

We could decide to also take them straight to the tasklist which should be an easy change.

### DIFF
--- a/app/deliver_grant_funding/admin/views.py
+++ b/app/deliver_grant_funding/admin/views.py
@@ -524,29 +524,33 @@ class PlatformAdminReportingLifecycleView(PlatformAdminBaseView):
         csv_writer.writeheader()
 
         grant_recipients = get_grant_recipients(grant=grant, with_data_providers=True)
-        for grant_recipient in grant_recipients:
-            for data_provider in sorted(grant_recipient.data_providers, key=lambda u: u.email):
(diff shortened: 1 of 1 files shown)


app/deliver_grant_funding/admin/views.py
```diff
@@ -524,27 +524,29 @@ def download_data_providers_csv(self, grant_id: UUID, collection_id: UUID) -> An
...
-            csv_writer.writerow(
-                {
-                    "email_address": data_provider.email,
-                    "grant_name": grant.name,
-                    "reporting_period": reporting_period,
-                    "report_deadline": report_deadline,
-                    "grant_report_url": grant_report_url,
-                }
-            )
+        for grant_recipient in grant_recipients:
```
samuelhwilliams: probably nice to sort grant recipients by org name? i've been doing that elsewhere; again one we could probably put on the relationship for consistency

PR #1106, opened 2025-11-25
FSPT-1018 Submission events v2

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1018

## 📝 Description
Updates the submission events table to prepare for "reducing" streams of submission events to allow workflows and state to go both backwards and forwards.

This will allow us to model submissions being signed off, declined, submitted and marked as overdue.

We'll also be able to streamline how individual forms/ sections are tracked with events so we're no longer deleting events when a user goes back and changes if its completed or not but can use the properties enabled by this change.

### Review Readiness
- [x] PR title and description provide sufficient context for reviewers
- [x] Commits are logical, self-contained, and have clear descriptions

### Testing
- [x] I have tested this change and it meets the acceptance criteria for the ticket
~- [ ] I need the reviewer(s) to pull and run this change locally~
- [x] New (non-developer) functionality has appropriate unit and integration tests
- [x] End-to-end tests have been updated (if applicable)
- [x] Edge cases and error conditions are tested

PR #1107, opened 2025-11-26
FSPT-1018 Submission events reducer lib

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1018

## 📝 Description
Depends on https://github.com/communitiesuk/funding-service/pull/1106

Note this adds the proposed core mechanism and in two following commits will be: 
- hooked into the existing "enum exists" event logic for existing flows
- extended to factor in all the events to be consumed by new workflows/ pages

Adds a small lib to represent submission events and their properties.

By processed events in a reduced stream like this, either grouped by a
form on the submission or the sumbission itself we should be able to
flexibility model the workflow and audit requirements of submissions:
- awaiting sign off (if certification required)
- being signed off (if certification required)
- being declined (if certification required)
- being submitted

Originally tried using pydantic but as the type validation didn't feel
useful and the overhead in syntax to make it do something it didn't seem
like it wanted to do did't feel worth it - uses native dataclasses
instead.

Shallow copy is used when extracting metadata interfaces from
submission event models as `asdict` creates a deep copy which means
you'll end up with two ORM references which doesn't feel right.

### Review Readiness

### Testing
~- [ ] I need the reviewer(s) to pull and run this change locally~

### DIFF
--- a/app/common/helpers/submission_events.py
+++ b/app/common/helpers/submission_events.py
@@ -12,6 +12,8 @@ if TYPE_CHECKING:
     from app.common.data.models_user import User
 
 
+# Mixins - schema protocol used to guarantee that the properties we're tracking using events
+# and our final state line up
 @dataclass
 class CompletedMixin(Protocol):
     is_completed: bool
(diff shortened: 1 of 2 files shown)


srh-sloan (changes requested):
Tiny little change requested but otherwise really great work on this. It's clear, well thought through and will work for all our existing scenarios as well as future ones 👍 ❇️ 🚀

PR #1110, opened 2025-11-26
BAU: VS Code settings for formatting HTML on save

MarcUsher:
## 🎫 Ticket
BAU

## 📝 Description
VS Code's default HTML format on save does not follow our desired HTML formatting rules, which are applied with prettier through Make commands and used in our pre-commit hooks.

This change adds all the configurations passed to the make format/check-html commands to the prettierrc config file and updates the VS Code settings.json to use these settings as the formatter for HTML files.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,18 @@
 {
-    "plugins": ["prettier-plugin-jinja-template"]
+    "tabWidth": 4,
+    "overrides": [
+        {
+            "files": "**/*.html",
+            "options": {
+                "plugins": [
+                    "prettier-plugin-jinja-template"
+                ],
+                "parser": "jinja-template",
+                "tabWidth": 2,
+                "htmlWhitespaceSensitivity": "css",
+                "bracketSameLine": true,
+                "printWidth": 240
+            }
+        }
+    ]
 }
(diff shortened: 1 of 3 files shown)


.prettierrc
```diff
@@ -1,3 +1,17 @@
...
+                    "prettier-plugin-jinja-template"
+                ],
+                "parser": "jinja-template",
+                "tabWidth": 2,
+                "htmlWhitespaceSensitivity": "css",
+                "bracketSameLine": true,
+                "printWidth": 240
+            }
+        }
+    ]
```
samuelhwilliams: Does this mean we can remove the config from Makefile?

```
.PHONY: check-html
check-html:
	npx prettier --plugin=prettier-plugin-jinja-template --parser=jinja-template --tab-width=2 --html-whitespace-sensitivity css --bracket-same-line=true --print-width=240 --check **/*.html

.PHONY: format-html
format-html:
	npx prettier --plugin=prettier-plugin-jinja-template --parser=jinja-template --tab-width=2 --html-whitespace-sensitivity css --bracket-same-line=true --print-width=240 --write **/*.html
```

MarcUsher: Yeah looks like this isn't needed anymore 🙌

PR #1112, opened 2025-11-26
Track when grants and collections change state

samuelhwilliams:
This might be useful for tracking how long the funding service and form designers need to prepare a grant, and a report, collectively to go live. Shorter times should be 'better'.


srh-sloan (commented):
Nice! Good to be able to track these 👍

PR #1129, opened 2025-12-03
Fspt 932 afghans new features

srh-sloan:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-932

## 📝 Description
Updates the communities for afghans seeded grant to use add another and reference data. Please pull and seed locally to sense check!

### Before
Screenshot 2025-12-03 at 10 37 11

### After

Screenshot 2025-12-03 at 10 30 53
Screenshot 2025-12-03 at 10 30 32

## 🧪 Testing
Have imported locally and run through the form

## 📚 Additional Notes
Did this in stages, hence several commits, but will squash for merging


srh-sloan:
> I think this question is redundant now?


Yeah I did wonder this, but cam edown on the side of leaving it in because then we can just reference it once we add the functionality to limit the number of add anothers, instead of adding in something almost identical at that point? But also don't mind removing it for now

samuelhwilliams (approved):
Really nice to have this feature used in a stock grant now, well done 🙇  (squash merge please 🫣 😂 )

PR #1131, opened 2025-12-04
BAU: Fix 'view submission' styling & breadcrumb

MarcUsher:
## 🎫 Ticket
BAU

## 📝 Description
Fixes a couple of things noticed during a demo:
- live submissions were showing the submitted user's email in the breadcrumb (as it is for test submissions) instead of the grant recipient org
- 'Submitted' status doesn't have a tag and was showing as raw text on the the view submission page, as previously we only saw this status in the list reports table where it would be styled with govuk table cell styling. The fix here doesn't look _great_ but we've talked about updating the layout of this page to match the 'view_locked_report' page, in which case it would become redundant as that doesn't display the status when submitted (or would display it elsewhere).

### Before
image

No change to styling in a table:
image

### After
image

No change to styling in a table:
image

### DIFF
--- a/app/common/templates/common/macros/status.html
+++ b/app/common/templates/common/macros/status.html
@@ -60,7 +60,7 @@
     }
   %}
   {% if status in show_without_tag %}
-    {{ status }}
+    <span class="govuk-body govuk-!-margin-bottom-0">{{ status }}</span>
   {% else %}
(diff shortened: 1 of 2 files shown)


sfount (approved):
Yeah in this context (outside of a table/ tasklist) this doesn't look quite right but not being unstyled is much better! As you said this page will need to be thought through now we've got the base workflows in

PR #1149, opened 2025-12-09
BAU: Send certify confirmation/decline emails to all users

MarcUsher:
## 🎫 Ticket
BAU

## 📝 Description
We previously sent the decline certification and certification/submit success email notifications only to those users who had sent the report for certification and had declined or certified it. In discussion with product it was decided that we should default to sending these emails to all grant recipient data provider and certifier users, rather than just the individuals who had triggered those submission events. This is something we will monitor and can roll back if it gets too spammy.

## 🧪 Testing
Decline or certify a report locally for a grant recipient where you have multiple data providers and certifiers. In the Notify API tab you should now see emails to all these users, not just those who sent the report for certification and declined or certified it.

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/access_grant_funding/routes/reports.py
+++ b/app/access_grant_funding/routes/reports.py
@@ -82,8 +82,7 @@ def view_locked_report(organisation_id: UUID, grant_id: UUID, submission_id: UUI
         submission.certify(user=user)
         submission.submit(user=user)
 
-        users_needing_confirmation = {submission.certified_by, submission.sent_for_certification_by}
-        for unique_user in users_needing_confirmation:
+        for unique_user in grant_recipient.unique_data_providers_and_certifiers:
             if unique_user is not None:
                 notification_service.send_access_submission_certified_and_submitted(
                     email_address=unique_user.email,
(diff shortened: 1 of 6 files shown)


app/access_grant_funding/routes/reports.py
```diff
@@ -76,8 +76,9 @@ def view_locked_report(organisation_id: UUID, grant_id: UUID, submission_id: UUI
         submission.certify(user=user)
         submission.submit(user=user)
 
-        users_needing_confirmation = {submission.certified_by, submission.sent_for_certification_by}
-        for unique_user in users_needing_confirmation:
+        unique_users_needing_confirmation = set(grant_recipient.data_providers + list(grant_recipient.certifiers))
```
srh-sloan: I needed this for the deadline reminder emails too - maybe we add an `all_unique_users` property to the grant_recipient?

tests/integration/access_grant_funding/routes/test_reports.py
```diff
@@ -285,6 +307,7 @@ def test_get_list_reports_not_grant_recipient(self, authenticated_grant_recipien
 
 
 class TestDeclineSignOff:
+    # FAILING
```
srh-sloan: Do you still need this comment?

MarcUsher: lol no thank you 😆

app/access_grant_funding/routes/reports.py
MarcUsher: @srh-sloan I've gone with `all_unique_data_providers_and_certifiers` as the property - we already have `users` on the grant_recipient model which we can just call `set()` on if we need them to be unique but as here we only want data providers and certifiers and not members (who might be view only even if that's not a thing yet) I thought it better to be explicit

srh-sloan (approved):

PR #1172, opened 2025-12-15
FSPT-921 Remove add another entry

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-921

## 📝 Description
While working on add another we included a link to remove entries that you no longer needed and built out the backend logic to support this. We never connected the frontend form runner pages and the ability to remove a given entry.

Add an add another context page to confirm if the entry should be removed and then connect the backend interface to remove data at a given entry.

In the future we probably want to allow you to remove add another entries before you've added all of the information for that entry (currently "Change" and "Remove" will only show when an entry is complete) but that thinking should probably be done alongside if we should show a check your entry answers for more involved add another question groups.

https://github.com/user-attachments/assets/dca75300-f9e2-4539-a4b9-502fe2e4639f

### Review Readiness

### Performance and security
- [-] No N+1 query problems introduced
- [-] Any new DB queries include appropriate where clauses based on the user's permissions

### Testing

### DIFF
--- a/app/deliver_grant_funding/routes/runner.py
+++ b/app/deliver_grant_funding/routes/runner.py
@@ -56,8 +56,8 @@ def submission_tasklist(grant_id: UUID, submission_id: UUID) -> ResponseReturnVa
     "/grant/<uuid:grant_id>/submissions/<uuid:submission_id>/<uuid:question_id>/<int:add_another_index>/<any('remove'):action>",
     methods=["GET", "POST"],
 )
-@has_deliver_grant_role(RoleEnum.MEMBER)
 @auto_commit_after_request
+@has_deliver_grant_role(RoleEnum.MEMBER)
 def ask_a_question(
     grant_id: UUID,
     submission_id: UUID,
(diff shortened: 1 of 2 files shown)


app/access_grant_funding/routes/runner.py
```diff
@@ -143,6 +148,7 @@ def ask_a_question(
         source=FormRunnerState(source) if source else None,
         add_another_index=add_another_index,
         grant_recipient_id=grant_recipient.id,
+        is_removing=action == "remove" if action else None,
```
samuelhwilliams:
```suggestion
        is_removing=action == "remove",
```

?

app/common/collections/runner.py
```diff
@@ -38,6 +43,7 @@ def __init__(
         form: Optional["Form"] = None,
         source: Optional["FormRunnerState"] = None,
         add_another_index: Optional[int] = None,
+        is_removing: Optional[bool] = None,
```
samuelhwilliams:
```suggestion
        is_removing: bool = False,
```

PR #1179, opened 2025-12-16
BAU: Consolidate confirmation pages

MarcUsher:
## 🎫 Ticket
BAU

## 📝 Description
We had two near-identical confirmation pages for when a submission is submitted to MHCLG, one for signed off & submitted (when a report requires certification) and just submitted (when a report doesn't require it). These had near-identical route handlers as well.

This brings these two pages and routes together and updates the template to handle the small content differences if a report requires certification or not (including the relevant feedback url).

Also fixes a URL path to have hyphens rather than underscores.

Submitted confirmation - no certification
image

Submitted confirmation - with certification
image

## 🧪 Testing
- Submit a report that requires certification and one that doesn't
- See that the confirmation screen content is correct (including feedback URLs) depending on whether it requires certification or not, but that the URL path and template is the same

### Review Readiness

### Testing


MarcUsher:
> Nice consolidation ✅

ty ty, don't love the slightly verbose state/auth checks but I think it's something we can come back to

PR #1215, opened 2025-12-29
Optional: add type hinting around submission events

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1047

## 📝 Description
Use `@overloads` and TypedDicts to add type safety around submission events, ensuring that only correct data can be stored for each event.

This is only for declined certification right now and the `declined_by` attribute.


sfount (approved):
really nice for dx 🙇  quite a few places in the form tech I think we could use overloads with literals to remove some interface sharp edges

PR #1216, opened 2025-12-29
BAU Add test org and grant recipient to seed data

sfount:
Let developers jump straight in with testing access grant funding by setting up a test org and grant recipient for the cheeseboards in parks grant and connecting it all to the default user a developer would log in with through the stub server.

Note this allows for platform admin permissions to be seeded but doesn't include platform admin permissions on the default user to allow the checkbox to still mean something.


samuelhwilliams (approved):
Lovely 🙇‍♂️

PR #1229, opened 2026-01-07
BAU Show form previewers

sfount:
funding communities gov localhost_8080_deliver_grant_caf0ce3f-5175-f69c-66a9-41e2c2245845_report_4e8f04f8-57c2-0399-464b-e62ff96b684a (7)

funding communities gov localhost_8080_deliver_grant_caf0ce3f-5175-f69c-66a9-41e2c2245845_report_4e8f04f8-57c2-0399-464b-e62ff96b684a (8)

PR #1233, opened 2026-01-08
BAU Extend get submissions interface to optionally eagerly load users

sfount:
To avoid lazily fetching each user that has made a submission when checking previews, add an option to eagerly load users.

We should think this through on any interface that has to load i.e submissions for many different organisations where we're likely fetching each organisation individually from the db - having a uniform way of setting up interfaces to either join load or not should help make this part of adding any new methods.


sfount:
Follows on from https://github.com/communitiesuk/funding-service/pull/1229

PR #1234, opened 2026-01-09
BAU Only show the testing Access grant funding banner if context isn't for a live org

sfount:
Provide the grant recipient context to determine if its a user who is testing access grant funding is loading a page thats for a test or live org.

This will then:
- for pages where there is no grant recipient context, being a deliver user will be enough to say you're testing
- for page where there is a grant recipient, if that grant recipients organisation is not test it will not do the testing behaviour

This applies to:
- showing the orange testing banner with a link back to Deliver
- listing all reports or only reports that are "Open"
- which submissions are checked on the list reports page, test or live submissions

We've run into a confusing behaviour in test where if you're setting up a scenario using an internal email (for one of us) it will still think you're testing even if the organisation is real - this leads to some unpredictable behaviour where it will show its test data even though its not and will sometimes fetch the test data rahter than the real data you've started showing a discrepancy in status.

By providing the grant recipient (if its known) we can be specific about if the page you're interacting with is on behalf of a test or live org as well as factoring in the users status.

In production its unlikely that a Deliver user will ever have access to a real organisations data but if that ever did happen we shouldn't (incrorrecly) show that they are testing!

### DIFF
--- a/app/common/auth/authorisation_helper.py
+++ b/app/common/auth/authorisation_helper.py
@@ -11,7 +11,7 @@ from app.common.data.models_user import User
 from app.common.data.types import OrganisationModeEnum, RoleEnum
 
 if TYPE_CHECKING:
-    from app.common.data.models import GrantRecipient
+    from app.common.data.models import Organisation
 
 
 class AuthorisationHelper:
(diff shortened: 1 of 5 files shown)


app/common/auth/authorisation_helper.py
srh-sloan: Can we amend this comment please to cover the updated logic pls?

srh-sloan (approved):
Nice! Pulled locally and all working 👍

PR #1442, opened 2026-03-20
Centralise submission data handling

samuelhwilliams:
## 🎫 Ticket
Off the back of https://mhclgdigital.atlassian.net/browse/FSPT-1217

## 📝 Description
Use the new SubmissionDataManager which is the central place for understanding
how we store answers in the `submission.data` blob, instead of having this
knowledge embedded in different places:

- DB interfaces (writing)
- SubmissionHelper (reading)
- Tests

This new interface also provides a little bit more confidence that the
submission data is only updated when we intend it to be updated. Previously, any
writes to `submission.data` would be tracked and committed at the end of the
request automatically, even without making a specific interface call, because
we've marked the `data` field as Mutable JSON which tracks any changes.

With the SubmissionDataManager, changes are only persisted to the submission
with an explicit call to `update_submission_data`. This should help make it
clearer when we're persisting data, and make it harder for any changes to the
data to be accidentally committed.

The SubmissionHelper has a lot of responsibilities at the moment, so carving out
the ownership of 'what is the shape of submission data' helps to keep it a
little bit more understandable.

No visible changes (hopefully!)

## 🧪 Testing
This is a comprehensive refactor to the internals of reading/writing submission data - I don't have a prescriptive test plan, but please satisfy yourself that the fundamental structure read and written into `submission.data` remains the same.

The first commit includes some assertions within our interfaces that attempts to programmatically ensure the old way and the new way result in the same underlying data.

The mainline code changes here are actually quite small, but giving responsibility for submission data to the SubmissionDataManager class means updating a lot of tests. These should be thoroughly checked, but a lot of it will be pattern recognition review. (But please watch out for anywhere I might have slipped up!)

### Review Readiness

### Performance and security
- [-] Any new DB queries include appropriate where clauses based on the user's permissions

### Testing


app/common/data/interfaces/collections.py
```diff
@@ -299,101 +299,13 @@ def update_collection(  # noqa: C901
...
-
-    data.pop(str(question.id), None)
-
-    # Make sure that our new SubmissionDataManager helper does exactly the same thing
-    submission.data_manager.remove(question, add_another_index=add_another_index)
-    assert submission.data_manager.data == submission.data
+def update_submission_data(submission: Submission) -> None:
+    # We make a deepcopy here so that if any changes are made to `submission.data_manager.data` after this is flushed,
+    # the changes are not reflected on the submission.
+    submission._data = deepcopy(submission.data_manager.data)
```
MarcUsher, line 305: Discussed whether it's immediately understandable when calling this function that it's using the SubmissionDataManager to update the Submission data because you just pass the Submission through to this interface - maybe worth a little docstring to explain why this is? Agree it feels a bit redundant to pass in submission and submission.data_manager when calling it, even if that would help understand how you're updating the submission.

MarcUsher (approved):
This is really gooooood man, I agree that it makes a lot of this a) more explicit in places rather than relying on the mutable json type and b) makes the tests a lot easier to read and more obvious what's going on. Very nice refactor. Have tested this locally and everything still seems to work - form runner, add another, validations/conditions (on questions/groups and within add another), view and export submissions on Deliver, view locked report on Access etc.  Pair reviewed on Slack as well for confidence in how it works.

Just one tiny nit commit which you can ignore/put in in a follow up PR if you want to.

PR #1529, opened 2026-04-09
Add another group validation

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1266

## 📝 Description
Enable final submission validation checks to re-validate add another groups that have group validation attached.

Add-another group validation already works at the point of data entry, but we also re-validate all data when trying to submit a report, and this would skip group validation on add-another groups. Now it doesn't.

This patch also updates the seeded calculations report to include an add-another group with validation.

image

image

image

### Review Readiness

### Testing

### DIFF
--- a/app/common/collections/validation.py
+++ b/app/common/collections/validation.py
@@ -46,7 +46,6 @@ class SubmissionValidator:
 
         for component in form.cached_all_components:
             if isinstance(component, Group):
-                # TODO[FSPT-1224]: run this for add-another groups with validations as well?
                 if (
                     component.validations
                     and component.add_another_container is None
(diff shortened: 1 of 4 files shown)


MarcUsher (approved):
This works nicely and catches it all, nice one! Have tested with all kinds of add anothers, no odd behaviour or inconsistencies.

I know it's on our list to revisit the entire submission revalidation, and I feel like this extension of question group validation to add another adds one more reason to do that bit of work - if you have multiple add another answers that initially pass validation and you then change the answer they're validating against to trigger the full submission validation error when submitting, the current error gives you no indication which add another answer sets are the ones that are failing, which makes it really unintuitive for the user. Feels like this'll be an important thing to revisit sooner rather than later.

PR #1532, opened 2026-04-09
Validation fixes

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Fixes two validation bugs:

- Unable to create custom validations on single questions in same-page groups - you could not reference the question itself.
- No requirement to reference at least one question in the group when creating group validation


MarcUsher (approved):
Niiiiice! We were pairing on these so I'm confident they work (cos I witnessed them), and good to have the tests in there 🙏

PR #1533, opened 2026-04-10
FSPT-1261 Feature flag pre-award features

sfount:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1261

## 📝 Description
Proposes a feature flag registry so that there is one place where our temporary feature flags live during development. This should let devs know where they can gate functionality and also make sure we've removed it from everywhere when we are finished.

Allows individual grants to specify if they should allow pre-award features or not which should allow us to test the features as we build them in different environments without interrupting exsiting users in Deliver.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -51,6 +51,7 @@ from app.common.filters import (
     to_ordinal,
 )
 from app.common.helpers.collections import SubmissionAuthorisationError
+from app.common.helpers.feature_flags import FeatureFlags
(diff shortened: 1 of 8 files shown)


samuelhwilliams (approved):
mild agree with 🤖  but good to go! nice thin wrapper to centralise this, we don't need anything more sophisticated for now

samuelhwilliams (approved):

PR #1534, opened 2026-04-10
Remove component self-references

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
We've had a couple of form design friction points lately:

- Unable to delete a question that has validation on itself only (eg answer >= 0) - forced to delete the validation first.
- Can't delete a question group where the questions inside that group also contain only references to things inside the group - forced to clean all of those internal references first.

### Review Readiness

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -1057,9 +1057,13 @@ def raise_if_question_has_any_dependencies(question: Question | Group) -> Never 
     child_components_ids = [
         c.id for c in [question] + (question.cached_all_components if isinstance(question, Group) else [])
     ]
+    # Only consider references whose dependent component is *outside* the subtree being deleted.
+    # Internal references (e.g. a group validation that references its own child questions) will be
+    # cascaded away alongside the subtree, so they should not block deletion.
     component_reference = (
         db.session.query(ComponentReference)
         .where(ComponentReference.depends_on_component_id.in_(child_components_ids))
+        .where(ComponentReference.component_id.notin_(child_components_ids))
         .all()
     )
     if component_reference:
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -1634,7 +1634,12 @@ def _validate_and_sync_expression_references(expression: Expression) -> None:  #
                 question_to_test=None,
             )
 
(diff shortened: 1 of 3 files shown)


app/common/data/interfaces/collections.py
```diff
@@ -1634,7 +1638,12 @@ def _validate_and_sync_expression_references(expression: Expression) -> None:  #
                 question_to_test=None,
             )
 
-        referenced_questions.add(expr_impl.referenced_question)  # ty:ignore[unresolved-attribute]
+        # For a managed validation attached to a question, expr_impl.referenced_question *is* the
+        # question the expression is attached to (the answer being validated). Don't record a
+        # self-reference — it would create a bogus ComponentReference row that falsely blocks
+        # deletion and same-page grouping.
+        if expr_impl.referenced_question != expression.question:  # ty:ignore[unresolved-attribute]
+            referenced_questions.add(expr_impl.referenced_question)  # ty:ignore[unresolved-attribute]
```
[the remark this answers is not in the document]
samuelhwilliams: done in https://github.com/communitiesuk/funding-service/pull/1543

app/common/data/interfaces/collections.py
```diff
@@ -1661,6 +1670,8 @@ def _validate_and_sync_expression_references(expression: Expression) -> None:  #
             referenced_questions.add(referenced_question)
 
     for referenced_question in referenced_questions:
+        if referenced_question == expression.question:
```
sfount, line 1673: Will these early exits be needed once we're no longer including the same question in reference questions and we've removed any from the database? I couldn't tell if there was a sneaky way they could fit in or if maybe these could be cleaned up once we've made this change

sfount (approved):
This is a really important cleanup for quality of life - I think its important it doesn't feel like a chore to use certain features because you know it will make it harder to change.

I've pulled locally and tried out as many scenarios as I can think of and the "Can delete self contained dependencies" feels intuitive and didn't spot anything unexpected

app/common/data/interfaces/collections.py
samuelhwilliams: Should be able to be cleaned up, but maybe we make them error instead as a check against system in an unexpected state?

PR #1537, opened 2026-04-10
Add filter for uppercasing first word of a phrase

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
This is so we can display phrases/fragments and still have them render grammatically correct if we're not inserting them into the middle of a sentence.


MarcUsher (approved):
Noice! Paired on this and all tested & checked

PR #1538, opened 2026-04-13
Fspt 1262 add configuration option to collections

israr-ulhaq:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1262
## 📝 Description
Adds `allow_public_sign_up` to collections so grant managers can configure whether a collection allows public sign up. This is gated behind the PRE_AWARD feature flag on the grant — the option only shows up in the Manage collection section when allow_pre_award is enabled on the grant.

### Before
Screenshot 2026-04-13 at 11 50 35

### After
[

https://github.com/user-attachments/assets/67fddae7-c804-4149-8460-14ce98801b3a

]

## 🧪 Testing
Enable `allow_pre_award` on a grant via the admin panel, then open a collection, you should see "Configure public sign up" in the Manage collection section under reports. Clicking it takes you to a yes/no page. Saving on a Draft collection works, saving on a live collection should shows an error.

### Review Readiness

### Testing


app/developers/data/grants.json
```diff
@@ -4676,7 +4683,7 @@
       "user_id": "78371bda-d2e5-4113-8cee-3daad6dd55ab"
     },
     {
-      "id": "003a0551-2703-403c-ac64-604183b3288b",
+      "id": "68e3b145-1ce2-4561-b0fc-0225c39f87ef",
```
sfount: 👀 I'm not sure but I suspect this shouldn't change as part of the export, it might be something with how we've set up the local permissions but if we could remove this change while keeping the new "allow_public_sign_up" property on this file

app/common/data/migrations/versions/054_allow_public_sign_up.py
```diff
@@ -0,0 +1,31 @@
...
+from alembic import op
+
+revision = "054_allow_public_sign_up"
+down_revision = "053_pre_award_flag"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
sfount: We can remove these comments on upgrade and downgrade

app/deliver_grant_funding/templates/deliver_grant_funding/reports/configure_public_sign_up.html
```diff
@@ -0,0 +1,47 @@
...
+  }}
+{% endblock %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <form method="post" novalidate>
+        {{ form.csrf_token }}
+
+        {% set sign_up_url = request.host_url ~ "access/grants/" ~ grant.code %}
```
sfount: Can we move this property on to the collection class? I think it would be good to start with having it in one place even if its a placeholder, that way we can add TODO comments explaining how we'll hook it up

i.e 
```python
class Collection(BaseModel):
    # ...
    
    @property
    def public_sign_up_url(self):
        # TODO: link to the grants/ collection public sign up page when registered as an endpoint in Access
        return url_for("auth.request_a_link_to_sign_in", _external=True)
```

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_report_sections.html
```diff
@@ -116,6 +116,11 @@ <h2 class="govuk-heading-m govuk-!-margin-top-5">Manage collection</h2>
         <li>
           <a class="govuk-link govuk-link--no-visited-state" href="{{ url_for('deliver_grant_funding.collection_configure_multiple_submissions', grant_id=grant.id, report_id=report.id) }}"> Configure multiple submissions </a>
         </li>
+        {% if feature_flags.PRE_AWARD.is_enabled %}
```
sfount, line 119: Nice, good to see this working!

sfount (commented):
PR looks great - I've left a few comments for quick tidy ups but this meets all of the criteria! Very nice coverage too

app/developers/data/grants.json
israr-ulhaq: I ran the command to export grants but nice I'll remove this bit.

app/deliver_grant_funding/templates/deliver_grant_funding/reports/configure_public_sign_up.html
israr-ulhaq: Make sense, thanks for flagging this.

app/deliver_grant_funding/templates/deliver_grant_funding/reports/configure_public_sign_up.html
```diff
@@ -0,0 +1,47 @@
...
+    })
+  }}
+{% endblock %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <form method="post" novalidate>
+        {{ form.csrf_token }}
+        <!-- display sign_up_url here once the public sign up endpoint is registered -->
```
sfount, line 21: 🤏 nit that we can address after merging, `<!-- -->` (HTML) comments will be rendered in the HTML that is sent to the end user but `{# #}` (Jinja) comments will be stripped out when rendering. As this is an internal comment we'll use the latter

sfount (approved):
Nice 🙂 pulled locally and verified db migrations and pages, looks great

PR #1545, opened 2026-04-14
Export grants to s3

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Allow exporting grants/reports from a deployed environment to s3, anonymously, so that we can import their config locally for review and testing.

Corresponding terraform PR: https://github.com/communitiesuk/funding-service-terraform/pull/93

I've deployed this to dev and run an export successfully.

### Review Readiness

### Testing

### DIFF
--- a/app/developers/commands.py
+++ b/app/developers/commands.py
@@ -3,6 +3,7 @@ import hashlib
 import json
 import uuid
 from collections import defaultdict
+from io import BytesIO
 from pathlib import Path
 from typing import Any, TextIO, TypedDict, cast
 from uuid import UUID
@@ -12,7 +13,7 @@ from flask import current_app
 from pydantic import BaseModel as PydanticBaseModel
 from sqlalchemy import delete, inspect, or_, select
 from sqlalchemy.exc import NoResultFound
-from werkzeug.datastructures import MultiDict
+from werkzeug.datastructures import FileStorage, MultiDict
 
 from app.common.collections.forms import build_question_form
 from app.common.data.base import BaseModel
@@ -62,7 +63,7 @@ from app.common.expressions import ExpressionContext
 from app.common.helpers.collections import SubmissionHelper
 from app.common.utils import slugify
 from app.developers import developers_blueprint
-from app.extensions import db
+from app.extensions import db, s3_service
 
 export_path = Path.cwd() / "app" / "developers" / "data" / "grants.json"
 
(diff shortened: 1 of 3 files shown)


app/developers/commands.py
```diff
@@ -248,31 +271,45 @@ def export_grants(grant_ids: list[uuid.UUID], output: str) -> None:  # noqa: C90
...
-            first_name = faker.first_name()
-            last_name = faker.last_name()
-            user_data["email"] = f"{first_name.lower()}.{last_name.lower()}@test.communities.gov.uk"
-            user_data["name"] = f"{first_name} {last_name}"
+        for user in users:
+            export_data = __replace_id(export_data, str(user.id), str(placeholder_user.id))
 
-        export_data["users"].append(user_data)
+        export_data["users"] = [to_dict(placeholder_user)]
+        export_data["user_roles"] = []
```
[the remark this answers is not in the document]
samuelhwilliams: not bothered about performance here

app/developers/commands.py
```diff
@@ -292,9 +328,23 @@ def export_grants(grant_ids: list[uuid.UUID], output: str) -> None:  # noqa: C90
...
+            )
+            s3_service.upload_file(buf, key=s3_key)
+            bucket = current_app.config["AWS_S3_BUCKET_NAME"]
+            click.echo(f"Written {len(grants)} grants to s3://{bucket}/{s3_key}")
+
 
 @developers_blueprint.cli.command("seed-grants", help="Load exported grants into the database")
 def seed_grants() -> None:  # noqa: C901
+    if current_app.config["IS_PRODUCTION"]:
+        raise click.ClickException("seed-grants must not be run in production; it deletes and recreates grants.")
```
sfount, line 346: Feels like a helpful check 👌

app/developers/commands.py
```diff
@@ -173,17 +174,38 @@ def _import_organisations_and_handle_org_ids(export_data: ExportData) -> ExportD
...
+        exclude_users = True
+    else:
+        exclude_users = bool(exclude_users)
+
+    if output == "s3":
+        if not s3_key:
+            raise click.ClickException("--s3-key is required when --output=s3")
+        required_prefix = current_app.config["GRANT_EXPORT_FILES_PREFIX"].rstrip("/") + "/"
+        if not s3_key.startswith(required_prefix):
+            raise click.ClickException(f"--s3-key must start with {required_prefix!r}")
```
sfount, line 201: Today I learned about interpolating with `!r`

app/developers/commands.py
```diff
@@ -248,31 +270,45 @@ def export_grants(grant_ids: list[uuid.UUID], output: str) -> None:  # noqa: C90
...
-        # Anonymise the user, but in a consistent way
-        if user.email not in USERS_NOT_TO_ANONYMISE:
-            faker = Faker()
-            faker.seed_instance(int(hashlib.md5(str(user_data["id"]).encode()).hexdigest(), 16))
-            first_name = faker.first_name()
-            last_name = faker.last_name()
-            user_data["email"] = f"{first_name.lower()}.{last_name.lower()}@test.communities.gov.uk"
-            user_data["name"] = f"{first_name} {last_name}"
+        for user in users:
+            export_data = __replace_id(export_data, str(user.id), str(placeholder_user.id))
```
sfount, line 281: Just to sense check this part anonymises everywhere a "created_by" or similar points to a user ID, which then the whole list of users gets replaced with the one anonymous user

sfount (approved):
All looks good to me, I did wonder if this script should just focus on exporting to stdout and then have the ad-hoc script calling it upload to s3 but given the number of changes it will have to make to behaviour anyway and the fact it already has stdout or file out options this interface all feels nice

app/developers/commands.py
samuelhwilliams: yeah that's right. It's more about reducing the number of users that get exported than it is about anonymisation - the standard script always makes an attempt to anonymise personal details on user models. But for exporting grant/report config from prod, we don't care about who did what and there's potential for a lot of different created bys

PR #1549, opened 2026-04-17
Make context aware editor handle parens inside references

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Our simplistic regexes for finding/extracting 'references' in strings were not working if those included parens, which was broken in two places:

- the frontend JS on our context-aware editors that maps human-readable labels to our internal reference format.

- the backend invalid reference detection

If a section had a name with parens in, eg `Report -> Risk(s) -> Risk name` then the frontend JS would not map this correctly to an internal reference like `((q_123))`. It would instead pass the raw label to the backend.

The backend _should_ have caught this and flagged it as an invalid/unknown reference, but we have the same regex on the Python side which was also failing to recognise this as a correct reference, so it just skipped over it entirely.

These two regexes need to stay in sync and detect the same references in the same way. We tie these together with a shared test data file that is tied to some new JS and Python tests, which should mean they are known to find the same set of references.

This also fixes one other nasty edge case we had where you couldn't wrap a reference in parens without adding spaces, which made some things like weird, eg we had an error message that would have to render like `You cannot spend more than your allocation ( £100,000 )`.

Now we can remove the spacing and have it like `You cannot spend more than your allocation (£100,000)`

| State | Before | After |
|---|---|---|
| valid parens reference |  |  |
| invalid parens reference |  |  |

### Review Readiness

### Testing
- [-] I need the reviewer(s) to pull and run this change locally

### DIFF
--- a/tests/fixtures/reference-regex-validation.json
+++ b/tests/fixtures/reference-regex-validation.json
@@ -1,0 +1,50 @@
+{
+    "comment": "This file is used to hold a set of test cases that run against both JavaScript (highlighting.test.js) and Python (unit/common/data/interfaces/test_collections.py).",
+    "test_cases": [
+        {
+            "pattern": "Text with no references",
+            "references": []
+        },
+        {
+            "pattern": "Text with a ((reference))",
+            "references": [
+                "((reference))"
+            ]
+        },
+        {
+            "pattern": "Text with a partial ((reference)",
+            "references": []
+        },
+        {
+            "pattern": "Text with a ((reference)) and ((another reference))",
+            "references": [
+                "((reference))",
+                "((another reference))"
+            ]
+        },
+        {
+            "pattern": "Text with a (( spaced reference ))",
+            "references": [
+                "(( spaced reference ))"
+            ]
+        },
+        {
+            "pattern": "Text with some (( reference(s) ))",
+            "references": [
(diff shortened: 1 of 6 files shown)


tests/fixtures/reference-regex-validation.json
```diff
@@ -0,0 +1,34 @@
+[
```
srh-sloan: Can we put comments in json if it's gonna be read by the tests? If so can we explain why this is here and flag the 2 places it's used, or else put something to that effect in both the JS and python tests so we know it's referenced by both?

srh-sloan (approved):
Looks good, well done on further investigating the regex. Little comment on just making sure we remember that json file is used by both the python and js code, but otherwise all good 👍

tests/fixtures/reference-regex-validation.json
samuelhwilliams: JSON doesn't take comments :( But I can tweak the structure to add an unused field with comments yeah, one mo 👍

samuelhwilliams: Done 🙇

srh-sloan (approved):
Nice! Really nice solution to making sure we have tested the same thing in both languages that use this ⭐

PR #1551, opened 2026-04-17
FIX Don't raise when adding the same add another expression context

sfount:
While checking component and component reference visibility, `_get_component_visibility_state_internal` sets up evaluation contexts and calls itself.

When there is a combination of an add another group and nested group with a condition (on a previous question in the section) the combination of checks + building contexts will end up creating an evaluation context and trying to add the add another context on top (for a second time).

This is sub-optimal as its checking visibility where it already has before (walking back up the nested group) but isn't technically a problem.

This patches the integrity check to make sure you're not creating an invalid evalutation context to allow the same add another context to be set however it does not address the very complicated and error prone approach to recursively checking visibility.

We have an additional proposed patch that moves condition checking to a directed graph which ensures checks and context are only built and evaluated once - this should be merged very shortly following.

### DIFF
--- a/app/common/expressions/__init__.py
+++ b/app/common/expressions/__init__.py
@@ -208,9 +208,6 @@ class ExpressionContext(ChainMap[str, Any]):
         Creates a new `ExpressionContext` with `add_another_context` set to the provided `add_another_context`, and the
         other contexts set to the same values as this context
         """
-        if self._add_another_context:
-            raise ValueError("add_another_context is already set on this ExpressionContext")
-
         if not component.add_another_container:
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):

samuelhwilliams:
We paired together on this patch. We consider this a short-term fix for an immediate issue affecting some users and preventing submission.

This should be followed up by work to refactor the visibility checking of collection questions into a single forward-pass/directed acyclic graph to avoid recursion and hard-to-follow convoluted logic.

PR #1555, opened 2026-04-17
BAU Firm up remove add another entry integrity checks

sfount:
Adds `check_entries` property to the form runner when removing an add
another entry. This is an optional check just to make sure that there
are the same number of entires as when we originally clicked through to
remove.

This should address some edge cases we've seen through errors where the
user presses back in the browser and attempts to re-remove entries. If
there multiple entries it should also avoid accidentally removing.

Any additional properties on the form runner library are completely
unmanagable, at this stage it needs to be split up into smaller pages so
that each of the interface methods can define what it cares about and
how to calulate valid states, where to go next, etc. - as it is there
are far too many things to check to make sure you're in the right
context.

### DIFF
--- a/app/access_grant_funding/routes/runner.py
+++ b/app/access_grant_funding/routes/runner.py
@@ -232,6 +232,7 @@ def ask_a_question(
     action: str | None = None,
 ) -> ResponseReturnValue:
     source = request.args.get("source")
+    check_entries = request.args.get("check_entries")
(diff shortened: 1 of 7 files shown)


srh-sloan (approved):
Tested locally and definitely feels like it makes more sense now if you click back after removing something. Nice little patch until we can refactor this to make it a bit more streamlined.

samuelhwilliams (approved):
Re-approved on @srh-sloan's behalf

PR #1556, opened 2026-04-20
BAU - adding caption to heading on grant recipient admin pages

srh-sloan:
## 📝 Description
Adding the grant name and report name as a caption on the heading of a couple of admin pages so that when adding grant recipients, you get some reassurance you are adding them to the right grant.

Also means that if you tick 'send report is ready to complete email', you know which report they are getting a link to.

### Before
Screenshot 2026-04-20 at 12 51 28

### After
Screenshot 2026-04-20 at 12 51 11


samuelhwilliams (approved):
yeahhh lovely spot, good stuff 🙇

PR #1568, opened 2026-04-22
Add e2e test coverage for group validation on number questions

srh-sloan:
## 🎫 Ticket
[FSPT-1202](https://mhclgdigital.atlassian.net/browse/FSPT-1202)

## 📝 Description
Adds config to the e2e tests to create a new group containing number questions, and add calculated group validation to check they add up to less than a previous answer.

### After
Screenshot 2026-04-22 at 10 47 56

## 🧪 Testing
Run the e2e tests

[FSPT-1202]: https://mhclgdigital.atlassian.net/browse/FSPT-1202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/e2e/deliver_grant_funding/reports_pages.py
+++ b/tests/e2e/deliver_grant_funding/reports_pages.py
@@ -891,6 +891,57 @@ class CreateCustomExpressionPage(ReportsBasePage):
         return edit_question_page
 
 
+class AddGroupValidationPage(ReportsBasePage):
+    add_validation_button: Locator
+
+    def __init__(
+        self, page: Page, domain: str, grant_name: str, report_name: str, section_name: str, group_name: str
+    ) -> None:
+        super().__init__(
+            page,
+            domain,
+            grant_name=grant_name,
+            heading=page.get_by_role("heading", name="Create a group validation"),
+        )
+        self.report_name = report_name
+        self.section_name = section_name
+        self.group_name = group_name
+        self.add_validation_button = self.page.get_by_role("button", name="Add group validation")
+
+        self.custom_expression_textarea = get_context_aware_textbox_locator_by_name(page, "Calculation")
+        self.add_data_to_expression_button = self.page.get_by_role("button", name="Reference data for calculation")
+
+        self.custom_message_textarea = get_context_aware_textbox_locator_by_name(page, name="Error message")
+        self.add_data_to_message_button = self.page.get_by_role("button", name="Reference data for error message")
+
+    def configure_custom_expression(
+        self, expression: str, expression_references: dict[str, DataReferenceConfig]
+    ) -> None:
+        wait_for_context_aware_textarea_to_be_ready(self.page, "custom_expression")
+        _fill_custom_field(
+            self, self.custom_expression_textarea, self.add_data_to_expression_button, expression, expression_references
+        )
+
+    def configure_custom_message(self, expression: str, message_references: dict[str, DataReferenceConfig]) -> None:
+        wait_for_context_aware_textarea_to_be_ready(self.page, "custom_message")
+        _fill_custom_field(
+            self, self.custom_message_textarea, self.add_data_to_message_button, expression, message_references
+        )
+
+    def click_create_group_validation_expression(self) -> "EditQuestionGroupPage":
+        self.add_validation_button.click()
+        edit_question_group_page = EditQuestionGroupPage(
+            self.page,
+            self.domain,
+            grant_name=self.grant_name,
+            report_name=self.report_name,
+            section_name=self.section_name,
+            group_name=self.group_name,
+        )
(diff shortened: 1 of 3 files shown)


tests/e2e/deliver_grant_funding/test_create_preview_collection.py
```diff
@@ -859,30 +937,52 @@ def complete_question_group(
...
+                    answer=nested_question["answers"][answer_idx].answer,
                 )
-        if group_to_test["display_options"] == GroupDisplayOptions.ONE_QUESTION_PER_PAGE:
             question_page.click_continue()
-    if group_to_test["display_options"] == GroupDisplayOptions.ALL_QUESTIONS_ON_SAME_PAGE:
-        question_page.click_continue()
+            if answer_idx < answer_count - 1 and group_to_test.get("validation"):
+                expect(
+                    question_page.page.get_by_text(group_to_test["validation"].evaluatable_expression.message)
+                ).to_be_visible()
```
samuelhwilliams: I think this doesn't feel totally aligned with the way we assert on error messages at the moment, which is by looking at a QuestionResponse's error message. Is there something we could do to tie this up more nicely? As this won't work if we ever want to test interpolated values in the group validation's error message 🤔

samuelhwilliams (approved):
Approved! But with a thought on structure/how we test the group validation error message 🙏

tests/e2e/deliver_grant_funding/test_create_preview_collection.py
```diff
@@ -851,38 +944,80 @@ def add_condition(
         add_condition_page.click_add_condition(edit_question_page)
 
 
+def answer_questions_and_check_for_expected_errors(
+    questions_on_this_page: list[QuestionDict],
+    question_page: RunnerQuestionPage,
+    group_validation_error: str | None = None,
+    same_page_group: bool = False,
```
samuelhwilliams, line 951: nit: this interface feels like it could be cleaner if you just passed the entire `QuestionGroup`? Here you're having to pull two values out of the dict (validation error message, same page group setting).

tests/e2e/deliver_grant_funding/reports_pages.py
```diff
@@ -1420,10 +1475,10 @@ def respond_to_question(self, question_type: QuestionDataType, question_text: st
                 expect(self.page.locator("[class='autocomplete__wrapper']")).to_be_attached()
                 element = self.page.get_by_role("combobox")
                 element.click()
-                element.fill(answer)
+                element.fill(answer)  # ty:ignore[invalid-argument-type]
```
samuelhwilliams: I think this type check error is telling you that the code is not doing what you think; ignoring this doesn't feel correct.

The conditional statement `question_type == QuestionDataType.YES_NO or question_type == QuestionDataType.RADIOS and isinstance(answer, str)` is trying to address this type error but the order of operations is getting in the way. Adding some parens around the `question_type` checks (or changing them to a single `in` check) should get rid of this type error.

Our codebase maybe leans us a bit towards just throwing ignores on places but this is maybe a bad habit for us to pick up.

I think we should also address the type ignore on 1497 and properly close the loop on what type we expect to reach certain places and what would be invalid.

samuelhwilliams (commented):
Nice extension to the e2e testing, feels great to have this in and help assure the functionality 🙏 

A few small nits, optional other than the type hint one.

tests/e2e/deliver_grant_funding/test_create_preview_collection.py
srh-sloan: I did it this way because then I can reuse this `answer_questions_and_check_for_expected_errors` in `complete_task` where it might not be a group, just a question on a page (a list of 1), so we can just have one place that responds to a question, see if we expect errors, check for them, and then go back round the loop again.

tests/e2e/deliver_grant_funding/reports_pages.py
```diff
@@ -1457,44 +1457,44 @@ def __init__(
...
+                date_to_enter = (
+                    datetime.date(*[int(a) for a in answer])
+                    if not approx_date
+                    else datetime.date(int(answer[0]), int(answer[1]), 1)
+                )
+                ReportsBasePage.fill_in_date_fields(
+                    self.page.get_by_role("group", name=question_text),
+                    date_to_enter,
+                    approx_date=approx_date,
+                )
```
samuelhwilliams, line 1475: nit: ideally we should close the gaps here and throw an error if we don't handle this. Presumably the e2e test would fail and we'd work out what went wrong, but raising an error here would give better debugging info and feel more 'correct'.

samuelhwilliams (approved):

samuelhwilliams (approved):

PR #1570, opened 2026-04-22
BAU Line up CPU and memory between infra and deploy

sfount:
The deploy action for AppRunner updates the service with the new image every time we release - IIRC as we depend on an external action there was no way to not include the CPU/ memory configuration when updating the service. If we don't include values it will override whatever was provisioned with defaults.

As we've increased this in the infra lets bump what the deploy script sets - we'll look to avoid this discrepancy when working with ECS.


srh-sloan (approved):
Good spot!

PR #1571, opened 2026-04-23
Bugfix: Fix submission status with not needed form

MarcUsher:
## 🎫 Ticket
Bugfix

## 📝 Description
A user raised a support ticket that they successfully submitted their report but the report was still showing as 'in progress'.

On investigation we found that they had successfully submitted their report but that the submission status calculation was incorrect - one of the sections in the form was marked as NOT_NEEDED but the status calculation wasn't taking this into account and was expecting all sections to be completed. This led to a successful submission but what the user saw was 'In Progress' and they were seemingly stuck in a loop.

This fix makes sure the submission status calculation takes into account NOT_NEEDED sections correctly.

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -338,14 +338,19 @@ class SubmissionHelper:
         submission_state = self.events.submission_state
 
         form_statuses = {self.get_status_for_form(form) for form in self.collection.forms}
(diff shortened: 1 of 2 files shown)


samuelhwilliams (approved):
We've all tested this, works as expected and fixes the bug 🙇

PR #1576, opened 2026-04-27
Order grant recipients by organisation name consistently

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Consistently sort grant recipients by organisation name so that the list of grant recipients in Deliver will be ordered the same as the list in the admin panels.


MarcUsher (approved):
Noice fix 🙏

PR #1578, opened 2026-04-27
Fix n+1 queries in platform admin pages

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
In lots of pages where we get grant recipients, we then fetch the organisation for each grant recipient as well. If we don't eagerly fetch these we end up emitting an extra query for each grant recipient to fetch its organisation. This loads them up front to improve performance/satisfy Sentry.

Fixes: https://communitiesuk.sentry.io/issues/112150917


srh-sloan (approved):
Nice, thanks for fixing. Added some grant recipients locally and can see the query number reducing from main to this branch 👍

sfount (approved):
Agree with the robot on `with_data_providers` unless its needed (this feels like something that could end up getting really big in a join), we know we'll be alerted about it if it becomes a problem!

PR #1579, opened 2026-04-27
Export grant config via email

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
Having to log into s3 and find/download the file has proven annoying - let's allow it to be sent directly to a developer by email.

Used like so: `awsedev -- uv run ./scripts/run-ad-hoc-task.py --command "flask developers export-grants <uuid> --output email --email <email> --exclude-users"`


MarcUsher (approved):
Nice tests & guarding against external emails receiving the files 💪

PR #1584, opened 2026-04-28
FIX bug where changing the conditions operator with a calculated condition failed

srh-sloan:
## 📝 Description
When multiple conditions exist for a single question or group, form designers have the option to choose if they must ALL match, or if ANY of them could match.

Clicking 'Change' for this operator was resulting in an error as the page to change between ANY and OR wasn't updated to work for non-managed expressions.

This PR updates that page to cope with both types of expression in the same way as the main list of conditions on the group/question page.

When these conditions exist:
Screenshot 2026-04-28 at 10 06 33
This page was failing to load but now works:
Screenshot 2026-04-28 at 10 04 03

## 🧪 Testing
- I've checked the page loads with different combinations of managed/unmanaged conditions, and checked the operator actually works when selected


samuelhwilliams (approved):
Optionally a test since this was throwing a 500 in mainline code :)

PR #1586, opened 2026-04-28
FSPT-1105 Remove feature flag from calculations

srh-sloan:
## 🎫 Ticket
This would remove the Platform Member requirement for the following tickets, allowing form designers to start using these features:
- [FSPT-1124](https://mhclgdigital.atlassian.net/browse/FSPT-1124) - Calculated validations
- [FSPT-1200](https://mhclgdigital.atlassian.net/browse/FSPT-1200) - Calculated conditions
- [FSPT-1202](https://mhclgdigital.atlassian.net/browse/FSPT-1202) - Group validation

## 📝 Description
Removes feature flags from functionality supporting the above tickets. They were written over the time where we introduced a feature flagging implementation, so were flagged in different ways:
- Calculated validations and conditions were flagged with a manual check on each route that the user had the required permissions. These have been removed
- Validation of a group of questions was flagged in the frontend using the `FeatureFlags` functionaltiy. The `GROUP_VALIDATION` flag and checks that use it have been removed.

Form designers will now see the group validation section:
Screenshot 2026-04-28 at 12 06 00

They will now be asked if their condition needs a calculation:
Screenshot 2026-04-28 at 12 06 13

They will have the option to create a calculated validation
Screenshot 2026-04-28 at 12 10 08

## 🧪 Testing
- Login as a platform admin user who is not a platform member
- Check you can create and edit calculated validations and conditions, and add validation to a group

[FSPT-1124]: https://mhclgdigital.atlassian.net/browse/FSPT-1124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FSPT-1200]: https://mhclgdigital.atlassian.net/browse/FSPT-1200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FSPT-1202]: https://mhclgdigital.atlassian.net/browse/FSPT-1202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_group_questions.html
+++ b/app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_group_questions.html
@@ -279,10 +279,7 @@
           {% endif %}
         {% else %}
           {# TODO: content review #}
-          <p class="govuk-body">
-            Validation on groups is only available when the question group is set to display all questions on the same page. Change the display setting to 'show all questions on the same page' if you need to add validation across multiple
-            questions in this group.
-          </p>
+          <p class="govuk-body">You can only add validations to a question group when questions are displayed on the same page. To add a validation, change the display setting to 'All questions on the same page'.</p>
         {% endif %}
       </div>
 
--- a/app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_group_questions.html
+++ b/app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_group_questions.html
@@ -254,39 +254,37 @@
         }}
       </div>
 
-      {% if feature_flags.GROUP_VALIDATION %}
-        <div class="govuk-body govuk-form-group">
-          <h2 class="govuk-heading-m govuk-!-margin-top-5">Validations</h2>
-          {% if group.same_page %}
-            <p class="govuk-body">Add validations to show an error message if a calculation including at least one answer from the question group does not meet a requirement.</p>
-            <p class="govuk-body">For example, validation to stop users from entering numbers that add up to more than a required amount.</p>
-            {% set rows = [] %}
-            {% with component=group %}
-              {% include 'deliver_grant_funding/partials/_explicit-validations.html' %}
-            {% endwith %}
+      <div class="govuk-body govuk-form-group">
+        <h2 class="govuk-heading-m govuk-!-margin-top-5">Validations</h2>
+        {% if group.same_page %}
+          <p class="govuk-body">Add validations to show an error message if a calculation including at least one answer from the question group does not meet a requirement.</p>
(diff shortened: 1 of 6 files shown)


samuelhwilliams:
Just on the two screenshots for calculated conditions/validation - did you check in with design on whether the landing page for validation should be the same structure: "Do you need a calculation for the validation? yes/no"?

It feels a bit like a lingering design discrepancy that we wouldn't ever revisit if we let it go now.

chaneldiep:
> Just on the two screenshots for calculated conditions/validation - did you check in with design on whether the landing page for validation should be the same structure: "Do you need a calculation for the validation? yes/no"?
> 
> It feels a bit like a lingering design discrepancy that we wouldn't ever revisit if we let it go now.

@samuelhwilliams thinking is that inconsistency is preferable for helping users get to the thing quicker, since we think users are likelier to need the more straightforward validations of greater than, less than or between two numbers than a custom calculation. 

We're sort of forcing users to choose a binary yes/no in the conditions journey to make it make sense (for this design anyway) but wanted to avoid adding the additional screen for the validations journey if we didn't need it.

Happy to monitor and see if the inconsistency feels a bit jarring to users though

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_group_questions.html
```diff
@@ -254,39 +254,37 @@ <h2 class="govuk-heading-m govuk-!-margin-top-5">Conditions</h2>
...
+                "text": "Add more validation" if group.validations else "Add validation",
+                "href": url_for('deliver_grant_funding.add_group_validation', grant_id=grant.id, group_id=group.id),
+                "classes": "govuk-button--secondary govuk-!-margin-bottom-3",
+              })
+            }}
           {% endif %}
-        </div>
-      {% endif %}
+        {% else %}
+          {# TODO: content review #}
```
samuelhwilliams, line 281: has this been done now?

srh-sloan: No I think we missed this, will ask Chanel next week when she's back

srh-sloan: No I think we missed this, will ask Chanel next week when she's back

samuelhwilliams (approved):

MarcUsher (approved):

PR #1590, opened 2026-04-28
Fix n+1 on `list_reports` page

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
The `list_reports` page iterates over all components of a report and its forms in order to calculate the status of the report. This means computing visibility for all components and checking references using `component.owned_component_references`. As we weren't preloading this, it was sending an extra SQL query for each component iterated. This will be tens/hundreds of components in all of our real reports.

Let's nix this by preloading all component references.

### Before
image

### After
image

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -359,6 +359,10 @@ def get_all_submissions_with_mode_for_collection(
             .joinedload(Collection.forms)
             .selectinload(Form.components)
             .joinedload(Component.expressions),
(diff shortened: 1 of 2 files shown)


MarcUsher (approved):
Noice, thanks for noodling through this fix! Makes sense to me and the test is really helpful at illustrating the behaviour this is guarding against now. Chunky ol query but it's doing a lot of things.

PR #1591, opened 2026-04-29
Fix/calculations error messages

srh-sloan:
## 📝 Description
Fixing error message bugs found during PO sign off for the calculations work (overarching parent: [FSPT-1105](https://mhclgdigital.atlassian.net/browse/FSPT-1105))

| Notes | Before | After | Testing |
|--------|--------|--------|--------|
| Using generic error message instead of ast node name for unavailable symbols | Screenshot 2026-04-29 at 06 39 17 | Screenshot 2026-04-29 at 06 40 29 | Use an unavailable symbol, eg. `**`, in a calculation |
| Error message displaying safe_qid not question name when trying to reference a question in the same single page group | Screenshot 2026-04-29 at 06 38 11 | Screenshot 2026-04-29 at 06 37 11 | Paste in the reference to another question in the same single page group |
| If you pasted in a reference to a later question, it told you it didn't exist rather than it was in the wrong order | Screenshot 2026-04-29 at 06 42 23 | Screenshot 2026-04-29 at 07 00 05 | Paste in a reference to a later question |

[FSPT-1105]: https://mhclgdigital.atlassian.net/browse/FSPT-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -1664,11 +1664,11 @@ def _validate_reference(  # noqa:C901
             attached_to_component, referenced_question
         ):
             raise InvalidReferenceInExpression(
-                f"Reference is not valid: {wrapped_reference}",
+                f"Reference is not valid: {referenced_question.name}",
                 field_name=field_name_for_error_message,
                 bad_reference=wrapped_reference,
-                form_error_message=f"You cannot use {wrapped_reference} because it is in the same group as "
-                f"{referenced_question.name}",
+                form_error_message=f"You cannot use {referenced_question.name} because it is in the same group as "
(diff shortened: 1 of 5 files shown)


samuelhwilliams (approved):
I think a test on the change to context keys+labels end point would be slightly nice - I had done that intentionally so if we're changing the behaviour here it might be nice to lock down 🙇  (caveat I obviously didn't write a test to lock it down in the first place 🫣)

samuelhwilliams (approved):

PR #1592, opened 2026-04-29
Allow seeding grants from any file path

samuelhwilliams:
Previously this was hardcoded to read from `app/developers/data/grants.json` - now you can point it at anything on your file system. Makes it easier to read eg a downloaded grant config in your downloads folder.


MarcUsher (approved):
Very nice! This all works for me locally (well, there's still that db check constraint error but that's another matter but the file path works, and nice to have the default so we don't have to update the docker compose etc)

PR #1593, opened 2026-04-29
Fix/calculated condition incompatible data types

srh-sloan:
## 🎫 Ticket
Bug found during PO sign off of calculations work ([FSPT-1105](https://mhclgdigital.atlassian.net/browse/FSPT-1105))

## 📝 Description
If a form designer were to copy and paste a reference to a non-number question into the calculation field of a custom condition, or of a group validation expression, this should fail validation. However we discovered this was passing validation.

This change adds a check to `validate_reference` to make this correctly fail validation.

### Before
The condition would save with a reference to a previous text question (called name)
Screenshot 2026-04-29 at 13 52 10

### After
Validation failure message when trying to use a text question reference (called name)
Screenshot 2026-04-29 at 13 51 33

[FSPT-1105]: https://mhclgdigital.atlassian.net/browse/FSPT-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -1556,6 +1556,7 @@ def _validate_reference(  # noqa:C901
 ) -> ExpressionReference:
     wrapped_reference = reference.wrapped
     unwrapped_reference = reference.unwrapped
+    is_calculation: bool = field_name_for_error_message == "custom_expression"
 
     if not unwrapped_reference:
         raise InvalidReferenceInExpression(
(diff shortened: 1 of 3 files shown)


app/common/data/interfaces/collections.py
```diff
@@ -1585,6 +1586,16 @@ def _validate_reference(  # noqa:C901
         )
 
     if referenced_question := reference.question:
+        # For calculations, the reference must of type number
+        if is_calculation and referenced_question.data_type != QuestionDataType.NUMBER:
```
samuelhwilliams, line 1590: comment/potential future refactor: i wonder if we should define in one place the set of data types that are available in custom expressions; this is hard-coded in a few different places and it feels like this might benefit from declaring the data types for all of our different scenarios in one place? eg interpolation vs custom expressions vs dataset referencing?

samuelhwilliams (approved):
Nice fix and test coverage :)

app/common/data/interfaces/collections.py
srh-sloan: Yeah, agreed. Added as a reference in this ticket as makes sense to do in one go I think: https://mhclgdigital.atlassian.net/browse/FSPT-1302

PR #1594, opened 2026-04-29
Address n+1 queries on data sources on list_submissions

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
This view needs to evaluate the status of all submissions, which is currently done by walking every form and component for every submission. Walking the components for a submission needs to pull the data sources to evaluate certain conditions.

Because we don't eagerly load data sources for each submission, we end up emitting an extra query for each submission as we load the page. 

By eagerly loading the data source we can make sure it's available to SQLAlchemy when iterating submissions, avoiding the n+1 query loop.

This will be required for performance until we start caching each submission's status and can avoid calculating it dynamically on demand each time.

### Before
image

### After
image

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -364,6 +364,7 @@ def get_all_submissions_with_mode_for_collection(
             .selectinload(Form._all_components)
             .selectinload(Component.owned_component_references),
             selectinload(Submission.events),
+            selectinload(Submission.data_sources),
             joinedload(Submission.created_by),
         )
     elif with_users:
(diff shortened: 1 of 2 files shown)


MarcUsher (approved):
Nice fix thanks Sam! 

In my head I feel like a mini review of some of our interfaces at some point might be helpful, some are really intentional in what they do and some rely a lot on the ORM and some have a lot of additional bits cos they're used in places that have very different requirements, could be helpful to do a bit of a sweep at some point (prob more for me than anything, just spotting where we've done things differently and why) 
- but also that doesn't feel the highest priority.

PR #1596, opened 2026-04-30
Lock report PDF generation

samuelhwilliams:
## 🎫 Ticket
BAU

## 📝 Description
We've seen intermittent errors on the export_report_pdf endpoint where Playwright errors because it's "already in" an async event loop.

This is happening when multiple requests are fired to the server at once to export a report, eg double clicking the "Export PDF" link. If these two requests reach the backend at the same time, gevent can spin up two workers for the requests in the same thread. These two requests can then both try to go into Playwright's sync API, which under the hood just wraps an async loop temporarily.

When the first request launches the browser to build the PDF it starts an async loop; when the second request arrives it tries to enter the sync loop and spawn another async loop under the hood, which then throws the error.

This patch adds a lock around PDF generation so that only one worker in a thread can be handling playwright at a time. The second worker will block until the first is complete.

We've only seen this in deployed environments because of the difference in how we run the apps (locally we use flask's debugging server which is single thread/single worker).

Fixes: https://communitiesuk.sentry.io/issues/116414809

## 🧪 Testing
Deployed to dev and confirmed fix by spamming 'Download PDF' - no longer erroring and all PDFs download correctly, albeit some delayed due to the locking.

### Review Readiness

### Testing

### DIFF
--- a/app/access_grant_funding/routes/reports.py
+++ b/app/access_grant_funding/routes/reports.py
@@ -1,5 +1,6 @@
 import io
 import os
+import threading
 from uuid import UUID
 
 from flask import abort, current_app, flash, redirect, render_template, send_file, url_for
@@ -20,6 +21,13 @@ from app.common.helpers.collections import CollectionHelper, SubmissionHelper
 from app.extensions import auto_commit_after_request
 from app.metrics import MetricEventName, emit_metric_count
 from app.types import FlashMessageType
+
+# Playwright's sync API runs an asyncio loop while it talks to chromium. Under gunicorn+gevent,
+# multiple greenlets share an OS thread and asyncio's running-loop is thread-local, so concurrent
+# PDF exports on one worker would observe each other's loop and Playwright would raise. Serialise
+# PDF generation per worker. threading.Lock is monkey-patched by gevent into a greenlet-aware lock,
+# so blocked greenlets cooperatively yield rather than block the OS thread.
+_pdf_export_lock = threading.Lock()
 
 
 @access_grant_funding_blueprint.route(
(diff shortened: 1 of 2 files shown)


app/access_grant_funding/templates/access_grant_funding/view_locked_report.html
```diff
@@ -45,7 +45,8 @@
       <a
         class="govuk-button govuk-button--secondary"
         download
-        href="{{ url_for('access_grant_funding.export_report_pdf', organisation_id=grant_recipient.organisation.id, grant_id=grant_recipient.grant.id, submission_id=submission.submission.id) }}">
+        href="{{ url_for('access_grant_funding.export_report_pdf', organisation_id=grant_recipient.organisation.id, grant_id=grant_recipient.grant.id, submission_id=submission.submission.id) }}"
+        data-prevent-double-click="true">
```
sfount: Looking at this a bit more closely I don't think the `data-prevent-double-click="true"` is actually registering which is likely because its a link styled as a button rather than an actual button.

There's probably a middle ground for making this still work but I don't have a suggestion off the top of my head - I'm confident in the backend locking mechanism in the meantime but we can either think through a solution for this or probably best to remove (if corroborated!)

sfount (approved):

app/access_grant_funding/templates/access_grant_funding/view_locked_report.html
samuelhwilliams: yeah removed 🤦

sfount (approved):

PR #1600, opened 2026-04-30
Govuk notify callback

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1297

## 📝 Description
Start handling callbacks from GOV.UK Notify so that we can get notified of delivery failures for emails we send to prompt offline/manual handling (for now).

The callback will only be enabled in the production environment. Due to our GOV.UK Notify set up (and limitations of the Notify service), callbacks for all environments will go to the production service. Due to this, and our dev/test env testing set up, we will need to ignore any failures from non-prod environments. This is done naively for now through an ignore on our internal domains list (and old @levellingup emails).

The new environment variable/secret has already been configured and merged for all environments.

## 🧪 Testing
There's an element of "test this in prod" going on here. To provide confidence pre-merge, I hooked up a separate test Notify service to send callbacks to my local machine and verified that the payload is correct, token authorises correctly, and 2xx responses can be given. But I've not tested edge cases around failed emails because my trial service can only send to my own (valid) email address. I'll check this in prod / we'll quickly see failed emails starting to come through.

### Review Readiness

### Testing


app/deliver_grant_funding/routes/api/guidance.py
```diff
@@ -1,24 +1,23 @@
 import itertools
 from uuid import UUID
 
-from flask import Blueprint, current_app, jsonify
+from flask import current_app, jsonify
 from flask.typing import ResponseReturnValue
 
-from app.common.auth.authorisation_helper import AuthorisationHelper
+from app import AuthorisationHelper
```
MarcUsher, line 7: ~~Has this done that weird thing you were talking about where it's importing things from `app` rather than the proper path?~~

MarcUsher: Lol ignore you fixed it in this next commit 😆

app/deliver_grant_funding/routes/api/callbacks.py
```diff
@@ -0,0 +1,85 @@
...
+
+@deliver_grant_funding_api_blueprint.post("/govuk-notify-callback")
+def govuk_notify_callback() -> ResponseReturnValue:
+    if not current_app.config["GOVUK_NOTIFY_CALLBACK_TOKEN"]:
+        return jsonify(), 500
+
+    if (
+        not request.authorization
+        or not request.authorization.token
+        or not secrets.compare_digest(
```
MarcUsher, line 47: Ooooh I didn't know this `compare_digest` thing, that's cool

app/deliver_grant_funding/routes/api/callbacks.py
```diff
@@ -0,0 +1,85 @@
...
+        return jsonify(), 400
+
+    try:
+        callback_data = GovukNotifyCallbackModel.model_validate_json(request.data)
+    except ValidationError as e:
+        sentry_sdk.capture_exception(e)
+        return jsonify(), 400
+
+    with sentry_sdk.new_scope() as scope:
+        scope.set_context("notify_callback", callback_data.model_dump(mode="json"))
```
MarcUsher, line 66: This is also cool!

app/deliver_grant_funding/routes/api/callbacks.py
```diff
@@ -0,0 +1,85 @@
...
+        not request.authorization
+        or not request.authorization.token
+        or not secrets.compare_digest(
+            request.authorization.token.encode(), current_app.config["GOVUK_NOTIFY_CALLBACK_TOKEN"].encode()
+        )
+    ):
+        return jsonify(), 403
+
+    if request.mimetype != "application/json":
+        sentry_sdk.capture_message(
```
MarcUsher, line 54: Noice I don't think we do this anywhere else/rarely use the sentry_sdk in isolation over current_app.logger - assuming this is cos we don't necessarily need this going to the service logs, we just want sentry to shout at us if any of these various cases happen so we can do something?

MarcUsher (approved):
This is very cool, nice tidy scoped API implementation, feels like a very nice example for other little APIs we might need to bring into the service.

Approving, no nits, comments are just general noodlin questions from me

app/deliver_grant_funding/routes/api/callbacks.py
samuelhwilliams: This should be used over basic string comparison when doing any sensitive text equality checks to avoid [timing attacks](https://sqreen.github.io/DevelopersSecurityBestPractices/timing-attack/python)

app/deliver_grant_funding/routes/api/callbacks.py
samuelhwilliams: yeaaa basically! I don't think we'll ever see this, but if we do then it means GOV.UK Notify is doing something we don't expect and we may need to update our integration

PR #1604, opened 2026-05-01
Fspt 1298 reopen frontend

srh-sloan:
## 🎫 Ticket
FSPT-1298

## 📝 Description
This PR adds the frontend routes and template to be able to reopen a submission:
- New form
- New template
- New route
- Uses the backend logic introduced in the previous PRs, but moves the permissions checks into `AuthorisationHelper` so they can be reused in the template.
- Fixes the notify email body so that the reopened reason correctly indents the text (doing it in the template didn't work if the reason spanned multiple lines)

https://github.com/user-attachments/assets/424c9c6f-b8fe-42d0-be28-4cc4e534fd43

## 🧪 Testing
- Tested locally with test and live submissions

### Review Readiness
- [ ] PR title and description provide sufficient context for reviewers

### Performance and security
- [ ] No N+1 query problems introduced

### Testing

### DIFF
--- a/app/common/auth/authorisation_helper.py
+++ b/app/common/auth/authorisation_helper.py
@@ -4,14 +4,14 @@ from uuid import UUID
 from flask import request
 from flask_login import AnonymousUserMixin
 
-from app.common.data.interfaces.collections import get_collection
+from app.common.data.interfaces.collections import get_collection, get_submission
 from app.common.data.interfaces.grant_recipients import get_grant_recipient
 from app.common.data.interfaces.grants import get_grant
 from app.common.data.models_user import User
 from app.common.data.types import OrganisationModeEnum, RoleEnum
 
 if TYPE_CHECKING:
-    from app.common.data.models import Organisation
+    from app.common.data.models import Organisation, Submission
 
 
 class AuthorisationHelper:
@@ -189,6 +189,26 @@ class AuthorisationHelper:
             return False
 
         return False
+
+    @staticmethod
+    def can_reopen_submission(user: User, submission: "Submission | UUID") -> bool:
+        # Platform admin should be able to reopen submissions
+        # The grant team should be able to reopen submissions. Their roles would be:
+        # - grant member
+        # Form designers should not be able to reopen submissions. Their roles would be:
+        # - deliver org admin / member
+        if isinstance(submission, UUID):
+            submission = get_submission(submission)
+
+        if AuthorisationHelper.is_platform_admin(user):
+            return True
+
+        has_deliver_grant_role = AuthorisationHelper.has_deliver_grant_role(
+            submission.collection.grant.id, RoleEnum.MEMBER, user
+        )
+        is_deliver_org_member = AuthorisationHelper.is_deliver_org_member(user)
+
+        return has_deliver_grant_role and not is_deliver_org_member
 
     @staticmethod
     def has_access_org_access(user: User | AnonymousUserMixin, organisation_id: UUID) -> bool:
--- a/app/deliver_grant_funding/forms.py
+++ b/app/deliver_grant_funding/forms.py
@@ -1307,8 +1307,8 @@ class ReopenSubmissionForm(FlaskForm):
     reopened_reason = TextAreaField(
         "Why are you reopening this submission?",
         validators=[
-            DataRequired(),
-            WordRange(max_words=REASON_MAX_WORDS, field_display_name="description"),
+            DataRequired("Enter the reason for reopening this submission"),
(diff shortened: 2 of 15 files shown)


app/common/auth/authorisation_helper.py
```diff
@@ -190,6 +190,26 @@ def can_edit_collection(user: User | AnonymousUserMixin, collection_id: UUID) ->
...
 
+    @staticmethod
+    def can_reopen_submission(user: User, submission: "Submission | UUID") -> bool:
+        # Platform admin should be able to reopen submissions
+        # The grant team should be able to reopen submissions. Their roles would be:
+        # - grant member
+        # Form designers should not be able to reopen submissions. Their roles would be:
+        # - deliver org admin / member
+        if isinstance(submission, UUID):
+            submission = get_submission(submission)
```
samuelhwilliams, line 201: nit: one to keep an eye on - this would emit SQL every time it's called because it will bypass the SQLAlchemy object cache. I think we might in the future benefit from a `get_submission_by_id` which does a PK lookup to benefit from caching.

However we don't pass a `submission_id` in anywhere currently.

app/deliver_grant_funding/templates/deliver_grant_funding/reports/reopen_submission.html
```diff
@@ -0,0 +1,63 @@
...
+          {% endif %}
+        </ul>
+      {% endset %}
+      <form method="post" novalidate>
+        {{ form.csrf_token }}
+        {{
+          form.reopened_reason(
+                    params={
+                      "label":{
+                        "isPageHeading":true,
```
samuelhwilliams: This wraps the label in an `<h1>` block, but you're passing in another `<h1>` block - so this is a bit broken.

I think just removing the `<h1>` wrapper from the label_html should be right.

app/deliver_grant_funding/templates/deliver_grant_funding/reports/reopen_submission.html
```diff
@@ -0,0 +1,63 @@
...
+                        "html": label_html
+                      },
+                      "hint":{
+                        "html":hint_html
+                      },
+                      "maxwords": form.REASON_MAX_WORDS
+                    }
+          )
+        }}
+        {{ form.submit }}
```
samuelhwilliams: I think this page generally needs a bit of restructuring to align with https://mhclgdigital.atlassian.net/wiki/spaces/FS/pages/599228418/How+we+structure+question+pages+and+Change+pages#Using-paragraph-text

We need to show paragraph text on the page, which means we need to use separate headings and form field labels rather than stuffing things inside the hint text. Input hint text is recommended to be a single sentence; here we're using multiple sentences and a list.

We can probably handle this on the tech side without needing too much content reworking, but we may need to get separate visually-hidden text for the input label.

app/deliver_grant_funding/forms.py
```diff
@@ -1300,3 +1300,16 @@ class SelectConditionCalculationForm(FlaskForm):
...
     submit = SubmitField("Continue", widget=GovSubmitInput())
+
+
+class ReopenSubmissionForm(FlaskForm):
+    REASON_MAX_WORDS = 200
+    reopened_reason = TextAreaField(
+        "Why are you reopening this submission?",
+        validators=[
+            DataRequired("Enter the reason for reopening this submission"),
+            WordRange(max_words=REASON_MAX_WORDS, field_display_name="description"),
```
samuelhwilliams: content nit:

```suggestion
            WordRange(max_words=REASON_MAX_WORDS, field_display_name="reason for reopening the submission"),
```

tests/unit/common/auth/test_authorisation_helper.py
```diff
@@ -449,3 +449,21 @@ def test_has_access_grant_recipient_role_rejects_deliver_only_user(self, factori
         factories.user_role.build(user=user, permissions=[RoleEnum.MEMBER], organisation=organisation, grant=None)
 
         assert AuthorisationHelper.has_access_grant_recipient_role(user=user) is False
+
+    @pytest.mark.parametrize("user_fixture", ["platform_member_user", "deliver_org_admin_user"])
```
samuelhwilliams, line 453: Not sure I feel quite as good about these new user fixtures as the other ones - I think it becomes a bit less clear to me when to use these directly, when to use the authenticated client users, and what you get with them - these come associated with submitted submissions but that feels really unclear and like a bit of an unexpected dependency/setup.

I don't really want to block this PR on changing these, but this is definitely starting to feel a bit intertwined and hard to follow.

samuelhwilliams (approved):

PR #1611, opened 2026-05-06
FSPT-1298 - Reopen info in admin views

srh-sloan:
Couple of small tweaks to the admin views following the reopen submissions work:

## Adding the reopened reason to the submission timeline
Screenshot 2026-05-06 at 14 57 46

## Formatting the event data as JSON for Submission Events
Screenshot 2026-05-06 at 14 58 34

### DIFF
--- a/app/deliver_grant_funding/admin/entities.py
+++ b/app/deliver_grant_funding/admin/entities.py
@@ -638,3 +638,6 @@ class PlatformAdminSubmissionEventView(FlaskAdminPlatformAdminAccessibleMixin, P
         "submission.mode": "Submission mode",
         "created_by.email": "Created by",
     }
+    column_formatters_detail = {
+        "data": _format_json_data,
+    }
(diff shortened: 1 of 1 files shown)


app/deliver_grant_funding/admin/entities.py
```diff
@@ -553,7 +553,10 @@ def _build_submission_timeline_items(submission: Submission) -> list[MoJTimeline
             title += f": {form_titles.get(event.related_entity_id, '(form no longer exists)')}"
         elif event.event_type == SubmissionEventType.SUBMISSION_DECLINED_BY_CERTIFIER:
             description = event.data.get("declined_reason") or ""
-
+        elif event.event_type == SubmissionEventType.SUBMISSION_REOPENED:
+            description = ""
```
samuelhwilliams: Nice use of markupsafe escape to keep things safe 🙏

nit: if we make this a Markup object, we should be able to avoid using |safe in the template (which is a bit of a code smell for me)

samuelhwilliams (approved):

PR #1639, opened 2026-05-13
Make app deploys match terraform values for CPU and memory

srh-sloan:
Fixing a small issue whereby our terraform code sets different values for CPU and memory depending on the target environment (dev vs test/prod) but our app deploy always uses the prod values.

--- 

Parameterising the values for cpu and memory on the deployed app runner image. 

This means that when we deploy to dev we match our terraform settings of having different values for dev and test/prod. Without this, every time we run terraform in dev it is generating changes to the cpu and memory values which is unnecessary noise for other deployments

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,10 +14,22 @@ on:
         required: true
         default: dev
       run_e2e_tests:
-          description: "Run end-to-end tests?"
-          type: boolean
-          default: true
-          required: true
+        description: "Run end-to-end tests?"
+        type: boolean
+        default: true
+        required: true
+      image_cpu:
+        # Keep in sync with terraform otherwise this value will be reverted on the next terraform apply
+        description: "CPU units for the App Runner service"
+        type: number
+        default: 2
+        required: false
+      image_memory:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -14,10 +14,20 @@ on:
...
-          description: "Run end-to-end tests?"
-          type: boolean
-          default: true
-          required: true
+        description: "Run end-to-end tests?"
+        type: boolean
+        default: true
+        required: true
+      image_cpu:
+        description: "CPU units for the App Runner service"
```
samuelhwilliams, line 23: Maybe a tiny comment here saying this needs to be kept in sync per environment with our terraform config?

Or can we make terraform ignore cpu/memory changes with some clever config so that this is the source of truth?

Or maybe we just don't care to do any of this because we're switching to ECS soon(tm)

samuelhwilliams (approved):

.github/workflows/deploy.yml
srh-sloan: Added a comment to keep in sync. I'm on the side of 'we are moving off ECS so do a small change to fix it not a complicated one to prevent future issues'

samuelhwilliams (approved):

PR #1643, opened 2026-05-14
Fspt 1344 data source factories

srh-sloan:
## 🎫 Ticket
[FSPT-1344](https://mhclgdigital.atlassian.net/browse/FSPT-1344): Add default columns for grant recipient data sources in factories

## 📝 Description
    Updating the DataSourceFactory to create a grant recipient data source if `type==DataSourceType.GRANT_RECIPIENT`.
    - uses _adjust_kwargs to set defaults for certain attributes if type is grant_recipient
    - changes items to use a `factory.Maybe` to determine whether to create related DataSourceItems. This isn't in adjust_kwargs as that doesn't work nicely with things that need the id of the object created (`data_source_item` `needs data_source_id`)

- Also updates any existing tests that were creating a simple grant recipient data source with a single column to just use the new defaults

[FSPT-1344]: https://mhclgdigital.atlassian.net/browse/FSPT-1344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/common/data/test_models.py
+++ b/tests/integration/common/data/test_models.py
@@ -750,7 +750,7 @@ class TestComponentReferenceModel:
         grant = factories.grant.create()
         collection = factories.collection.create(grant=grant)
         data_source = factories.data_source.create(
-            name="Budget data", grant=grant, collection=collection, type=DataSourceType.GRANT_RECIPIENT, items=None
+            name="Budget data", grant=grant, collection=collection, type=DataSourceType.GRANT_RECIPIENT
         )
         question = factories.question.create(form__collection=collection)
 
@@ -780,7 +780,10 @@ class TestComponentReferenceModel:
         grant = factories.grant.create()
         collection = factories.collection.create(grant=grant)
         data_source = factories.data_source.create(
-            name="Budget data", grant=grant, collection=collection, type=DataSourceType.GRANT_RECIPIENT, items=None
+            name="Budget data",
+            grant=grant,
+            collection=collection,
+            type=DataSourceType.GRANT_RECIPIENT,
         )
         q1 = factories.question.create(form__collection=collection)
         q2 = factories.question.create(form=q1.form)
@@ -800,7 +803,10 @@ class TestComponentReferenceModel:
         grant = factories.grant.create()
         collection = factories.collection.create(grant=grant)
         data_source = factories.data_source.create(
-            name="Budget data", grant=grant, collection=collection, type=DataSourceType.GRANT_RECIPIENT, items=None
+            name="Budget data",
+            grant=grant,
+            collection=collection,
+            type=DataSourceType.GRANT_RECIPIENT,
         )
         question = factories.question.create(form__collection=collection)
 
@@ -819,7 +825,10 @@ class TestComponentReferenceModel:
         grant = factories.grant.create()
         collection = factories.collection.create(grant=grant)
         data_source = factories.data_source.create(
-            name="Budget data", grant=grant, collection=collection, type=DataSourceType.GRANT_RECIPIENT, items=None
+            name="Budget data",
+            grant=grant,
+            collection=collection,
+            type=DataSourceType.GRANT_RECIPIENT,
         )
         question = factories.question.create(form__collection=collection)
         db_session.add(
@@ -844,7 +853,10 @@ class TestDataSourceModel:
         report = factories.collection.create(grant=grant)
         org1, org2, org3, org4 = factories.organisation.create_batch(4)
         data_source = factories.data_source.create(
-            name="Test data set", grant=grant, collection=report, type=DataSourceType.GRANT_RECIPIENT, items=None
+            name="Test data set",
+            grant=grant,
+            collection=report,
+            type=DataSourceType.GRANT_RECIPIENT,
         )
         org1_item = factories.data_source_organisation_item.create(
             data_source=data_source, external_id=org1.external_id
--- a/tests/models.py
+++ b/tests/models.py
@@ -13,6 +13,7 @@ from dataclasses import field
 from typing import Any, cast
 from uuid import uuid4
 
+import factory
 import factory.fuzzy
 import faker
 from factory.alchemy import SQLAlchemyModelFactory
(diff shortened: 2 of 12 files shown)


tests/models.py
```diff
@@ -890,6 +891,26 @@ class Meta:
...
+    }
+)
+
+
+class _GrantRecipientDataSourceFactory(_DataSourceFactory):
+    type = DataSourceType.GRANT_RECIPIENT
+    name = "Test data set"
+    schema = _GRANT_RECIPIENT_DEFAULT_SCHEMA
+    file_metadata = DataSourceFileMetadata.model_validate({"s3_key": "file/key", "original_filename": "test-file.csv"})
+    items = factory.RelatedFactoryList(_DataSourceItemFactory, size=0, factory_related_name="data_source")
```
MarcUsher: I think this can just be `items = None` as GR data sources shouldn't ever have normal DataSourceItems, but might have DataSourceOrganisationItems?

MarcUsher: It's sooo nice how much this tidies up the other tests, but just wondering about future test writing if it'd be obvious to use this one or if people would still create using the DataSource factory and then just add a bunch of noise again? I think there are already a few defaults using lambdas if the DataSource being created isn't CUSTOM (aka the normal radio/checkbox data sources), could maybe just extend these lambda defaults to also create the schema and anything else needed if we're explicitly creating a grant recipient data source, which would make it easier for future tests to all point to the same place?

MarcUsher (commented):
Thanks for picking this up Sarah! Cuts down on soooo much noise in tests.

Added one comment re: where we're putting this factory/defaults, happy for you to push back on that though! 

And there are maybe a couple more tests that could be cleaned up, a couple in test_data_sets.py and common/expressions/test_forms.py and test_init.py (though a lot of the latter mutate the thing on purpose to check error states etc so maybe not many that need swapping out there), and quite a few in test_reports.

tests/models.py
```diff
@@ -890,6 +891,26 @@ class Meta:
...
+            data_options=QuestionDataOptions(number_type=NumberTypeEnum.INTEGER),
+            original_column_name="Allocation",
+        )
+    }
+)
+
+
+class _GrantRecipientDataSourceFactory(_DataSourceFactory):
+    type = DataSourceType.GRANT_RECIPIENT
+    name = "Test data set"
```
samuelhwilliams: nit: Maybe call this `Grant allocations`, just to keep things grounded in the domain a bit more? We often use the suffix `data set` on the page as well, so this would render as `Test data set data set` which just reads awkwardly if we ever come across it.

samuelhwilliams (commented):
Nice tidy up, this feels a lot cleaner and simpler 🙏 Just one comment to think about.

tests/models.py
```diff
@@ -873,23 +887,43 @@ class Meta:
...
         factory_related_name="data_source",
     )
+    items = factory.Maybe(
+        factory.LazyAttribute(lambda o: o.type == DataSourceType.CUSTOM),
+        yes_declaration=factory.RelatedFactoryList(
+            _DataSourceItemFactory,
+            factory_related_name="data_source",  # Triggers parent relation mapping
+            size=3,
+            # Pass the database-generated ID to the child factory
+            data_source_id=factory.SelfAttribute("..id"),
```
MarcUsher: This is new! Is it needed if the DataSourceItemFactory explicitly sets the data_source_id to be the ID of the `data_source` if one exists? Is this just to be explicit/intentional?

tests/models.py
```diff
@@ -873,23 +887,43 @@ class Meta:
...
+                kwargs.setdefault("schema", _GRANT_RECIPIENT_DEFAULT_SCHEMA)
+
+                kwargs.setdefault(
+                    "file_metadata",
+                    DataSourceFileMetadata.model_validate({"s3_key": "file/key", "original_filename": "test-file.csv"}),
+                )
+            case DataSourceType.CUSTOM:
+                pass
+
+        return kwargs
```
MarcUsher, line 924: Love this, so tidy 🧑‍🍳

tests/models.py
```diff
@@ -873,23 +887,43 @@ class Meta:
...
     type = DataSourceType.CUSTOM
-    name = factory.LazyAttribute(lambda o: None if o.type == DataSourceType.CUSTOM else "Test data set")
-    schema = factory.LazyAttribute(lambda o: None if o.type == DataSourceType.CUSTOM else DataSourceSchema({}))
-    file_metadata = factory.LazyAttribute(
-        lambda o: (
-            None
-            if o.type == DataSourceType.CUSTOM
-            else DataSourceFileMetadata.model_validate({"s3_key": "file/key", "original_filename": "test-file.csv"})
-        )
-    )
```
MarcUsher, line 884: Nit: Do we still want to include these in the factory, just defaulting to `None`, with a comment explaining why they're there/how the _adjust_kwargs is working? Not needed really, just wonder if it makes the factory a little more self documenting.

MarcUsher (approved):
This feels really tidy, a lot less to think about when creating a data source with the factory now and it reads really clearly!

A couple of small nits below, and maybe some additional tests that can be cleaned up in a few other places - test_reports, test_data_sets and test_init for expressions? Approving though if you wanted to get it in and update these later.

tests/models.py
srh-sloan: I did think this, but it actually breaks it with the `adjust_kwargs`. Because we are setting the default value in `kwargs`, that only takes effect if there are no other values set for something. Setting `name=None` is providing a value, so our value of `Grant allocation` is never used.

tests/models.py
srh-sloan: Good spot, not needed 👍

tests/integration/common/data/test_models.py
MarcUsher, line 1: Small nit but we can prob drop the `name=...` properties on these data sources?

MarcUsher (approved):
Tests are so much easier to read and follow now with this 🙌

PR #1644, opened 2026-05-15
E2e/refactor separate reports - part 1

srh-sloan:
First PR for restructuring E2E tests:
- adds more typed dicts for specifying report and section information
- Updates existing tests to work with this new data structure

Still to come:
- Create a single grant
- Separate out the existing test report into different reports so the tests are smaller and more targeted, and they are all created in the same grant
- Fix remaining warnings in the test report data structures


samuelhwilliams (approved):
Nice clean ups, great to see the big yellow squiggles gone 🙏

PR #1675, opened 2026-05-27
Platform admin page for reviewing certifiers

samuelhwilliams:
## 🎫 Ticket
N/A

## 📝 Description
Requires https://github.com/communitiesuk/funding-service-terraform/pull/107

We maintain our own list of certifiers seeded initially from Delta's list.

Delta maintains their list actively and makes an export available nightly. This patch loads their export and provides a platform admin page to compare their list of certifiers with our own to aid in keeping our list up-to-date and reviewing incoming support requests to check whether people should be authorised to certify reports.

This page is a bare MVP and could be improved in multiple ways:

- Sorting 
- Filtering 
- An action to easily sync/set up certifiers from Delta's list

But these are left for future work to keep this a simple readonly page for now, as there are other priorities.

This purposely avoids a lot of thinking around the processes for updating certifiers, instead choosing to just expose the raw information. We can't easily automate the process until it's well known, understood and defined. Hopefully this will enable us to get that more closely locked down and facilitate future automation. It's also likely to be somewhat unperformant in deployed environments because of doing a lot of work (pull from S3, unzip, read+process the full list, etc). Improvements left for the future due to other priorities and the limited blast radius of this change (internal to our team only).

image

image

## 🧪 Testing
Should be able to test this locally (albeit rubbish data) and as of 8am 27th May it's deployed on dev where it can check against Delta's (rubbish) staging data. The table will be full of rubbish but you can validate that it loads and see the diffs.

### Review Readiness

### Testing

### DIFF
--- a/app/deliver_grant_funding/admin/views.py
+++ b/app/deliver_grant_funding/admin/views.py
@@ -1205,10 +1205,10 @@ def _build_delegated_rows(
 ) -> list[DelegatedRow]:
     fs_grant_certifiers_by_org: dict[Organisation, dict[str, _FSGrantCertifier]] = {}
     for grant in get_all_grants():
-        override_certifiers = get_grant_override_certifiers_by_organisation(grant_id=grant.id)
+        override_certifiers = get_grant_override_certifiers_by_organisation(
+            grant_id=grant.id, organisation_mode=OrganisationModeEnum.LIVE
+        )
         for fs_org, users in override_certifiers.items():
-            if fs_org.mode != OrganisationModeEnum.LIVE:
-                continue
             for user in users:
                 email = user.email.lower()
                 entry = fs_grant_certifiers_by_org.setdefault(fs_org, {}).setdefault(
@@ -1262,7 +1262,7 @@ class PlatformAdminDeltaCertifiersView(FlaskAdminPlatformAdminGrantLifecycleMana
         certifiers_by_org = get_certifiers_by_organisation()
 
         return self.render(
-            "deliver_grant_funding/admin/certifiers.html",
+            "deliver_grant_funding/admin/delta-certifiers.html",
             org_rows=_build_org_certifier_rows(delta_officers, certifiers_by_org),
             delegated_rows=_build_delegated_rows(delta_officers, certifiers_by_org),
             delta_last_updated_at=delta_last_updated_at,
--- a/app/deliver_grant_funding/admin/views.py
+++ b/app/deliver_grant_funding/admin/views.py
@@ -1,10 +1,16 @@
 import csv
+import datetime
+import os
+import zipfile
+from dataclasses import dataclass
 from io import BytesIO, StringIO
-from typing import Any
+from typing import TYPE_CHECKING, Any, Literal, Sequence, TypedDict, cast
 from uuid import UUID
 
+import boto3
 import markupsafe
 from flask import abort, current_app, flash, make_response, redirect, send_file, url_for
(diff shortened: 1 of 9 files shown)


app/deliver_grant_funding/admin/views.py
```diff
@@ -1086,3 +1096,174 @@ def stop(self) -> Any:
...
+    path_or_key = current_app.config["DELTA_S151_CSV_PATH_OR_KEY"]
+    if path_or_key.startswith("s3://"):
+        bucket, _, key = path_or_key.removeprefix("s3://").partition("/")
+        s3_object = boto3.client("s3").get_object(Bucket=bucket, Key=key)
+        last_updated_at = s3_object["LastModified"]
+        zip_bytes = s3_object["Body"].read()
+        with zipfile.ZipFile(BytesIO(zip_bytes)) as archive:
+            csv_name = next(n for n in archive.namelist() if n.endswith(".csv"))
+            csv_text = archive.read(csv_name).decode("utf-8-sig")
+    else:
```
sfount, line 1146: ~Not a problem in any way but I think OK to require this to come from S3 and start avoiding any workflows that involve developers managing this file locally, with aws-vault it should be fairly straight forward to point it at dev or similar to verify behaviour locally~

E: I can see you've put in a local version to compare against the local seed, that feels sensible for this not becoming a feature that devs don't interact with

app/deliver_grant_funding/admin/views.py
```diff
@@ -1086,3 +1096,174 @@ def stop(self) -> Any:
...
+    organisation_code: str
+    organisation_name: str
+    officer_type: Literal["s151-officer", "deputy-s151-officer", "delegated-s151-officer"]
+    email: str
+    name: str
+    delegated_access_groups: str
+    last_modified_date: datetime.datetime
+
+
+class OrgCertifierRow(TypedDict):
```
sfount, line 1112: 🍬 these make following the template much nicer

app/deliver_grant_funding/templates/deliver_grant_funding/admin/certifiers.html
```diff
@@ -0,0 +1,148 @@
...
+    </tbody>
+  </table>
+{% endset %}
+
+{% block action_panel %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      {{ layout.messages() }}
+
+      <h1 class="govuk-heading-l">Certifiers</h1>
```
sfount: Smallest nit for this and the tab name is maybe something like "Delta certifiers" as this isn't the place you should go to set up or manage certifiers otherwise - thats all done contextually in the lifecycle tasklist so I don't think anything else would need this name so its not a problem

sfount (approved):

app/deliver_grant_funding/templates/deliver_grant_funding/admin/certifiers.html
samuelhwilliams: I did start with Delta Certifiers here (I think there's some crumbs left in the code) but then I think in my head the v2 of this page would give you an action to actually set up/sync the certifiers from this tab so I changed it.

Happy with either, will merge this so as not to re-review and throw up another PR for us to take/leave.

samuelhwilliams: E: nvm I fixed one thing up for copilot so I did this as well

sfount (approved):

PR #1673, opened 2026-05-26
FSPT-1337: Remove static & project level DS logic

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1337

## 📝 Description
Now we know we're not going to support Static or Project Level data sets in the short term and the upload journey in the UI only allows Grant Recipient data sets to be uploaded, we can safely remove the logic that caters for the other two types to avoid redundant code - if we return to this work and extend it in future we can put some of this logic back or, better yet, implement it fresh when it's tightly scoped.

A subsequent PR (https://github.com/communitiesuk/funding-service/pull/1677) will deal with dropping the enum values and data migration - keeping those separate to avoid any gnarly instances of enum values being dropped and old code still running on the deployed app.

### Review Readiness
- [X] PR title and description provide sufficient context for reviewers
- [X] Commits are logical, self-contained, and have clear descriptions

### Testing
- [X] I have tested this change and it meets the acceptance criteria for the ticket
- ~[ ] I need the reviewer(s) to pull and run this change locally~
- ~[ ] New (non-developer) functionality has appropriate unit and integration tests~
- [X] End-to-end tests have been updated (if applicable)
- ~[ ] Edge cases and error conditions are tested~

PR #1677, opened 2026-05-27
FSPT-1337: Drop unneeded data set type enum values

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1337

## 📝 Description
We're not supporting STATIC or PROJECT_LEVEL data sets for the time being so now that all code and logic referencing these is removed from the service (https://github.com/communitiesuk/funding-service/pull/1673) and they can't be created we can safely drop the values from the enum and accompanying logic in our models.

## 🧪 Testing
DB should be able to go up and down safely with this migration, any lingering STATIC or PROJECT_LEVEL data sources should be removed as part of the upgrade.

### Review Readiness

### Testing
- ~[ ] Edge cases and error conditions are tested~

### DIFF
--- a/app/common/data/migrations/versions/064_remove_data_set_types.py
+++ b/app/common/data/migrations/versions/064_remove_data_set_types.py
@@ -1,0 +1,74 @@
+"""Remove redundant data set types
+
+Revision ID: 064_remove_data_set_types
+Revises: 063_submission_status_not_null
+Create Date: 2026-05-26 16:14:58.306592
+
+"""
+
+from alembic import op
+from alembic_postgresql_enum import TableReference
+
+revision = "064_remove_data_set_types"
+down_revision = "063_submission_status_not_null"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("data_source", schema=None) as batch_op:
+        batch_op.drop_constraint(
+            "ck_data_source_non_custom_requires_name_grant_collection_and_schema_and_file_metadata",
+            type_="check",
+        )
+
+    # I've checked and we only have GRANT_RECIPIENT and CUSTOM data sources in deployed envs, so this is for safety/
+    # local dev just in case people have STATIC or PROJECT_LEVEL data sources knocking around
+    op.execute(
+        "DELETE FROM data_source_item WHERE data_source_id IN (SELECT id FROM data_source WHERE type = 'STATIC')"
+    )
+    op.execute("DELETE FROM data_source WHERE type IN ('STATIC', 'PROJECT_LEVEL')")
+
+    op.sync_enum_values(  # ty: ignore[unresolved-attribute]
+        enum_schema="public",
+        enum_name="data_source_type_enum",
+        new_values=["CUSTOM", "GRANT_RECIPIENT"],
+        affected_columns=[TableReference(table_schema="public", table_name="data_source", column_name="type")],
+        enum_values_to_rename=[],
+    )
+
+    with op.batch_alter_table("data_source", schema=None) as batch_op:
+        batch_op.create_check_constraint(
+            "ck_data_source_non_custom_requires_name_grant_collection_and_schema_and_file_metadata",
+            (
+                "type = 'CUSTOM' OR "
+                "(name IS NOT NULL AND grant_id IS NOT NULL AND collection_id IS NOT NULL "
+                "AND schema IS NOT NULL AND file_metadata IS NOT NULL)"
+            ),
+        )
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("data_source", schema=None) as batch_op:
+        batch_op.drop_constraint(
+            "ck_data_source_non_custom_requires_name_grant_collection_and_schema_and_file_metadata",
+            type_="check",
+        )
+
+    op.sync_enum_values(  # ty: ignore[unresolved-attribute]
+        enum_schema="public",
+        enum_name="data_source_type_enum",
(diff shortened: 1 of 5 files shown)


app/common/data/models.py
```diff
@@ -1147,23 +1147,19 @@ def data_source_id(self) -> uuid.UUID:
...
 
-        2D (GRANT_RECIPIENT or STATIC, though STATIC is bit of a weird one): dict -> dict[str, DataSourceAnswerTypes]
-        3D (PROJECT_LEVEL): list[dict] -> list[dict[str, DataSourceAnswerTypes]]
+        2D (GRANT_RECIPIENT): dict -> dict[str, DataSourceAnswerTypes]
 
         The typed answers carry the presentation & data options (prefix, suffix etc.) so that we can call all the usual
         answer methods eg. get_value_for_interpolation().
         """
         if not self.schema:
             return {} if isinstance(data, dict) else []
```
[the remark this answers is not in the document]
MarcUsher: Have fixed up

app/common/data/migrations/versions/064_remove_data_set_types.py
```diff
@@ -0,0 +1,71 @@
...
+def upgrade() -> None:
+    with op.batch_alter_table("data_source", schema=None) as batch_op:
+        batch_op.drop_constraint(
+            "ck_data_source_non_custom_requires_name_grant_collection_and_schema_and_file_metadata",
+            type_="check",
+        )
+
+    # I've checked and we only have GRANT_RECIPIENT and CUSTOM data sources in deployed envs, so this is for safety/
+    # local dev just in case
+    op.execute("DELETE FROM data_source WHERE type IN ('STATIC', 'PROJECT_LEVEL')")
```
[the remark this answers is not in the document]
MarcUsher: This would be needed if STATIC and PROJECT_LEVEL types had ever been widely used but it was never possible to reference either of these data source types in the service so there won't be any component refs or static data sources attached to components - this line is just for local dev to clear any lingering STATIC or PROJECT_LEVEL data sources uploaded by devs previously.

I have added a line removing data source items for STATIC data sources, as these don't cascade delete (unlike org items for PROJECT_LEVEL which do).

app/common/data/migrations/versions/064_remove_data_set_types.py
```diff
@@ -0,0 +1,74 @@
...
+
+def upgrade() -> None:
+    with op.batch_alter_table("data_source", schema=None) as batch_op:
+        batch_op.drop_constraint(
+            "ck_data_source_non_custom_requires_name_grant_collection_and_schema_and_file_metadata",
+            type_="check",
+        )
+
+    # I've checked and we only have GRANT_RECIPIENT and CUSTOM data sources in deployed envs, so this is for safety/
+    # local dev just in case people have STATIC or GRANT_RECIPIENT data sources knocking around
```
srh-sloan: Should this say `STATIC or PROJECT_LEVEL`?

app/common/data/migrations/versions/064_remove_data_set_types.py
```diff
@@ -0,0 +1,74 @@
...
+    op.execute(
+        "DELETE FROM data_source_item WHERE data_source_id IN (SELECT id FROM data_source WHERE type = 'STATIC')"
```
srh-sloan, line 28: Do we need `PROJECT_LEVEL` in this where clause or would they not have `data_source_item`s?

app/common/data/migrations/versions/064_remove_data_set_types.py
MarcUsher: Yes it should thank you 🤦

app/common/data/migrations/versions/064_remove_data_set_types.py
MarcUsher: Ah good question, PROJECT_LEVEL would have data source org items (rather than data source items). Org items cascade delete if you delete the data source, but data source items don't (I think because of how they work when you set up radios/checkboxes on a question), so needed to be explicitly deleted here as STATIC data sets have normal data source items as they were always intended to be used for those kinds of list options on questions.

Have tested the upgrade/downgrade and it doesn't touch any existing data source items but maybe worth a sanity check on your side too to make sure? Wouldn't want this migration to blat stuff it shouldn't do when this line is only there for safety for local dev, and there are no static or project_level data sources in deployed envs.

srh-sloan: Have run the up/downgrade locally and all looks good 👍

srh-sloan (approved):

app/common/data/migrations/versions/064_remove_data_set_types.py
```diff
@@ -0,0 +1,74 @@
...
+            "ck_data_source_non_custom_requires_name_grant_collection_and_schema_and_file_metadata",
+            type_="check",
+        )
+
+    # I've checked and we only have GRANT_RECIPIENT and CUSTOM data sources in deployed envs, so this is for safety/
+    # local dev just in case people have STATIC or PROJECT_LEVEL data sources knocking around
+    op.execute(
+        "DELETE FROM data_source_item WHERE data_source_id IN (SELECT id FROM data_source WHERE type = 'STATIC')"
+    )
+    op.execute("DELETE FROM data_source WHERE type IN ('STATIC', 'PROJECT_LEVEL')")
```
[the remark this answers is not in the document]
MarcUsher: Nope, project level data sources never had data source items, only data source org items

PR #1678, opened 2026-05-28
Make uv respect a minimum release age

samuelhwilliams:
Renovate is trying to maintain our lock file by bumping dependencies. It hands off to uv for this, and uv keeps pulling in too-recent deps. This should bring uv in line with renovate.

Eg problem PR https://github.com/communitiesuk/funding-service/pull/1647


MarcUsher (approved):
🧑‍🍳😗👌

MarcUsher (approved):

PR #1681, opened 2026-06-01
Save submission status

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1405

## 📝 Description
We're seeing performance issues on pages like Deliver's "list submissions" page,
which can show the status of 200+ submissions at once.

At the moment we calculate status dynamically on demand, and it's CPU intensive
work that scans all components and evaluates a lot of conditions. For more than
one submission that is causing extremely long page loads (20+ seconds).

This patch starts to save the submission status in the DB whenever a relevant
action happens

- changing an answer
- performing a key event, such as completing forms or certifying the submission

This will then allow us to just read a cached submission status, making certain
pages much more performant.

This change does not affect anything shown in the UI at the moment - it starts filling the `Submission.status` column in the DB but keeps everything in the UI using the dynamically-calculated status.

After this has been deployed we need to run the `populate-status-for-all-submissions` developer command until it's finished all submissions. Then we can retire the dynamic calculation and point it at the saved submission status.

## 🧪 Testing
Run through a submission locally (try out all manner of answering+questions) and check that the submission status shown in the service always matches the submission status in the DB, excluding any 'Overdue' modifier.

### Review Readiness

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -135,6 +135,11 @@ def update_collection(  # noqa: C901
     submission_name_question_id: uuid.UUID | None | TNotProvided = NOT_PROVIDED,
     submission_guidance: str | None | TNotProvided = NOT_PROVIDED,
 ) -> Collection:
+    """Update the various attributes of a collection.
+
+    NOTE: When transitioning a collection's status, this may need to update the status for all of its submissions, eg
+    when closing a report to change any unsubmitted submissions statuses to NOT_SUBMITTED.
+    """
     if name is not NOT_PROVIDED:
         collection.name = name
         collection.slug = slugify(name)
@@ -296,6 +301,17 @@ def update_collection(  # noqa: C901
                         f"You cannot close the report for submissions before "
                         f"the submission period end date of {collection.submission_period_end_date}"
                     )
+
+                from app.common.helpers.collections import SubmissionHelper
+
+                # TODO don't do this here but we need it for _update_submission_status to calculate the new status below
+                collection.status = status
+
+                for submission in collection._submissions:
+                    # TODO replace this with add_submission_event to record that the colleciton was closed
+                    #   But for expediency now, just call _update_submission_status which will calculate the new status
+                    #   based on the collection now being closed
+                    SubmissionHelper(submission)._update_submission_status()
 
             case _:
                 raise StateTransitionError("Collection", collection.status, status)
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -51,6 +51,7 @@ from app.common.data.types import (
     QuestionPresentationOptions,
     SubmissionEventType,
     SubmissionModeEnum,
+    SubmissionStatusEnum,
 )
 from app.common.data.utils import generate_submission_reference
 from app.common.exceptions import WTFormRenderableException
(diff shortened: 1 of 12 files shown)


app/common/data/migrations/versions/062_add_submission_status.py
```diff
@@ -0,0 +1,63 @@
...
+    sa.Enum(
+        "NOT_STARTED",
+        "IN_PROGRESS",
+        "READY_TO_SUBMIT",
+        "AWAITING_SIGN_OFF",
+        "SUBMITTED",
+        "NOT_SUBMITTED",
+        "PARTIALLY_SUBMITTED",
+        name="submission_status_enum",
+    ).create(op.get_bind())
```
srh-sloan: Can we define this enum at the top and then reference elsewhere, instead of listing it out 3 times?

srh-sloan (commented):
Litlte tidy up for the migration

app/common/data/interfaces/collections.py
```diff
@@ -296,6 +302,12 @@ def update_collection(  # noqa: C901
                         f"the submission period end date of {collection.submission_period_end_date}"
                     )
 
+                from app.common.helpers.collections import SubmissionHelper
+
+                for submission in collection._submissions:
+                    if not SubmissionHelper(submission).is_submitted:
+                        update_submission(submission, status=SubmissionStatusEnum.NOT_SUBMITTED)
+
```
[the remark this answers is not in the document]
samuelhwilliams: will be addressed in the future when we read the saved submission status

app/common/data/interfaces/collections.py
```diff
@@ -296,6 +302,12 @@ def update_collection(  # noqa: C901
                         f"the submission period end date of {collection.submission_period_end_date}"
                     )
 
+                from app.common.helpers.collections import SubmissionHelper
+
+                for submission in collection._submissions:
+                    if not SubmissionHelper(submission).is_submitted:
+                        update_submission(submission, status=SubmissionStatusEnum.NOT_SUBMITTED)
```
sfount: As we are loading all of the submission statuses I guess it could just trigger a sync now that the collection status has changed?  to avoid having this direct update even temporarily

Will try and make sure I understand what reading from saved submission status means later

sfount: Brief discussion: we'll likely model things like the report closing and the submission deadline being reached as submission events - this means that this would just be adding an event and the submission itself would take care of the syncing

Some implications on how overdue and other mechanisms work which we'll have to pull through so it would be done as a separate change

For now if we can get this to ask the submission to sync its event rather that setting precedent for any other code calling `update_submission()` that feels good but if we plan to change it anyway there shouldn't be an issue

app/common/helpers/collections.py
```diff
@@ -342,8 +346,38 @@ def get_status(submission: Submission | None, collection: Collection) -> Submiss
 
         return SubmissionHelper(submission).status
 
-    @property
-    def status(self) -> SubmissionStatusEnum:
+    def clear_caches(self):
```
sfount, line 349: Nice this happening centrally feels a lot less sprawling

sfount (commented):
This all looks sensible and well thought through under pressure, persisting the state feels like base in any case and then something to think through on how we undo doing all of the calculations when listing multiple submissions, thinking through the current slight edge cases of Not submitted and Overdue

Code looks good I'll be trying it out a bit locally to help think through any edge cases

sfount (approved):
I've reviewed the behaviour locally against the app and the db and it all looks as expected, this should be good to bring in with the proposed event review following. Great work both 👍

PR #1685, opened 2026-06-02
Protect list_submissions endpoint against overwhelm

samuelhwilliams:
We're currently seeing this endpoint cause large performance issues across the service with a grant that has a lot of submissions. We need to effectively take this offline to protect the service while we're doing the proper fix.

We'll need to take this out ASAP because obviously it's bad. But we need to do the status performance fixes first from FSPT-1405.

### DIFF
--- a/app/deliver_grant_funding/routes/reports.py
+++ b/app/deliver_grant_funding/routes/reports.py
@@ -3131,6 +3131,11 @@ def list_submissions(grant_id: UUID, report_id: UUID, submission_mode: Submissio
             )
         )
 
+    # TODO[FSPT-1405]: remove me; very short term service stability guard
+    submissions = report.live_submissions if submission_mode == SubmissionModeEnum.LIVE else report.test_submissions
+    if len(submissions) > 100:
+        abort(503)
+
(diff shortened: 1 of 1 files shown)


app/deliver_grant_funding/routes/reports.py
```diff
@@ -3133,6 +3133,10 @@ def list_submissions(grant_id: UUID, report_id: UUID, submission_mode: Submissio
 
     helper = AllSubmissionsHelper(collection=report, submission_mode=submission_mode)
 
+    # TODO[FSPT-1405]: remove me; very short term service stability guard
+    if len(helper.submissions) > 100:
```
sfount: To get around the all submission helper doing all the work thats likely causing this to be slow we could maybe fetch the submissions (without full schema etc.) at the top of this using the interface and count those

Then +1 on 503 to indicate it'll come back!

samuelhwilliams: the submission helper doesn't really do any work, but I've changed this anyway for you

srh-sloan (approved):

sfount (approved):

PR #1687, opened 2026-06-03
Read submission status from db

srh-sloan:
## 📝 Description
- Reinstates the list_submissions route for all reports
- Updates the submission_helper to read the status directly from the submission model rather than calculating it dynamically
- Updates the list_submissions endpoint and template to not use submission helpers so we reduce the db load and processing time
- Fixes tests that were relying on the dynamic status calculation

## To do before merge

- [x]  Re-examine the queries and performance times using the profiler after [latest commit ](https://github.com/communitiesuk/funding-service/commit/6756f1d4e73f041ebfd2a6e2472be13364c41d76)

## Future thinking
- Can we get the submission event factory to auto calculate the status and update the submission object?
- Or should we change the tests to setup submissions in the right state in a different way?
- Can maybe tie this in with the thinking we already wanted to do around the submission_xxx fixtures in integration tests
- Should we change the multi submissions path to just use grant recipients and then filter out those with no submissions
- Now we have less-intensive interface methods and a persisted status, can we reinstate the helpers with their improved performance?

### DIFF
--- a/app/common/data/interfaces/grant_recipients.py
+++ b/app/common/data/interfaces/grant_recipients.py
@@ -5,7 +5,7 @@ from sqlalchemy import String, cast, func, select
 from sqlalchemy.orm import joinedload, selectinload
 
 from app.common.data.interfaces.exceptions import flush_and_rollback_on_exceptions
-from app.common.data.models import Grant, GrantRecipient, Organisation
+from app.common.data.models import Grant, GrantRecipient, Organisation, Submission
 from app.common.data.models_user import User, UserRole
 from app.common.data.types import GrantRecipientModeEnum, RoleEnum, SubmissionModeEnum, SubmissionStatusEnum
 from app.extensions import db
@@ -18,6 +18,8 @@ def get_grant_recipients(
     with_data_providers: bool = False,
     with_certifiers: bool = False,
     with_organisations: bool = False,
+    with_submissions_for_collection: uuid.UUID | None = None,
+    submission_mode: SubmissionModeEnum = SubmissionModeEnum.LIVE,
 ) -> Sequence[GrantRecipient]:
     stmt = select(GrantRecipient).where(GrantRecipient.grant_id == grant.id, GrantRecipient.mode == mode)
 
@@ -29,6 +31,15 @@ def get_grant_recipients(
 
     if with_organisations:
         stmt = stmt.options(selectinload(GrantRecipient.organisation))
+
+    if with_submissions_for_collection:
+        stmt = stmt.options(
+            selectinload(
+                GrantRecipient.submissions.and_(Submission.collection_id == with_submissions_for_collection).and_(
+                    Submission.mode == submission_mode
+                )
+            ).joinedload(Submission.events)
+        )
 
     stmt = stmt.join(Organisation, GrantRecipient.organisation_id == Organisation.id).order_by(Organisation.name)
 
--- a/app/common/data/models.py
+++ b/app/common/data/models.py
@@ -396,6 +396,13 @@ class Submission(BaseModel):
     )
 
     @property
+    def last_updated_at_utc(self) -> datetime.datetime:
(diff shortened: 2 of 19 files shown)


app/common/data/models.py
```diff
@@ -395,6 +395,13 @@ class Submission(BaseModel):
         nullable=False,
     )
 
+    @property
+    def last_updated_at_utc(self) -> datetime.datetime:
+        from app.common.helpers.submission_events import SubmissionEventHelper
+
+        event_helper = SubmissionEventHelper(self)
+        return max(filter(None, [event_helper.latest_event_utc, self.updated_at_utc]))
```
samuelhwilliams, line 403: for future us: need to think here carefully about how we load related entities to avoid us triggering additional SQL queries unexpectedly, eg by raising errors on lazy loading by default so that we clearly and explicitly define what data we expect available and where/when.

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_submissions.html
```diff
@@ -106,22 +106,20 @@ <h1 class="govuk-heading-l">
...
           {%
             do rows.append([
               {
                 "html": link_to_submission,
               }, {
-                "text": submission_helper.submission.grant_recipient.organisation.name,
+                "text": submission.grant_recipient.organisation.name,
               }, {
-                "html": submissionStatusTag(submission_helper.status, submission_helper.is_overdue)
+                "html": submissionStatusTag(submission.status,  report.is_closed and not submission.status == enum.submission_status.SUBMITTED),
```
samuelhwilliams, line 120: nit: to think about getting this business logic out of the template and somewhere more encapsulated/testable/reusable.

app/common/data/interfaces/grant_recipients.py
```diff
@@ -30,6 +32,15 @@ def get_grant_recipients(
     if with_organisations:
         stmt = stmt.options(selectinload(GrantRecipient.organisation))
 
+    if with_submissions_for_collection:
+        stmt = stmt.options(
+            selectinload(
+                GrantRecipient.submissions.and_(Submission.collection_id == with_submissions_for_collection).and_(
+                    Submission.mode == submission_mode
+                )
```
samuelhwilliams, line 40: nit: this is maybe a little bit spooky as it does a partial load on the `submissions` relationship, which could come across as a bit confusing when seeing it used in the wild. Compare this with the Collection model which has a `_submissions` relationship to pull everything, and then python-side @propertys to filter that explicitly:

```
    @property
    def preview_submissions(self) -> list[Submission]:
        return list(submission for submission in self._submissions if submission.mode == SubmissionModeEnum.PREVIEW)

    @property
    def test_submissions(self) -> list[Submission]:
        return list(submission for submission in self._submissions if submission.mode == SubmissionModeEnum.TEST)

    @property
    def live_submissions(self) -> list[Submission]:
        return list(submission for submission in self._submissions if submission.mode == SubmissionModeEnum.LIVE)
```

samuelhwilliams (approved):
Much more performant, excellent work on getting this over the line. A few comments just to crystallise things we've mostly already spoken about ✅

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_submissions.html
srh-sloan: Added to list: https://mhclgdigital.atlassian.net/browse/FSPT-1406

app/common/data/models.py
srh-sloan: On the list: https://mhclgdigital.atlassian.net/browse/FSPT-1406

app/common/data/interfaces/grant_recipients.py
srh-sloan: Added to list: https://mhclgdigital.atlassian.net/browse/FSPT-1406

PR #1690, opened 2026-06-04
Multi submission statuses

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1198

## 📝 Description
When a multi-submission report is managed by the service, we automatically fill
in the 'submission name' question. This currently makes the entire submission
appear as 'In progress' (because an answer exists) even though the grant
recipient has not really started doing anything with the submission yet.

This patch makes the SubmissionHelper understand multi-submission question names
are not really user-provided data, and excludes them from submission status
calculations, therefore making a submission that only contains the
submission-name question's answer show as 'Not started'.

We also remove the "Have you completed this section?" question from the check your answers page if the entire section is managed by the service.

This adds a developer command which we can run to re-sync the statuses for
multi-submissions, since these are now stored in the DB.

### List submissions page

| Before | Script | After |
|---|---|---|
| image | image | image |

### Check your answers page (entirely managed form)

| Before | After |
|---|---|
| image | image |

## 🧪 Testing

- On main branch, create/clone a managed multi-submission report to your local and seed a bunch of reports
- See that status for everything is 'In progress' by default
- Run the script on this branch
- See statuses for everything change to 'Not started'

### Review Readiness

### Testing

### DIFF
--- a/app/common/helpers/collections.py
+++ b/app/common/helpers/collections.py
@@ -369,7 +369,13 @@ class SubmissionHelper:
     def _calculate_submission_status(self) -> SubmissionStatusEnum:
         submission_state = self.events.submission_state
 
-        form_statuses = {self.get_status_for_form(form) for form in self.collection.forms}
+        form_statuses_by_id = {
+            form.id: self.get_status_for_form(form)
+            for form in self.collection.forms
+            if not self.form_is_managed_by_service(form)
+        }
+
+        form_statuses = {status for form_id, status in form_statuses_by_id.items()}
         all_forms_completed_or_not_needed = form_statuses <= {
             TasklistSectionStatusEnum.COMPLETED,
             TasklistSectionStatusEnum.NOT_NEEDED,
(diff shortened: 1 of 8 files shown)


app/common/data/models.py
```diff
@@ -300,7 +300,7 @@ def s3_key_prefix(self, submission_mode: SubmissionModeEnum) -> str:
             name="ck_monitoring_certification_not_null",
         ),
         CheckConstraint(
-            "submission_name_question_id IS NULL OR allow_multiple_submissions = true",
+            "(submission_name_question_id IS NOT NULL) = (allow_multiple_submissions = true)",
             name="ck_submission_name_question_requires_multiple_submissions",
         ),
```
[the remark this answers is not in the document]
samuelhwilliams: this isnt true

app/common/helpers/collections.py
```diff
@@ -519,6 +525,29 @@ def format_s3_key(self, question_id: uuid.UUID, add_another_index: int | None =
...
+
+        submission_name_form = self.collection.submission_name_question.form
+        if form.id != submission_name_form.id:
+            return False
+
+        submission_name_form_questions = self.cached_get_ordered_visible_questions(submission_name_form)
+        if (
+            len(submission_name_form_questions) == 1
+            and submission_name_form_questions[0] == self.collection.submission_name_question
+        ):
```
sfount, line 546: Nice this feels like a sensible place to have this logic to not have it distributed around the code base

app/developers/commands.py
```diff
@@ -941,20 +941,53 @@ def emit(line: str = "") -> None:
...
+        submission = helper.submission
+
+        last_updated_at = submission.updated_at_utc
+        from_status = helper.status
+
+        # Manually assign updated_at_utc to prevent it being automatically updated
+        db.session.execute(
+            update(Submission)
+            .where(Submission.id == submission_id)
+            .values(updated_at_utc=last_updated_at, status=helper._calculate_submission_status())
```
sfount, line 967: I presume you've given this a go I'm not actually 100% sure who would win between the on update to set the property and explicitly setting it, I'm guessing as its managed by the ORM it would respect the value passed in above any defaults

sfount (approved):

app/developers/commands.py
samuelhwilliams: Yeah I've tested this (and the assertion below does catch problems if you fiddle with it - happy to hop on a call and demonstrate)

PR #1691, opened 2026-06-08
List submissions optimisations

samuelhwilliams:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1406

## 📝 Description
When a collection has a lot of submissions, the 'list submissions' page was
having to pull back a lot of data from the DB to correctly calculate statuses,
submission names, last updated at timestamps, etc. We were seeing the DB
roundtrip alone take 600ms+ for one of our larger collections.

In reality all we need to display this page is:

- grant recipient name
- submission status 
- submission last updated at (including submission events)
- for multi-submission: the submission's name (based on an answer)

This patch optimises the endpoint to just pull this specific information from
the DB instead of the submission and a huge swathe of rleated entities (grant
recipient, organisation, submission events, all of the data from the submission)
through a few new helper properties in the Submission class:

- `name` - for multi-submission only, this tells SQLAlchemy how to calculate the
  submission's name from SQL (get the collection's 'name' question and find its
  associated answer in the data blob)

- `last_updated_at_utc` - the greatest of the latest timestamp on the submission
  and the latest submission event 

With these calculated in SQL we can just fetch the required attributes directly
and skip sending huge swathes of JSON data and related entities from the DB and
processing them in python.

This should get us down to a reliably fast list submissions endpoint, with the
caveat that the structure of the submission data blob now needs to be understood
by the `Submission.name` property rather than just the SubmissionDataManager.
This optimisation and additional bit of complexity feels acceptable when
contrasted to the performance we'd be leaving on the table by not doing this. We
also feel fairly comfortable with the schema used to store submission data and
its unlikely to churn significantly from now on.

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -468,9 +468,7 @@ def get_submission_list_for_collection(
         "last_updated_at_utc"
     )
 
-    grant_recipient_mode = (
-        GrantRecipientModeEnum.LIVE if submission_mode == SubmissionModeEnum.LIVE else GrantRecipientModeEnum.TEST
-    )
+    grant_recipient_mode = GrantRecipientModeEnum.from_similar(submission_mode)
     name_column = Submission.name.label("name") if collection.allow_multiple_submissions else null().label("name")
     order_by: List[Any] = [Organisation.name]
     if collection.allow_multiple_submissions:
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -2,11 +2,12 @@ import datetime
 import uuid
 from collections.abc import Sequence
 from copy import deepcopy
-from typing import Any, Literal, Never, Protocol, Unpack, overload
+from dataclasses import dataclass
+from typing import Any, List, Literal, Never, Protocol, Unpack, overload
 from uuid import UUID
 
 from flask import current_app
-from sqlalchemy import and_, delete, or_, select, text
+from sqlalchemy import and_, delete, func, null, or_, select, text
 from sqlalchemy.exc import IntegrityError, NoResultFound
 from sqlalchemy.orm import joinedload, selectinload
 
@@ -34,6 +35,7 @@ from app.common.data.models import (
     Grant,
     GrantRecipient,
     Group,
+    Organisation,
     Question,
     Submission,
     SubmissionEvent,
@@ -45,6 +47,7 @@ from app.common.data.types import (
     ConditionsOperator,
     DataSourceType,
     ExpressionType,
+    GrantRecipientModeEnum,
     GrantStatusEnum,
     QuestionDataOptions,
     QuestionDataType,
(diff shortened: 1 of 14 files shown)


app/common/data/interfaces/collections.py
```diff
@@ -358,7 +361,6 @@ def get_all_submissions_with_mode_for_collection(
     *,
     with_full_schema: bool = True,
     with_users: bool = False,
-    with_submission_summary_info: bool = False,
```
srh-sloan, line 361: Nice to remove this now we don't need it 👍

app/common/data/interfaces/collections.py
```diff
@@ -405,15 +407,106 @@ def get_all_submissions_with_mode_for_collection(
...
+        stmt = (
+            select(
+                Organisation.name.label("organisation_name"),
+                submission_name,
+                Submission.id.label("submission_id"),
+                Submission.status,
+                Submission.last_updated_at_utc.label("last_updated_at_utc"),
+            )
+            .select_from(Submission)
+            .outerjoin(GrantRecipient, Submission.grant_recipient_id == GrantRecipient.id)
```
srh-sloan: Should we also filter on `GrantRecipient.mode` like we do for single submissions below?

samuelhwilliams: The mode filtering happens on submission on line 461 - but tweaking this all in light of @sfount's comment to stop us having two separate queries. Coming soon 👀

app/common/data/models.py
```diff
@@ -395,12 +409,24 @@ class Submission(BaseModel):
...
+    @classmethod
+    def _last_updated_at_utc_expression(cls) -> ColumnElement[datetime.datetime]:
+        # Postgres GREATEST ignores NULLs, so this falls back to updated_at_utc when there are no events.
+        return func.greatest(
+            cls.updated_at_utc,
+            select(func.max(SubmissionEvent.created_at_utc))
+            .where(SubmissionEvent.submission_id == cls.id)
+            .correlate(cls)
+            .scalar_subquery(),
+        )
```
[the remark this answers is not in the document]
samuelhwilliams: No: https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-GREATEST-LEAST

app/common/data/types.py
```diff
@@ -97,16 +97,28 @@ class SubmissionModeEnum(enum.StrEnum):
     PREVIEW = "preview"
     LIVE = "live"
 
+    @staticmethod
+    def from_similar(mode: GrantRecipientModeEnum | OrganisationModeEnum) -> SubmissionModeEnum:
```
srh-sloan, line 101: Nice ✨

srh-sloan (approved):
Great stuff!
The performance is so much better and the interface and data access much cleaner. Also nice to tidy up the template and be consistent between multi and single submission reports.

PR #1694, opened 2026-06-09
Fix: Submission list overdue status [FSPT-1408] 

albertkol:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1408

## 📝 Description
Submission list statuses won't the **(Overdue)** tag accurately even if the submission period ended.
It should show the **(Overdue)** tag example: `Awaiting sign off (Overdue)`.

Fix status tag next to the title. It's height is too big.

### Before
Overdue status missing before:
image
Status height is distorted before:
image

### After
Overdue status missing after:
image
Status height is distorted after:
image

## 🧪 Testing
- create a report lifecycle
- create a submission and send it for review
- update report lifecycle, set up the submission end date to a past date
- go to /deliver and check the submission status
- status should show the "Overdue" tag

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/collections.py
+++ b/app/common/data/interfaces/collections.py
@@ -431,6 +431,7 @@ class ListSubmissionData:
     submission_id: UUID | None
     status: SubmissionStatusEnum | None
     last_updated_at_utc: datetime.datetime | None
+    is_overdue: bool | None
 
 
 def get_submission_list_for_collection(
@@ -481,6 +482,7 @@ def get_submission_list_for_collection(
             Submission.id.label("submission_id"),
             Submission.status,
             last_updated_at_utc,
+            Submission.is_overdue.label("is_overdue"),
         )
         .select_from(GrantRecipient)
         .join(Organisation, GrantRecipient.organisation_id == Organisation.id)
@@ -492,6 +494,7 @@ def get_submission_list_for_collection(
                 Submission.mode == submission_mode,
             ),
         )
+        .outerjoin(Collection, Collection.id == Submission.collection_id)
         .outerjoin(latest_event, latest_event.c.submission_id == Submission.id)
         .where(GrantRecipient.grant_id == collection.grant_id, GrantRecipient.mode == grant_recipient_mode)
         .order_by(*order_by)
@@ -503,6 +506,7 @@ def get_submission_list_for_collection(
             submission_id=row.submission_id,
             status=row.status,
             last_updated_at_utc=row.last_updated_at_utc,
+            is_overdue=row.is_overdue,
         )
         for row in db.session.execute(stmt).all()
     ]
--- a/app/common/data/models.py
+++ b/app/common/data/models.py
@@ -511,6 +511,21 @@ class Submission(BaseModel):
             cls.reference,
         )
 
+    @hybrid_property
+    def is_overdue(self) -> bool:
+        # todo: make sure this is resilient to timezones, drift, etc. this is likely something that should
+        #       a batch job decision that is then added as a submission event rather than calculated by the server
+        return self.collection.is_overdue and not self.status == SubmissionStatusEnum.SUBMITTED
+
+    @is_overdue.inplace.expression
+    @classmethod
+    def _is_overdue_expression(cls) -> ColumnElement[bool]:
+        return (
+            Collection.submission_period_end_date.isnot(None)
+            & (Collection.submission_period_end_date < func.timezone("Europe/London", func.now()))
+            & (cls.status != SubmissionStatusEnum.SUBMITTED)
+        )
+
     __table_args__ = (
         CheckConstraint(
             "mode = 'TEST' OR grant_recipient_id IS NOT NULL",
--- a/tests/integration/common/data/test_models.py
+++ b/tests/integration/common/data/test_models.py
@@ -1,8 +1,13 @@
+from datetime import date
+
 import pytest
 from psycopg.errors import ForeignKeyViolation
 from sqlalchemy.exc import IntegrityError
 
 from app import QuestionDataType
+from app.common.collections.types import (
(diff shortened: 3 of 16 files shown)


app/common/data/models.py
```diff
@@ -485,6 +485,12 @@ def _name_expression(cls) -> ColumnElement[str]:
             cls.reference,
         )
 
+    @property
+    def is_overdue(self) -> bool:
+        # todo: make sure this is resilient to timezones, drift, etc. this is likely something that should
+        #       a batch job decision that is then added as a submission event rather than calculated by the server
+        return self.collection.is_overdue and not self.status == SubmissionStatusEnum.SUBMITTED
```
samuelhwilliams: If we make this a `@hybrid_property` then we can query it in SQL and `get_submission_list_for_collection` can include it easily in its own data structure

albertkol: Converted it. And updated `get_submission_list_for_collection` to use the expression.

app/deliver_grant_funding/templates/deliver_grant_funding/reports/list_submissions.html
```diff
@@ -132,9 +132,9 @@ <h1 class="govuk-heading-l">
         {% endif %}
 
         {% if submission.submission_id %}
-          {% do cells.append({"html": submissionStatusTag(submission.status, report.is_overdue and not submission.status == enum.submission_status.SUBMITTED)}) %}
+          {% do cells.append({"html": submissionStatusTag(submission.status, submission.is_overdue)}) %}
         {% else %}
-          {% do cells.append({"html": submissionStatusTag(enum.submission_status.NOT_SUBMITTED if report.is_overdue else enum.submission_status.NOT_STARTED, report.is_closed)}) %}
+          {% do cells.append({"html": submissionStatusTag(enum.submission_status.NOT_SUBMITTED if report.is_closed else enum.submission_status.NOT_STARTED, report.is_overdue)}) %}
```
albertkol: I switched the `is_closed` and `is_overdue` properties here. Please confirm if it's the correct condition, @samuelhwilliams.

samuelhwilliams: this is right, thanks!

tests/integration/common/data/test_models.py
```diff
@@ -79,6 +86,45 @@ def test_grant_recipient_submissions_relationship(self, factories):
...
+                related_entity_id=question.form.id,
+            ),
+            factories.submission_event.create(
+                event_type=SubmissionEventType.SUBMISSION_SUBMITTED,
+                related_entity_id=submission.id,
+            ),
+        ]
+        submission.status = SubmissionStatusEnum.SUBMITTED
+
+        assert submission.is_overdue is False
```
samuelhwilliams, line 128: nit: These can probably all be unit tests using `.build` instead of `.create`, which will make them faster to run and help us keep the test suite fast.

app/common/data/models.py
```diff
@@ -485,6 +485,21 @@ def _name_expression(cls) -> ColumnElement[str]:
...
+
+    @is_overdue.inplace.expression
+    @classmethod
+    def _is_overdue_expression(cls) -> ColumnElement[bool]:
+        return (
+            Collection.submission_period_end_date.isnot(None)
+            & (Collection.submission_period_end_date < func.current_date())
```
samuelhwilliams: nit: We probably need to use a localised current date for Europe/London here 🙏

app/common/data/interfaces/collections.py
```diff
@@ -492,6 +494,7 @@ def get_submission_list_for_collection(
                 Submission.mode == submission_mode,
             ),
         )
+        .outerjoin(Collection, Collection.id == Submission.collection_id)
```
samuelhwilliams, line 497: nit: This can just be a standard join; submissions *always* have a collection.

samuelhwilliams (commented):
Looks like a really clean fix with great test coverage, well done for getting to grips with this so quickly.

A few small nits but then good to go 🙏

app/common/data/interfaces/collections.py
albertkol: I am trying to wrap my head around it a bit with all the `join`'s and `outerjoin`'s.

`join` being an inner join
`outer join` being a left join

A `Submission` might always have a `Collection`. But `GrantRecipient` is the starting point.

- `GrantRecipint` joining `Organisation`, works because it's 1-1.
- We are doing a `left join` on `Submission`, not all `GrandRecipient` have a `Submission` so it can be `NULL`
- So doing an `inner join` between something that can be `NULL` (the `Subscription`) and `Collection` removes the row

And we do show `ListSubmissionData` records without subscription.

So it has to be `outerjoin`.

tests/integration/common/data/test_models.py
albertkol: Changed. Makes sense. But I am unsure when you guys choose to use `create()` and when `build()`.

I guess `build()` for unit tests. `create()` when you need the record in the db, perhaps for some integration tests.

app/common/data/models.py
albertkol: Added.

app/common/data/interfaces/collections.py
samuelhwilliams: You're right - thanks 🙇  I was remembering this being Submission focused from a few commits ago 🙏

tests/integration/common/data/test_models.py
samuelhwilliams: That's right yep - build for unit, create for integration.

Default should be unit, integration only if you want/need the DB involved

samuelhwilliams (approved):

PR #1699, opened 2026-06-11
Tests/datasource factory org items

srh-sloan:
## 🎫 Ticket
Extension to [FSPT-1344](https://mhclgdigital.atlassian.net/browse/FSPT-1344)

## 📝 Description
The above ticket updates the test factories to automatically create the right data for a grant recipient type data source. This PR extends that so that the factory will also create data source org items (the actual values for each grant recipient) for a new data source.

This is enabled by passing in `create_gr_org_items=True` to the factory. The factory assumes the schema is the default one provided by the factory (a single column called `c_allocation`) and adds data for each grant recipient for that column.

The exact data for each grant recipient can be supplied to the factory by passing `create_gr_org_items__data=[1,2,3]`, but if no data is supplied then a random number will be used for each data source organisation item.

The PR also updates the existing tests to use this rather than manually create the rows for each test.

[FSPT-1344]: https://mhclgdigital.atlassian.net/browse/FSPT-1344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/integration/common/data/interfaces/test_data_sets.py
+++ b/tests/integration/common/data/interfaces/test_data_sets.py
@@ -462,9 +462,14 @@ class TestGetDataSource:
             get_data_source(uuid.uuid4())
 
     def test_get_data_source_with_organisation_items(self, db_session, factories, track_sql_queries):
-        data_source = factories.data_source.create(items=None)
-        factories.data_source_organisation_item.create(data_source=data_source, external_id="E123")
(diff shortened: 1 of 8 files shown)


MarcUsher (approved):
Ah this is v clever! Really nice use of the post-generation hook to quickly create these, really straightforward to use as well 👌

PR #1700, opened 2026-06-11
FSPT-1342: Update missing data view and flow

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1342

## 📝 Description
The 'data set missing data' view would previously be displayed at the end of the data set upload flow as a final check. The revised designs now have this appearing after the initial file upload, allowing the user to proceed if they want or go back/cancel and reupload a fixed file rather than having to proceed through the flow.

This simplifies the upload flow significantly as we only have to check for missing data once after the initial file upload, rather than at every stage of data set validation before upload, and also simplifies the 'missing data' view itself as we no longer need to display the original CSV row number.

## 🧪 Testing
Upload a CSV with missing data - you should hit the 'data set missing data' view directly after upload now, as opposed to the end of the flow, and be able to navigate forward and back correctly through the flow. If you upload a CSV with no missing data, clicking 'back' on the map columns page should take you back to the initial upload page rather than the missing data view.

### Review Readiness

### Testing

### DIFF
--- a/app/deliver_grant_funding/routes/collections.py
+++ b/app/deliver_grant_funding/routes/collections.py
@@ -117,6 +117,7 @@ from app.deliver_grant_funding.data_sets import (
     BritishPoundsError,
     DataSetValidationResult,
     build_data_set_upload_s3_key,
+    check_missing_data,
     validate_data_set,
     validate_data_set_grant_recipients,
 )
@@ -3384,12 +3385,17 @@ def _parse_data_set_csv(file_storage: FileStorage) -> tuple[list[str], TUnvalida
     return columns, rows
 
 
-def _load_and_validate_data_set(
-    data_set_data: DataSetUploadSessionModel,
-) -> tuple[TUnvalidatedDataSetRows, DataSetValidationResult]:
+def _load_rows(data_set_data: DataSetUploadSessionModel) -> TUnvalidatedDataSetRows:
     file_bytes = s3_service.download_file(data_set_data.s3_key)
     file_storage = FileStorage(stream=io.BytesIO(file_bytes), filename=data_set_data.original_filename)
     _, rows = _parse_data_set_csv(file_storage)
+    return rows
+
+
+def _load_and_validate_data_set(
+    data_set_data: DataSetUploadSessionModel,
+) -> tuple[TUnvalidatedDataSetRows, DataSetValidationResult]:
+    rows = _load_rows(data_set_data)
     return rows, validate_data_set(data_set_data, rows)
 
 
@@ -3461,6 +3467,20 @@ def upload_data_set(grant_id: UUID, collection_type: CollectionType, collection_
                 collection=collection,
                 form=form,
                 gr_errors=gr_errors,
+            )
+
+        missing_result = check_missing_data(session_data.data_columns, rows)
+        if missing_result.has_missing_data:
+            session_data.has_missing_data = True
+            session["data_set_upload"] = session_data.model_dump(mode="json")
+
+            return redirect(
+                url_for(
+                    "deliver_grant_funding.data_set_missing_data",
+                    grant_id=grant_id,
+                    collection_type=collection_type,
+                    collection_id=collection_id,
+                )
             )
 
         return redirect(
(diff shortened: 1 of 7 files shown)


app/deliver_grant_funding/routes/collections.py
```diff
@@ -3736,44 +3735,15 @@ def data_set_missing_data(grant_id: UUID, collection_type: CollectionType, colle
             )
         )
 
-    rows, validation_result = _load_and_validate_data_set(data_set_data)
+    rows = _load_rows(data_set_data)
+    missing_result = check_missing_data(data_set_data.data_columns, rows)
 
     form = GenericSubmitForm()
```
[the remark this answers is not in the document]
MarcUsher: The exact scenario suggested in the comment wouldn't work (if you uploaded a corrected file then clicked 'back' in your browser it wouldn't use the updated file as that hasn't been uploaded yet) but this feels like a reasonable safety check - if there's no missing data then you shouldn't land on this page and redirecting on to the next part of the flow seems sensible. Will update.

app/deliver_grant_funding/routes/collections.py
```diff
@@ -3384,12 +3385,17 @@ def _parse_data_set_csv(file_storage: FileStorage) -> tuple[list[str], TUnvalida
     return columns, rows
 
 
-def _load_and_validate_data_set(
-    data_set_data: DataSetUploadSessionModel,
-) -> tuple[TUnvalidatedDataSetRows, DataSetValidationResult]:
+def _load_rows(data_set_data: DataSetUploadSessionModel) -> TUnvalidatedDataSetRows:
     file_bytes = s3_service.download_file(data_set_data.s3_key)
```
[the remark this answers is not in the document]
MarcUsher: This is moot - this kind of mismatch is validated and caught on the initial file upload by the `UploadDataSet` form and the file would be rejected, so by the time it's being reloaded from S3 this won't be happening.

app/deliver_grant_funding/templates/deliver_grant_funding/collections/data_sets/map_columns.html
```diff
@@ -8,7 +8,11 @@
 {% set page_title = "Map " ~ session_data.name ~ " data columns" %}
 
 {% block beforeContent %}
-  {% set back_link = url_for("deliver_grant_funding.upload_data_set", grant_id=grant.id, collection_type=collection.type, collection_id=collection.id) %}
+  {% if session_data.has_missing_data %}
+    {% set back_link = url_for("deliver_grant_funding.data_set_missing_data", grant_id=grant.id, collection_type=collection.type, collection_id=collection.id) %}
+  {% else %}
+    {% set back_link = url_for("deliver_grant_funding.upload_data_set", grant_id=grant.id, collection_type=collection.type, collection_id=collection.id) %}
+  {% endif %}
   {{ govukBackLink({ "text": "Back", "href": back_link }) }}
```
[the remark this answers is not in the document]
MarcUsher: Have updated one of the map columns `get` tests to check the back link href if missing data is true/false

app/deliver_grant_funding/templates/deliver_grant_funding/collections/data_sets/data_set_missing_data.html
```diff
@@ -23,38 +18,32 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">
         <span class="govuk-caption-l">{{ collection.name }}</span>
-        Missing data in {{ session_data.name }}
+        Review missing data in {{ session_data.name }} data set
```
srh-sloan, line 21: Should we align this with `page_title` above?

tests/integration/deliver_grant_funding/test_data_sets.py
```diff
@@ -284,6 +230,39 @@ def test_multiple_errors_across_rows(self, factories):
         assert sum(isinstance(e, DataTypeError) for e in result.blocking_errors) == 1
 
 
+class TestCheckMissingData:
```
srh-sloan, line 233: Should we add a test for `check_missing_data` when there is no missing data?

srh-sloan (commented):
Couple of small comments but looks good and all works locally. Definitely cleans up the logic in the routes for the rest of the upload flow!

app/deliver_grant_funding/templates/deliver_grant_funding/collections/data_sets/data_set_missing_data.html
MarcUsher: Ah that was my bad, I'd updated one and not the other, but just pushed a change that has them both updated!

tests/integration/deliver_grant_funding/test_data_sets.py
MarcUsher: Yes good shout - have added a test for this!

srh-sloan (approved):
Nice! Good to streamline the routing and call out missing data early 👍

PR #1708, opened 2026-06-15
FSPT-1342: Flag missing data set data in views

MarcUsher:
## 🎫 Ticket
https://mhclgdigital.atlassian.net/browse/FSPT-1342

## 📝 Description
There are a few places that need to be aware of whether or not a data set has missing data in it (in this instance, missing data specifically means grant recipient rows with missing column values, not missing grant recipients entirely which will come in a later PR). 

Namely:
- The `list_reports` view needs to add a 'Data missing' tag if any data sets belonging to that report have missing data
- The `list_collection_data_sets` view needs to add a 'Data missing' tag if a data set in the list has missing data
- The `view_data_source` view itself needs to add a warning if a data set has missing data

This PR adds a hybrid property to the DataSource model to be able to work this out consistently, and a couple of new interfaces for list_reports and list_collection_data_sets including using a dataclass for the latter to only pull back the key info we need to populate the table, rather than firing off lots of additional queries, following the pattern established by the ListSubmissionsData dataclass.

_NB: The SQL side of the hybrid property had some help from AI to get the jsonb path stuff_

## 🧪 Testing
Upload some data sets with missing data and some complete ones and see that the tag and warning panel appear in the correct places.

### Review Readiness

### Performance and security

### Testing

### DIFF
--- a/app/common/data/interfaces/data_sets.py
+++ b/app/common/data/interfaces/data_sets.py
@@ -1,9 +1,12 @@
+import datetime
 import uuid
+from collections.abc import Sequence
+from dataclasses import dataclass
 from decimal import Decimal
 
-from sqlalchemy import select
+from sqlalchemy import func, select
 from sqlalchemy.exc import IntegrityError
-from sqlalchemy.orm import lazyload, selectinload
+from sqlalchemy.orm import aliased, lazyload, selectinload
 
 from app.common.data.interfaces.collections import raise_if_data_source_has_references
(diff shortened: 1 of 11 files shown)


srh-sloan (approved):
Pulled locally and works as expected 👍 
Nicely split out into different commits which made reviewing it really straight forward! Good tests ✨

srh-sloan (approved):

